PR #7, opened 2020-09-14
Apply content styles for HTML fragments.

dbatiste:


grant-cleary (approved):

htmleditor.js
```diff
@@ -36,18 +36,39 @@ import { RtlMixin } from '@brightspace-ui/core/mixins/rtl-mixin.js';
...
 
+const rootFontSize = window.getComputedStyle(document.documentElement, null).getPropertyValue('font-size');
+
 const pathFromUrl = (url) => {
 	return url.substring(0, url.lastIndexOf('/'));
 };
 
 const baseImportPath = pathFromUrl(import.meta.url);
 
+const contentFragmentCss = css`
```
dbatiste, line 52: These are basically copied from the LMS, omitting some default tinyMCE styles, and also omitting the ISF styles that will be added when they are needed.

PR #11, opened 2020-09-24
Add ability to show different toolbar in normal vs fullscreen view.

dbatiste:
* adds styles to show/hide toolbars based on fullscreen state
* adds visual-diff test for fullscreen mode


htmleditor.js
```diff
@@ -207,9 +219,10 @@ class HtmlEditor extends RtlMixin(LitElement) {
 				skin_url: `${baseImportPath}/tinymce/skins/ui/oxide`,
 				statusbar: true,
 				target: textarea,
-				toolbar: this.inline
-					? 'bold italic underline'
-					: 'bold italic underline | strikethrough subscript superscript | bullist numlist | indent outdent | alignleft alignright aligncenter alignjustify | charmap hr | table | forecolor | styleselect fontselect fontsizeselect | undo redo | preview code fullscreen | ltr rtl | a11ycheck',
+				toolbar: this.inline ? 'bold italic underline' : [
+					'bold italic underline | bullist numlist | fullscreen',
+					'bold italic underline | styleselect fontselect fontsizeselect | forecolor a11ycheck | bullist numlist | indent outdent | alignleft alignright aligncenter alignjustify | strikethrough subscript superscript | charmap hr | table | undo redo | ltr rtl | preview code fullscreen'
```
grant-cleary, line 224: Ah okay, so we do just hide whichever toolbar row is inappropriate based on fullscreen status. I love that this is in the visual diff too. This is exactly the sort of thing that could break on TinyMCE upgrades, so it's nice to have an automated way to catch those failures. :)

grant-cleary (approved):

htmleditor.js
dbatiste: Yeah agreed, when we meet with the tiny people we can raise this as something that we're doing.  It would be nice if the fullscreen plugin provided some configuration to handle this in a way that we won't have to worry about breaking if the selectors change.  Unfortunately, as we push the limits of what tinyMCE is capable of, we'll inevitably need to start hooking in at a deeper level.

Yeah I haven't really spent any time setting up visual-diff tests because the design is still in flux, but in this case, I figured it was a good opportunity to set up a test that shows interaction with tinyMCE, which could be done via either a simulated click on the fullscreen toolbar button, or as a direct command to the editor as I've done here.  I don't think we need to set up tests for tinyMCE, but we should at least capture the more interesting aspects that would not be covered by tinyMCE's own tests.

PR #12, opened 2020-10-01
Add initial quicklinks plugin and dialog.

dbatiste:
The module includes two distinct blocks of code that may eventually be separated.  For now, they can live in the same module.
1. the tinyMCE plugin
    - defines the button for the tinyMCE toolbar
    - logic for getting editor caret context (create vs edit)
    - launches the dialog, waits for the result, and inserts it
2. the dialog to create/edit a quicklink
    - abstracts the logic for calling into the LMS quicklinks actions (if we had quicklinks components this would internally use d2l-dialog and it would not make these LMS calls)
    - opens the LMS dialog for either creating or editing a quicklink
    - creates the quicklink once an item has been selected
    - eventually the LMS calls (launching the dialog and the RPC calls) will likely be abstracted to make this ifrau-aware

This is the sort of stuff that is going to be pretty typical for custom integrations.  We did the same logic previously, however the tinyMCE plugin part was abstracted considerably.

### DIFF
--- a/components/quicklink-dialog.js
+++ b/components/quicklink-dialog.js
@@ -1,0 +1,160 @@
+import '@brightspace-ui/core/components/colors/colors.js';
+import 'tinymce/tinymce.js';
+import { css, LitElement } from 'lit-element/lit-element.js';
+import { getComposedActiveElement } from '@brightspace-ui/core/helpers/focus.js';
+import { icons } from '../icons.js';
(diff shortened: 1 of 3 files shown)


grant-cleary (approved):
Cool, I like that we're treating these as actual custom TinyMCE plugins. :)

PR #19, opened 2020-10-14
Add equation editor component.

dbatiste:
Includes:
* split button for launching editors
* ability to launch graphical (wiris) editor
* ability to launch chemistry (wiris) editor
* ability to launch latex editor
* ability to launch mathml editor
* editing existing equations


components/equation.js
```diff
@@ -0,0 +1,274 @@
+/* eslint no-useless-escape: 0 */
```
dbatiste, line 1: I was a little nervous about changing the regex statements, so I left them for the time being.

components/equation.js
```diff
@@ -0,0 +1,274 @@
...
+			// get the title of the equation
+			let title = match.match(/title=(\"|\')([^\"]*)(\"|\')/i);
+			if ((title !== null) && (title.length === 4) && (title[2] !== null)) title = title[2];
+			else title = '';
+
+			const annotationRegex = /<annotation encoding="([^"]*)">/gi;
+			const annotationMatch = annotationRegex.exec(match);
+
+			let removeAnnotation = false;
+			let classString = 'equation';
```
dbatiste, line 120: I am going to update to avoid the string concat here.  Will change to use a template literal.

components/equation.js
```diff
@@ -0,0 +1,274 @@
...
+
+			const previewImageSrc = `/wiris/editorservice.aspx/render.png?mml=${encodedMml}`;
+			const placeholderImageSrc = "'/d2l/img/LP/htmlEditor/equation_unavailable.png'";
+			const placeholderImageTitle = "'Preview image not available for this equation'";
+
+			// if there's an error loading preview image, placeholder image will be shown instead
+			const onError = `if (this.src != ${placeholderImageSrc}) this.src = ${placeholderImageSrc};
+				this.alt = ${placeholderImageTitle};
+				this.title = ${placeholderImageTitle};`;
+			const imageHtml = `<img class="${classString}"
```
dbatiste, line 152: This should be exactly the same as original, except that I converted it to a template literal.

components/equation.js
```diff
@@ -0,0 +1,274 @@
...
+				style="vertical-align: middle;"
+				hspace="5"
+				vspace="5"
+				onerror="${onError}"
+				data-d2l-mathml="${encodedMml}" \/>`;
+
+			return imageHtml;
+		};
+
+		const appletFunc = (match) => {
```
dbatiste, line 165: I'm almost templated to remove this applet stuff.  It's been many years since it's been used.

dbatiste:
There appears to be a bit of a timing issue with the toolbar being updated after focus is applied in the editor, resulting in inconsistent visual-diff test.  Will try to address that.

components/equation.js
```diff
@@ -0,0 +1,274 @@
...
+				icon: 'd2l-equation-mathml',
+				text: 'MathML Equation',
+				value: editorTypes.MathML,
+				disabled: contextNode && editorType !== editorTypes.MathML
+			}, {
+				type: 'choiceitem',
+				icon: 'd2l-equation-chemistry',
+				text: 'Chemistry Equation',
+				value: editorTypes.Chemistry,
+				disabled: contextNode && editorType !== editorTypes.Chemistry
```
grant-cleary, line 102: So I think this is a departure from existing behaviour, in that the chemistry and normal graphical editors are currently both valid options when selecting a `wiris` or `wiris-chemistry` equation. Not sure it really matters, since I doubt many people take advantage of that.

components/equation.js
grant-cleary: Yeah, I would be really surprised if there were any active courses that still had these lying around.

components/equation.js
dbatiste: I think I could enable that if it's a value-add.  I'm not sure what that really enables the user to do though...?  i.e. why they the user would do that.

grant-cleary: So the main driver is that the chemistry toolbar isn't actually unique. It's basically just a rearrangement of the normal toolbar with the periodic table added (that just inserts normal text for the element names). In that sense, it's possible to seamlessly switch between the two editors. I don't know that anyone does that though, so I'm personally not too bothered if we don't bother with this.

grant-cleary (approved):

PR #20, opened 2020-10-16
US121356 - Add image handling support (ImageTools and PowerPaste) to HTML editor component

grant-cleary:
Associated Rally ticket: [US121346](https://rally1.rallydev.com/#/detail/userstory/441139328652?fdp=true)

Here we do the actual component work to get PowerPaste/ImageTools working within the new editor. There's some stuff in here that isn't hooked up in the LMS yet, so this won't be the last PR.

Things that still need doing:
- Content needs to be updated to move files out of temp and into the content directory
- Handlers for the upload started/completed events need to be wired up in the event itself so we can do things like disable buttons while the upload is progressing
- `context.maxFileSize` is not yet passed through from the LMS so will always be undefined (for now). There are a couple different ways we could handle this so I don't want to commit (yet).
- The LMS does some stuff with `files` and `files_queued` that I haven't yet fixed up.
- The renderer needs to be updated to provide the `localImagePasting` attribute when appropriate.


htmleditor.js
```diff
@@ -87,6 +89,18 @@ const contentFragmentStyles = css`
 	}
 `.cssText;
 
+class FileData {
```
grant-cleary, line 92: No idea if there's a better place for this to live.

htmleditor.js
```diff
@@ -317,5 +351,63 @@ class HtmlEditor extends ProviderMixin(RtlMixin(LitElement)) {
 		}
 	}
 
+	_imageUploadHandler(blobInfo, success, failure) {
+		this._uploadImageCount++;
+		if (this._uploadImageCount === 1) { // only fire the upload started event on the first image being uploaded
+			this.dispatchEvent(new CustomEvent(
```
grant-cleary, line 357: I'm not sure if this is what we want here; I just used other plugins as an example. I haven't tested this yet because I haven't hooked handlers up on the LMS side.

htmleditor.js
```diff
@@ -317,5 +351,63 @@ class HtmlEditor extends ProviderMixin(RtlMixin(LitElement)) {
...
+				'd2l-htmleditor-image-upload-started', {
+					bubbles: true
+				}
+			));
+		}
+
+		// Local image upload requires an LMS context for now
+		if (!D2L.LP || !context || context.orgUnitId === null || context.maxFileSize === null) return;
```
grant-cleary, line 365: So `context.maxFileSize` is something I wanted to check up on here. This is required to build the upload location (unfortunately). This makes me wonder if we might want to just provide the upload location instead and build the route via .NET methods. I'm not sure how feasible this is given that we expect a `D2L.LP.Web.Http.UrlLocation` when feeding this into the actual upload route though. :(

htmleditor.js
```diff
@@ -317,5 +351,63 @@ class HtmlEditor extends ProviderMixin(RtlMixin(LitElement)) {
...
+							uploadedFile.Size,
+							src
+						)
+					);
+
+					success(src);
+
+					this._uploadImageCount--;
+					if (this._uploadImageCount <= 0) {
+						this.dispatchEvent(new CustomEvent(
```
grant-cleary, line 397: Same comment as the event above.

htmleditor.js
```diff
@@ -317,5 +351,63 @@ class HtmlEditor extends ProviderMixin(RtlMixin(LitElement)) {
...
+		D2L.LP.Web.UI.Html.Files.FileUpload.XmlHttpRequest.UploadFiles(
+			[blob],
+			{
+				UploadLocation: uploadLocation,
+				OnFileComplete: (uploadedFile) => {
+					const src = `/d2l/lp/files/temp/${uploadedFile.FileId}/View`;
+
+					this.filesQueued++;
+					this.files.push(
+						new FileData(
```
grant-cleary, line 384: The intent here is to mimic [this](https://search.d2l.dev/xref/lms/lp/framework/web/D2L.LP.Web.UI/Desktop/Controls/HtmlEditor/Renderers/WysiwygEditor/Core/TinyMCEv4/Wrapper.js?r=aa583098#392) to keep the APIs identical

grant-cleary:
General note: The ImageTools plugin does provide an Insert/Edit Image toolbar button, but we don't use it in our current editors (we rely on the provided context menu instead), so I opted to do the same here.

htmleditor.js
```diff
@@ -229,17 +256,24 @@ class HtmlEditor extends ProviderMixin(RtlMixin(LitElement)) {
...
+				plugins: `a11ychecker charmap code directionality emoticons ${this.fullPage ? 'fullpage' : ''} fullscreen hr image ${this.localImagePasting ? 'imagetools' : ''} lists powerpaste preview table d2l-equation d2l-isf d2l-quicklink`,
 				relative_urls: false,
 				resize: true,
 				setup: (editor) => {
 					editor.ui.registry.addIcon('resize-handle', icons['resize-handle']);
 
+					if (this.localImagePasting) {
+						editor.on('blur', () => {
+							editor.uploadImages();
+						});
```
grant-cleary, line 274: I'm not sure if there's a better way to wire up this event. This is done this way to force image uploads as soon as focus moves out of the editor. The ImageTools plugin doesn't do this automatically for some actions within its context menu (e.g. rotating the image).

htmleditor.js
grant-cleary: Hmm, I don't think this works as is anyways. Pretty sure the LMS file data object exposes its properties through functions. Probably would either have to reproduce those as well or alternatively just use the LMS object.

grant-cleary: Actually, I'll take another look at this. It might be fine?

dbatiste: To your question about a better place for it to live.  Not critical now, but I'd be inclined to move it down below the HtmlEditor class.  It's just a simple data object and unlikely to be the focus of anyone looking here.

grant-cleary: Fair point. I moved it up here because the linter yelled at me about using an undefined class, but I can move it down and make sure the linter ignores it.

htmleditor.js
```diff
@@ -200,10 +221,16 @@ class HtmlEditor extends ProviderMixin(RtlMixin(LitElement)) {
 			}
 
 			const powerPasteConfig = {
```
dbatiste, line 223: I'd probably opt for for succinct format.
```
const powerPasteConfig = {
    powerpaste_allow_local_images: this.localImagePasting,
    powerpaste_block_drop: !this.localImagePasting,
    powerpaste_word_import: context ? context.pasteFormatting : 'merge'
};
```

htmleditor.js
```diff
@@ -229,17 +256,24 @@ class HtmlEditor extends ProviderMixin(RtlMixin(LitElement)) {
 				font_formats: 'Arabic Transparent=arabic transparent,sans-serif; Arial (Recommended)=arial,helvetica,sans-serif; Comic Sans=comic sans ms,sans-serif; Courier=courier new,courier,sans-serif; Ezra SIL=ezra sil,arial unicode ms,arial,sans-serif; Georgia=georgia,serif; SBL Hebrew=sbl hebrew,times new roman,serif; Simplified Arabic=simplified arabic,sans-serif; Tahoma=tahoma,sans-serif; Times New Roman=times new roman,times,serif; Traditional Arabic=traditional arabic,serif; Trebuchet=trebuchet ms,helvetica,sans-serif; Verdana=verdana,sans-serif; 돋움 (Dotum)=dotum,arial,helvetica,sans-serif; 宋体 (Sim Sun)=simsun; 細明體 (Ming Liu)=mingliu,arial,helvetica,sans-serif',
 				fontsize_formats: '8pt 10pt 12pt 14pt 18pt 24pt 36pt',
 				height: this.height,
+				images_upload_handler: (blobinfo, success, failure) => this._imageUploadHandler(blobinfo, success, failure),
```
dbatiste, line 259: Nit: `blobinfo` -> `blobInfo`.

htmleditor.js
grant-cleary: Haha still not super used to `falsey` things in JS.

htmleditor.js
dbatiste: Should it be necessary, we can attach info to this event.  Not sure if that would be needed by consumers...?  No worries if not sure... just something to consider as we go forward.

htmleditor.js
dbatiste: What currently defines the `maxFileSize`?

For the moment, we can assume the UrlLocation is present.
```
const someUrl = new D2L.LP.Web.Http.UrlLocation('/d2l/lp/foo/bar');
```
This is what I am currently doing to build the urls that are required in the other integrations (quicklinks, isf, equation editor), since those methods also require a `UrlLocation`.

So, unless we can assume a static max file size, I think it would need to be added to the context, and then we use it  to build the url here, as I've done for the others.

Eventually when we deal with FRAs, the url construction will need to either be duplicated here, or it will need to be done through iFrau.  That's my current thinking.

htmleditor.js
grant-cleary: I _think_ consumers don't actually need to consume any info from this event. Right now, the callbacks passed in for these events are all argument-less and everything but News just uses the default callback [here](https://search.d2l.dev/xref/lms/le/D2L.LE/D2L.LE.Web.UI/Desktop/Utils/HtmlEditorHelper.cs?r=5d364101#29). Even in the case of News, [I don't think we actually pull any info from the editor](https://search.d2l.dev/xref/lms/le/news/webpages/newedit.xml?r=2a1ab9be#110).

grant-cleary: That's a blatant lie. We do pull the editor text.

htmleditor.js
grant-cleary: The `maxFileSize` is governed by a config variable and locally I have it set up to just feed into the context through the provider and a bit further on in the handler I do use it to construct the URL. Sounds like I'm doing it the way we want to do this for now then.

htmleditor.js
```diff
@@ -102,7 +116,10 @@ class HtmlEditor extends ProviderMixin(RtlMixin(LitElement)) {
 			noSpellchecker: { type: Boolean, attribute: 'no-spellchecker' },
 			type: { type: String },
 			width: { type: String },
-			_editorId: { type: String }
+			localImagePasting: { type: Boolean, attribute: 'local-image-pasting' },
+			files: { type: Array, },
+			filesQueued: { type: Number, attribute: 'files_queued' },
```
dbatiste, line 121: Do we need to expose `filesQueued` as a public property?  i.e. what will the consumer do with that?

grant-cleary: I think we could refactor the API so that this isn't the case. Right now the API does mutate this value though: https://search.d2l.dev/xref/lms/lp/framework/web/D2L.LP.Web.UI/Desktop/Controls/HtmlEditor/HtmlEditor.js?r=c79ec01c#89

htmleditor.js
grant-cleary: My first thought is that attaching the entire HTML body is likely overkill given that only one page uses that, mind you.

htmleditor.js
dbatiste: It seems like this state is specific to the MVC JS wrapper, and the web component should just expose what is necessary - so dispatch an event and expose the files.  The MVC JS wrapper can keep track of which files it has handled and which ones haven't.  The web component itself doesn't need to be concerned with which files are queued for handling.

grant-cleary: Sounds good to me.

htmleditor.js
dbatiste: I think this looks correct to me. 

Nit: unless more code is going to be added, I'd simplify it to:
``` javascript
editor.on('blur', () => editor.uploadImages());
```
or even...
``` javascript
editor.on('blur', editor.uploadImages);
```

grant-cleary: Cool, I can do that. I don't expect more to be added here.

dbatiste (approved):

htmleditor.js
```diff
@@ -229,17 +237,20 @@ class HtmlEditor extends ProviderMixin(RtlMixin(LitElement)) {
...
 				menubar: false,
 				object_resizing : true,
-				plugins: `a11ychecker charmap code directionality emoticons ${this.fullPage ? 'fullpage' : ''} fullscreen hr lists powerpaste preview table d2l-equation d2l-isf d2l-quicklink`,
+				plugins: `a11ychecker charmap code directionality emoticons ${this.fullPage ? 'fullpage' : ''} fullscreen hr image ${this.pasteLocalImages ? 'imagetools' : ''} lists powerpaste preview table d2l-equation d2l-isf d2l-quicklink`,
 				relative_urls: false,
 				resize: true,
 				setup: (editor) => {
 					editor.ui.registry.addIcon('resize-handle', icons['resize-handle']);
 
+					if (this.pasteLocalImages) editor.on('blur', () => editor.uploadImages());
```
grant-cleary, line 252: TinyMCE really didn't like `editor.on('blur', editor.uploadImages)`. It complained that the blur callback wasn't a function.

dbatiste: Weird.  🤷‍♂️

PR #22, opened 2020-10-20
US121648 - Add autosave plugin to new editor

grant-cleary:
Associated Rally ticket: [US121648](https://rally1.rallydev.com/#/detail/userstory/447898818428?fdp=true)

This PR does more or less what it says on the tin, really. Note that we have a pretty weird way of using autosave... We unload it as soon as the editor initializes so we don't actually get data saved in localStorage (but we still get to keep the prompt if a user tries to leave before saving).


htmleditor.js
```diff
@@ -94,7 +95,8 @@ class HtmlEditor extends ProviderMixin(RtlMixin(LitElement)) {
 
 	static get properties() {
 		return {
-			files: { type: Array, },
+			autosave: { type: Boolean },
```
grant-cleary, line 98: I'll get an LMS PR up for this shortly. Just juggling some other HTML editor related changes on a few branches so I don't want to muddy things too much.

htmleditor.js
```diff
@@ -223,6 +226,9 @@ class HtmlEditor extends ProviderMixin(RtlMixin(LitElement)) {
 				a11ychecker_allow_decorative_images: true,
 				allow_html_in_named_anchor: true,
 				allow_script_urls: true,
+				autosave_ask_before_unload: this.autosave,
+				autosave_restore_when_empty: false,
+				autosave_retention: "0s",
```
grant-cleary, line 231: Oh, you know, these can totally be in their own block like the PowerPaste ones.

htmleditor.js
dbatiste, line 98: For consistent property/attribute casing, we should name this property `autoSave`/`auto-save`.

htmleditor.js
```diff
@@ -297,7 +311,8 @@ class HtmlEditor extends ProviderMixin(RtlMixin(LitElement)) {
 				valid_elements: '*[*]',
 				width: this.width,
 				...fullPageConfig,
-				...powerPasteConfig
+				...powerPasteConfig,
+				...autosaveConfig
```
dbatiste, line 315: Nit: let's keep them alpha order.

htmleditor.js
```diff
@@ -296,6 +310,7 @@ class HtmlEditor extends ProviderMixin(RtlMixin(LitElement)) {
 				toolbar: this._getToolbarConfig(),
 				valid_elements: '*[*]',
 				width: this.width,
+				...autoSaveConfig,
```
grant-cleary, line 313: Is this what we want, or do we want these blocks interspersed in alpha order through the init list?

dbatiste: I think this is what we want.

grant-cleary:
Er... What the hell went wrong with resolving that merge conflict...

grant-cleary:
... Alright, I broke this in a way I can't easily undo. I'm just going to nuke this PR and open a new one. -_-

PR #25, opened 2020-10-21
Refactor file data class, image upload handler.

dbatiste:
* move FileData into the image plugin
* move the image upload handler into the image plugin
* add private helper method for adding files to the files array
* update the image plugin to add file to `files`

### DIFF
--- a/htmleditor.js
+++ b/htmleditor.js
@@ -339,10 +339,6 @@ class HtmlEditor extends ProviderMixin(RtlMixin(LitElement)) {
 		tinymce.EditorManager.get(this._editorId).focus();
 	}
 
-	_addFile(file) {
-		this.files.push(file);
-	}
-
 	_getToolbarConfig() {
 		if (this.type === editorTypes.INLINE_LIMITED) {
 			return 'bold italic underline | d2l-list d2l-isf emoticons';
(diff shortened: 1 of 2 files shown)


htmleditor.js
```diff
@@ -342,6 +339,10 @@ class HtmlEditor extends ProviderMixin(RtlMixin(LitElement)) {
 		tinymce.EditorManager.get(this._editorId).focus();
 	}
 
+	_addFile(file) {
```
dbatiste, line 342: I made this private since I don't want any consumer thinking they should use this.

htmleditor.js
```diff
@@ -1,7 +1,5 @@
-import './components/image.js';
 import './components/quicklink.js';
 import './components/equation.js';
-import '@brightspace-ui/core/components/colors/colors.js';
```
dbatiste, line 4: I don't think this is being used - I might have added it when I was experimenting with the daylight palette.

htmleditor.js
dbatiste: Note: I decided to remove this, since it's currently not needed.  My original idea was that we would eventually dispatch a `file-add` event here, but I'm not convinced we will actually do that... so for now, not adding this private helper.

grant-cleary (approved):

PR #28, opened 2020-10-22
US121648 - Emit d2l-htmleditor-ready event when TinyMCE editor has finished initialization so that consumers can e.g. focus on the component once it loads

grant-cleary:
Associated Rally ticket: [US121734](https://rally1.rallydev.com/#/detail/userstory/448887517928?fdp=true)

Two main things here:
1. Rename existing PowerPaste-specific events to follow best practice.
2. Emit a `d2l-htmleditor-ready` event when TinyMCE is done initializing so consumers can hook into the editor on load.


grant-cleary:
Ah bugger I accidentally made this change on an old branch. Deleting and re-creating.

PR #30, opened 2020-10-22
US121734 - Emit d2l-htmleditor-ready event when TinyMCE editor has finished initialization so that consumers can e.g. focus on the component once it loads

grant-cleary:
Associated Rally ticket: [US121734](https://rally1.rallydev.com/#/detail/userstory/448887517928?fdp=true)

Two main things here:
1. Rename existing PowerPaste-specific events to follow best practice.
2. Emit a `d2l-htmleditor-ready` event when TinyMCE is done initializing so consumers can hook into the editor on load.

### DIFF
--- a/components/image.js
+++ b/components/image.js
@@ -66,7 +66,7 @@ export function uploadImage(editor, blobInfo, success, failure) {
 				editor._uploadImageCount--;
 				if (editor._uploadImageCount <= 0) {
 					editor.dispatchEvent(new CustomEvent(
-						'd2l-htmleditor-image-upload-finish', {
+						'd2l-htmleditor-image-upload-complete', {
 							bubbles: true
 						}
 					));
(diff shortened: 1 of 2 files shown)


components/image.js
```diff
@@ -66,7 +66,7 @@ export function uploadImage(editor, blobInfo, success, failure) {
 				editor._uploadImageCount--;
 				if (editor._uploadImageCount <= 0) {
 					editor.dispatchEvent(new CustomEvent(
-						'd2l-htmleditor-image-upload-completed', {
+						'd2l-htmleditor-image-upload-finish', {
```
dbatiste, line 69: Nit: can we name it `d2l-htmleditor-image-upload-complete`?

dbatiste (approved):
Looks good - I'd just stick with the "complete" term.

dbatiste (approved):

PR #33, opened 2020-11-10
US122296 - Allow localization of D2L-specific lang terms in HTML editor

grant-cleary:
Pretty much does what it says on the tin; enabling support for localizing D2L-specific terms in the editor for custom functionality. This also localizes tooltips for current custom plugins based on existing lang terms in the LMS.

### DIFF
--- a/htmleditor.js
+++ b/htmleditor.js
@@ -22,6 +22,7 @@ import { css, html, LitElement, unsafeCSS } from 'lit-element/lit-element.js';
 import { getUniqueId } from '@brightspace-ui/core/helpers/uniqueId.js';
 import { icons } from './icons.js';
 import { isfStyles } from './components/isf.js';
+import { Localizer } from './lang/localizer.js';
 import { ProviderMixin } from '@brightspace-ui/core/mixins/provider-mixin.js';
 import { RtlMixin } from '@brightspace-ui/core/mixins/rtl-mixin.js';
 import { tinymceLangs } from './generated/langs.js';
@@ -32,7 +33,6 @@ import { uploadImage } from './components/image.js';
 // 2. copy new language packs from https://www.tiny.cloud/get-tiny/language-packages/ into tinymce/langs
 // 3. copy new enterprise plugins into tinymce/plugins
 
-// TODO: set powerpaste_word_import based on paste formatting config value (clean, merge, prompt)
 // TODO: review whether pasted content needs prepcessing to avoid pasted image links getting converted to images
 // TODO: configure paste_as_text if using tinyMCEs paste as text feature (fra editor sets to false if power paste enabled) - probably not needed
 // TODO: review whether we need to stop pasting of image addresses (see fra editor)
@@ -89,7 +89,7 @@ const contentFragmentStyles = css`
 	}
 `.cssText;
 
-class HtmlEditor extends ProviderMixin(RtlMixin(LitElement)) {
+class HtmlEditor extends ProviderMixin(Localizer(RtlMixin(LitElement))) {
 
 	static get properties() {
 		return {
@@ -127,9 +127,6 @@ class HtmlEditor extends ProviderMixin(RtlMixin(LitElement)) {
 			}
 			.tox .tox-statusbar {
 				border-top: none;
-			}
-			.tox .tox-statusbar__text-container {
-				display: none;
 			}
 			:host([type="inline"]) .tox-tinymce .tox-toolbar-overlord > div:nth-child(2) {
 				display: none;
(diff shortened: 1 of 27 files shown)


htmleditor.js
```diff
@@ -128,9 +128,6 @@ class HtmlEditor extends ProviderMixin(RtlMixin(LitElement)) {
 			.tox .tox-statusbar {
 				border-top: none;
 			}
-			.tox .tox-statusbar__text-container {
-				display: none;
-			}
```
grant-cleary, line 133: Ah crap, I didn't mean to let this change sneak in. This (plus `elementPath: false` in TinyMCE's init) is just a slightly cleaner way to remove the element path from the TinyMCE status bar.

htmleditor.js
```diff
@@ -267,7 +266,7 @@ class HtmlEditor extends ProviderMixin(RtlMixin(LitElement)) {
 				},
 				// inline: this.type === editorTypes.INLINE || this.type === editorTypes.INLINE_LIMITED,
 				language: tinymceLang,
-				language_url: `/tinymce/langs/${tinymceLang}.js`,
+				language_url: `${baseImportPath}/tinymce/langs/${tinymceLang}.js`,
```
grant-cleary, line 269: I guess when TinyMCE inits, its current directory is the "base" page, not necessarily the path where the component is being served from.

lang/ar.js
```diff
@@ -0,0 +1,12 @@
...
+
+export default {
+	"htmleditor.equationeditor.chemistrytooltip": "معادلة في الكيمياء",
+	"htmleditor.equationeditor.graphicaltooltip": "معادلة رسومية",
+	"htmleditor.equationeditor.latextooltip": "معادلة LaTeX",
+	"htmleditor.equationeditor.mathmltooltip": "معادلة MathML",
+	"htmleditor.image.tooltip": "إدراج صورة",
+	"htmleditor.insertstuff.tooltip": "إدراج الأغراض",
+	"htmleditor.preview.tooltip": "معاينة",
+	"htmleditor.quicklink.tooltip": "إدراج ارتباط سريع"
```
grant-cleary, line 11: I'm not super keen on these names, but I was tired last night and couldn't think of anything different. I can find/replace with better options.

lang/localizer.js
```diff
@@ -0,0 +1,74 @@
+import { LocalizeMixin } from '@brightspace-ui/core/mixins/localize-mixin.js';
+
+export const Localizer = superclass => class extends LocalizeMixin(superclass) {
```
grant-cleary, line 3: I thought it was cleaner for this to be a separate class as opposed to jamming this `getLocalizeResources` function into the `HtmlEditor` class.

htmleditor.serge.json
```diff
@@ -0,0 +1,27 @@
...
+      "ja-jp ja",
+      "ko-kr ko",
+      "nl-nl nl",
+      "pt-br pt",
+      "sv-se sv",
+      "tr-tr tr",
+      "zh-cn zh",
+      "zh-tw zh-tw"
+    ]
+  }
```
grant-cleary, line 27: I'm surprised the linter didn't yell at me for missing this new line. Unless it doesn't matter for JSON?

htmleditor.js
dbatiste, line 269: What was happening to require this?

grant-cleary: TinyMCE was trying to fetch these files from `{MyLocalInstancePath}/tinymce/langs/...` instead of `{BSIPath}/node_modules/@brightspace-ui/...`.

lang/en.js
```diff
@@ -0,0 +1,12 @@
+/* eslint quotes: 0 */
+
+export default {
+	"htmleditor.equationeditor.chemistrytooltip": "Chemistry Equation",
```
dbatiste, line 4: I don't feel strongly about it, but `htmleditor` part seems a bit redundant.  i.e. everything in here is going to have that, and I think the terms are scoped and should not conflict with any other components.  🤷‍♂️

grant-cleary: Is there any requirement that these be three-part like in the LMS? I stuck with LMS convention of `collection.package.termname` mostly because that's what I know, but I absolutely don't have strong opinions on that. I'm happy to make it less cumbersome and remove cruft.

dbatiste (approved):

htmleditor.js
dbatiste: Interesting I didn't notice this before - it seemed to be working.

lang/en.js
dbatiste: I don't think it needs to be 3 part.

htmleditor.js
grant-cleary: I'm curious if it works if I don't serve BSI locally, but I think probably not because the paths are so different. If it was working before I'm curious what changed... I don't think it's anything in this PR, but now I'm a bit concerned, haha.

PR #35, opened 2020-11-17
US122488 - Prevent loading of plugins requiring D2L-specific context in external editors

grant-cleary:
Associated Rally ticket: https://rally1.rallydev.com/#/detail/userstory/457626000916?fdp=true

When an editor is marked as external, we should prevent plugins requiring D2L-specific context from loading. Unauthenticated users can't make authenticated requests to the LMS, including opening LMS dialogs.


CODEOWNERS
```diff
@@ -1 +1,2 @@
 *       @dbatiste
+*       @grant-cleary
```
grant-cleary, line 2: I figured it would be appropriate to throw myself in here.

htmleditor.js
```diff
@@ -217,7 +219,7 @@ class HtmlEditor extends ProviderMixin(Localizer(RtlMixin(LitElement))) {
 			}
 
 			const powerPasteConfig = {
-				powerpaste_allow_local_images: this.pasteLocalImages,
+				powerpaste_allow_local_images: this.pasteLocalImages && !this.external,
```
grant-cleary, line 222: So on top of anything that launches an LMS dialog, we also can't allow local image pasting or the ImageTools plugin to function in these contexts because they require authenticated requests (specifically file uploads) to the LMS.

dbatiste (approved):
Might not be a bad idea to add a demo-snippet for the external case.  Not too worried about it, but will help show what things look like in such a case.

grant-cleary:
Looks like Design is going to revisit whether we need to do this. Closing for now.

PR #37, opened 2020-11-18
switch to GitHub actions

dlockhart:
This will temporarily disable visual diff tests, but I think you're OK with that. When we do add them, I think it'll make sense to do that as a separate job, at which point you may want some NPM caching to speed things up. But we can cross that bridge then.


.github/workflows/ci.yml
```diff
@@ -0,0 +1,25 @@
+name: CI
+on: pull_request
+jobs:
+  test:
+    name: Test
+    timeout-minutes: 10
```
dlockhart, line 6: Not sure how long your tests take, but this can be adjusted.

.github/workflows/release.yml
```diff
@@ -0,0 +1,28 @@
+name: Release
+on:
+  push:
+    branches:
+      - master
+      - '[0-9]+.x'
+      - '[0-9]+.[0-9]+.x'
```
dlockhart, line 7: Not sure if you care about releasing from `1.x` or `1.3.x` maintenance branches -- this can be removed if not.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,25 @@
...
+        run: npm run lint:eslint
+      - name: Lint (CSS)
+        run: npm run lint:style
+      - name: Lint (Lit)
+        run: npm run lint:lit
+      - name: SauceLabs
+        env:
+          SAUCE_USERNAME: Gaudi1
+          SAUCE_ACCESS_KEY: ${{ secrets.SAUCE_ACCESS_KEY }}
+        run: npm run test:sauce
```
grant-cleary, line 25: Oh nice, I didn't know this was supported for GHA!

grant-cleary (approved):

dlockhart:
Crazy, passed first try! 🚀

grant-cleary:
> Crazy, passed first try! 🚀

That's unpossible. REVERT.

dbatiste (approved):

PR #41, opened 2020-11-23
US122517 - Add support for overriding TinyMCE iframe title and aria-label attributes

grant-cleary:
TinyMCE puts a title on its iframe by default, but this doesn't particularly serve our purpose. Instead, allow the consumer of the web component to specify the title and (optionally) an aria-label to be used instead.

### DIFF
--- a/htmleditor.js
+++ b/htmleditor.js
@@ -103,7 +103,6 @@ class HtmlEditor extends ProviderMixin(Localizer(RtlMixin(LitElement))) {
 			fullPageFontSize: { type: String, attribute: 'full-page-font-size' },
 			height: { type: String },
 			html: { type: String },
-			label: { type: String },
 			noFilter: { type: Boolean, attribute: 'no-filter' },
 			noSpellchecker: { type: Boolean, attribute: 'no-spellchecker' },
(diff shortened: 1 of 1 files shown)


htmleditor.js
```diff
@@ -257,11 +261,19 @@ class HtmlEditor extends ProviderMixin(Localizer(RtlMixin(LitElement))) {
...
+
+					if (editor.plugins && editor.plugins.autosave) {
 						// removing the autosave plugin prevents saving of content but retains the "ask_before_unload" behaviour
 						delete editor.plugins.autosave;
 					}
 
+					const iframe = editor.getContentAreaContainer().firstElementChild;
+					if (iframe) {
+						iframe.title = this.title;
+						if (this.label) iframe.setAttribute('aria-label', this.label);
```
grant-cleary, line 274: I'm not entirely sure if this is how we want to implement these. I'm definitely open to rethinking this.

dbatiste (approved):

grant-cleary:
I'm going to merge this to keep conflicts to a minimum. If the LMS side of it doesn't make it in until next week, that seems fine.

PR #46, opened 2020-12-02
US121779 - Introduce word count functionality to the new editor

grant-cleary:
This PR introduces the word count algorithm to be used by the editor and wires that up into a custom plugin we use to display that. Note that at the moment we're just using a placeholder icon.


package.json
```diff
@@ -12,7 +12,8 @@
     "/components",
     "/generated",
     "/lang",
-    "/tinymce"
+    "/tinymce",
+    "/wordcount/wordcount.js"
```
grant-cleary, line 16: Eventually we want the server-side word count code to also live in this repo, so I'm pre-emptively splitting word count stuff into its own folder to enable that later.

components/wordcount.js
```diff
@@ -0,0 +1,368 @@
...
+tinymce.PluginManager.add('d2l-wordcount', function(editor) {
+
+	const wordCountInFooter = requestInstance(editor.getElement(), 'wordCountInFooter');
+	const localize = requestInstance(editor.getElement(), 'localize');
+
+	const root = editor.getElement().getRootNode();
+
+	const countButton = document.createElement('button');
+	countButton.id = 'd2l-wordcount-footer-button';
+	countButton.setAttribute('data-d2l-count-option', footerCountType.WORD);
```
grant-cleary, line 27: I think there were some questions about whether we actually wanted the footer to act as a button. I'm uncertain about the way I'm using this button. It works, but I'm not sure it's proper form.

components/wordcount.js
```diff
@@ -0,0 +1,368 @@
...
+		`];
+	}
+
+	constructor() {
+		super();
+		this.counts = {};
+		this.countType = footerCountType.WORD;
+		this.opened = false;
+		this.selectedCounts = {};
+		this._width = 500;
```
dbatiste, line 216: Looks like this could just get moved inline - no need to store this on the object.

components/wordcount.js
```diff
@@ -0,0 +1,368 @@
...
+
+	_handleSelectWordCountOption() {
+		this._setSelectedCountOption(footerCountType.WORD);
+	}
+
+	_renderCountSelectionDropdown() {
+		return html`
+			<d2l-dropdown-button-subtle slot="footer" id="d2l-wordcount-option" class="d2l-wordcount-option-button" text="${this._getCountOptionText()}">
+				<d2l-dropdown-menu>
+					<d2l-menu>
```
dbatiste, line 268: Note: `d2l-menu` should have a `label` applied.  While not visible, it gets announced to screen readers.

htmleditor.js
```diff
@@ -110,6 +111,7 @@ class HtmlEditor extends ProviderMixin(Localizer(RtlMixin(LitElement))) {
 			title: { type: String },
 			type: { type: String },
 			width: { type: String },
+			wordCountInFooter: { type: Boolean, attribute: 'word-count-in-footer' },
```
dbatiste, line 114: Iccky name eh.  If I understand correctly, if this is `false`, then they won't see the word count in the footer, but they will see the word count button in the toolbar.  Is that correct?

grant-cleary: That's right, yeah. It's a bit gross. ☹️

dbatiste: Just to double check - it looks like the default for this is going to be false.  What's the rationale behind that?  I'm goo either way... just want to make sure we define the API such that no attribute is required for the common/default case.

grant-cleary: So my understanding is that the "word count in footer" thing is tied to the server-side word count that's going to be implemented as well. Currently that only exists in Discussions for clients that have enabled a specific config variable and my expectation is that it's likely to stay relatively limited. It may end up being the case that this becomes more common, but right now the vast majority of editors for the vast majority of clients definitely don't have this enabled.

components/wordcount.js
```diff
@@ -0,0 +1,368 @@
...
+			default:
+				return isSelection ? 'wordcount.footer.selectionwordcount' : 'wordcount.footer.wordcount';
+		}
+	};
+
+	const openWordCountDialog = () => {
+		let dialog = root.querySelector('d2l-htmleditor-wordcount-dialog');
+		if (!dialog) dialog = root.appendChild(document.createElement('d2l-htmleditor-wordcount-dialog'));
+
+		const text = editor.getContent({ source_view: true, format: 'text' });
```
grant-cleary, line 53: Blech, I just remembered from some previous testing that this doesn't work if the `fullpage` plugin is used. It ends up supplying the text-formatted editor contents... Surrounded by the `fullpage` HTML. I'll play around and see if one of TinyMCE's existing methods works to filter that out. Otherwise I think we'll need to do a bit of light HTML stripping if the fullpage plugin is enabled.

dbatiste: What are the options for tinyMCE's API?  Maybe tinyMCE can provide what we need without us needing to strip things out.

grant-cleary: Yeah, I'm going to investigate that. TinyMCE doesn't really document what you can provide in the `options` block to this `getContent` call, so it's entirely possible there's some option that does what we want.

grant-cleary: I'm a bit suspicious of `source_view` so trying with that set to false is likely my first guess.

grant-cleary: Um. Actually, `getContent` as written above appears to work even with fullpage. This is a bit confusing because it's definitely not how I remember this behaving.

components/wordcount.js
```diff
@@ -0,0 +1,368 @@
...
+
+	if (!wordCountInFooter) return;
+
+	countButton.onclick = openWordCountDialog;
+
+	editor.on('init', () => {
+		const statusBar = root.querySelector('.tox-statusbar');
+		const statusBarResizeHandler = statusBar.querySelector('.tox-statusbar__resize-handle');
+
+		statusBar.insertBefore(countButton, statusBarResizeHandler);
```
dbatiste, line 129: I am a little worried that if the editor is rerendered, the count button might get blown away.  It's be worth testing this out, perhaps by changing one of the editor component properties after the initial render.

grant-cleary: Hmm, interesting thought, yeah. I'll take a look and see if that happens. I guess that feeds back into the question: Is the button the best place to store the current count mode? Right now that's how the code has it working, but if the button can get blown away, it might be better to find somewhere else to store it so we can just recreate the button on init and not have to worry about losing state.

components/wordcount.js
```diff
@@ -0,0 +1,368 @@
...
+	});
+
+	const onInputDelay = 100;
+	let lastRunTime = Date.now();
+	editor.on('change selectionchange textInput input', (e) => {
```
dbatiste, line 134: Does this capture cases where the contents of the editor have changed due to new HTML being set on the editor, either by the public API, the user modifying the source in the source code editor, insertion of symbols, and potentially other toolbar actions that insert content?  More of a curiosity - how is Math treated?

components/wordcount.js
```diff
@@ -0,0 +1,368 @@
...
+	countButton.onclick = openWordCountDialog;
+
+	editor.on('init', () => {
+		const statusBar = root.querySelector('.tox-statusbar');
+		const statusBarResizeHandler = statusBar.querySelector('.tox-statusbar__resize-handle');
+
+		statusBar.insertBefore(countButton, statusBarResizeHandler);
+	});
+
+	const onInputDelay = 100;
```
dbatiste, line 132: I'm not really sure what interval is reasonable here.  i.e. as the document gets reasonably big, does this tie up the main thread as the user is typing?  If this becomes a concern, we can consider ways to mitigate potential input latency.

components/wordcount.js
grant-cleary: Good question; I pulled this from the current editor but haven't verified what happens if the APIs (or TinyMCE's internals) are used to update the HTML. In terms of math, that's also a good question. I was relying on TinyMCE's text filtering removing any MathML as HTML, but I'm also unsure. Thinking on it now, I wouldn't be surprised if they stripped the tags but not the contents... Maybe that's okay? I don't know what I would expect to happen there. =/

components/wordcount.js
grant-cleary: Yeah, I just took this number from the existing algorithm, though I modified it slightly (original code [here](https://search.d2l.dev/xref/lms/lp/framework/web/D2L.LP.Web.UI/Desktop/Controls/HtmlEditor/Renderers/WysiwygEditor/Core/TinyMCEv4/Wrapper.js?r=aa583098#521)). I found that the `setTimeout` method had a bit of a bug in it where if you input quickly enough, multiple instances of the event could end up firing and thus `timerId` would never get reset, causing the count to effectively stop updating.

components/wordcount.js
```diff
@@ -0,0 +1,368 @@
...
+			countType: { type: String },
+			opened: { type: Boolean, reflect: true },
+			selectedCounts: { type: Object }
+		};
+	}
+
+	static get styles() {
+		return [
+			selectStyles,
+			css`
```
dbatiste, line 167: Nit... indentation seems off here, but it's just because of the css placement.  I'd move it up to the line above.
``` javascript
return [ selectedStyles, css`
    /* the styles */
`];
```

components/wordcount.js
grant-cleary: I did try to get a bit of perf testing in with a very large file, but what I found was that Chrome largely stopped being responsive entirely before I got to a chunk of text large enough to cause any problems with the input event. All that said, I'm also running a relatively powerful dev machine so I'm not super sure if those characteristics change on a weaker machine.

components/wordcount.js
grant-cleary: Ah, from the 5.6 changelog: ```FIXED fullpage plugin altering text content in editor.getContent().```
Note that I also need to make some changes here, because of this: ```FIXED getContent with text format returning a new line when the editor is empty.```
So I can get rid of the weird edge case where I have to omit newlines when the editor is empty.

components/wordcount.js
```diff
@@ -0,0 +1,368 @@
...
+		super.connectedCallback();
+		this._localize = this.requestInstance('localize');
+		this._wordCountInFooter = this.requestInstance('wordCountInFooter');
+	}
+
+	render() {
+		return html`
+			<d2l-dialog width="${this._width}" title-text="${this._localize('wordcount.dialog.title')}" ?opened="${this.opened}" @d2l-dialog-close=${(this._handleDialogClosed)}>
+				${this._renderWordCountInfo()}
+				<d2l-button slot="footer" primary data-dialog-action>${this._localize('wordcount.dialog.closebutton')}</d2l-button>
```
dbatiste, line 229: I usually try to avoid floats unless I really want it to "float".  We almost never use CSS `float` anymore since it is rife with problems for layout.  I think I would be inclined to:
```html
<div slot="footer">
    <d2l-button primary data-dialog-action>${this._localize('wordcount.dialog.closebutton')}</d2l-button>
    ${this._wordCountInFooter ? this._renderCountSelectionDropdown() : ''}
</div>
```

```css
[slot="footer"] {
    display: flex;
    flex-wrap: wrap;
}
d2l-button {
    flex: auto;
}
```

Not sure if the flex wrap is needed.  In theory if the dialog is too narrow for both the buttons to be on one line, then it would need to be allowed to wrap... probably requires a tiny bit of investigation to see if that scenario would every come up.

components/wordcount.js
grant-cleary: Ok, so the `setContent` API, options like inserting symbols, and the source code editor all do fire one of these events and the count updates accordingly. Math entered through any of our editors doesn't count towards the count because TinyMCE strips it out (they're all images in the editor view). We do the same image conversion for MathML directly entered through the source editor, so the same applies. I imagine it's possible to find a way to shoehorn math in there some other way that doesn't get "image-ified", in which case I imagine the "text" inside the math tags would probably be counted.

components/wordcount.js
```diff
@@ -0,0 +1,368 @@
...
+		this._setSelectedCountOption(footerCountType.NOTHING);
+	}
+
+	_handleSelectWordCountOption() {
+		this._setSelectedCountOption(footerCountType.WORD);
+	}
+
+	_renderCountSelectionDropdown() {
+		return html`
+			<d2l-dropdown-button-subtle slot="footer" id="d2l-wordcount-option" class="d2l-wordcount-option-button" text="${this._getCountOptionText()}">
```
dbatiste, line 266: Assuming we can switch to using flexbox, this `d2l-wordcount-option-button` class should not be needed.

components/wordcount.js
grant-cleary: Sounds good to me! I more or less just copied what the LMS does in some places, so I'm more than happy to implement a better way. This button has a relatively large lang term so I think we could run into this scenario in some edge cases, so I'll poke at the wrapping.

components/wordcount.js
```diff
@@ -0,0 +1,368 @@
...
+
+	return editor.getBody().getElementsByTagName('p').length;
+}
+
+function _countSelectedParagraphs(editor) {
+	if (!editor.selection || editor.selection.isCollapsed()) return 0;
+
+	const html = editor.selection.getContent({ format: 'html' });
+	const template = document.createElement('template');
+	template.insertAdjacentHTML('afterbegin', html);
```
dbatiste, line 355: What happens if the user starts a selection in the middle of paragraph 1, and ends the selection in paragraph 2?

components/wordcount.js
grant-cleary: Boo, changing the HTML directly on the element (as we do with our `setHtml` API) doesn't seem to trigger these events.

components/wordcount.js
grant-cleary: This seems to work. I'm not sure if this is a valid test, but I changed the html property through the browser dev console and the button was functional.

components/wordcount.js
grant-cleary: It gets counted as two paragraphs. TinyMCE implicitly wraps the text in opening and closing `p` blocks such that we don't get malformed HTML. e.g. If we have text like this:
```Hello
there
everyone
```
and we select "llo \n th" from that list, `getContent({ format: 'html' })` outputs `<p>llo</p><p>th</p`

dbatiste: Perfect.  😄

components/wordcount.js
dbatiste: Sounds good, and yep, that should be enough to trigger the rerender.

components/wordcount.js
grant-cleary: Ok, so very specifically: the moment the cursor becomes active within the editor, the count updates, but this is _not_ the case when the editor's init runs. This means toolbar buttons and stuff work fine because they return focus to the editor window, but when the editor first opens on any page that doesn't immediately set focus into the editor, the word count won't be correct. I think the solution is likely to just calculate and display the count as part of the `init` event.

grant-cleary: That does mean we'll get some double counts on pages where focus immediately moves into the active editor (one count on init and one when the selectionchange event fires), but I think that's okay.

dbatiste: Sounds reasonable.

components/wordcount.js
```diff
@@ -0,0 +1,366 @@
...
+
+	editor.on('init', () => {	
+		const statusBar = root.querySelector('.tox-statusbar');
+		const statusBarResizeHandler = statusBar.querySelector('.tox-statusbar__resize-handle');
+
+		statusBar.insertBefore(countButton, statusBarResizeHandler);
+	});
+
+	let timerId = 0;
+	editor.on('change input selectionchange setcontent textInput', (e) => {
```
grant-cleary, line 136: I switched this back to use a `setTimeout` like the original code, but this time I clear the timeout if another event comes in during the debounce period. It turns out this is necessary because TinyMCE fires a `selectionchange` event every time the carat moves, which means having those events selectively bypass debouncing basically means we do no debouncing. -_-

Also note the addition of `setcontent` - TinyMCE fires this event whenever the `setContent` API is invoked, which should cover cases where we load in the HTML via setting the `html` property on the editor.

components/wordcount.js
```diff
@@ -0,0 +1,366 @@
...
+			:host([hidden]) {
+				display: none;
+			}
+			[slot="footer"] {
+				display: flex;
+				flex-wrap: wrap;
+				justify-content: space-between;
+			}
+			d2l-button {
+				flex: 0 1 auto;
```
grant-cleary, line 177: Despite `0 1 auto` being the default for the `flex` property, `flex: auto` was setting this to `1 1 auto` instead, causing the Close button to balloon comically across the dialog.

lang/en.js
```diff
@@ -8,5 +8,26 @@ export default {
...
+	"wordcount.dialog.paragraphs": "Paragraphs",
+	"wordcount.dialog.selectionlabel": "Selection",
+	"wordcount.dialog.title": "Word Count",
+	"wordcount.dialog.words": "Words",
+	"wordcount.footer.charactercount": "{count, plural, one { 1 Character } other { { count } Characters }}",
+	"wordcount.footer.selectioncharactercount": "{count, plural, one { 1 Character } other { { count } Characters }} in selection",
+	"wordcount.footer.selectionwordcount": "{count, plural, one { 1 Word } other { { count } Words }} in selection",
+	"wordcount.footer.wordcount": "{count, plural, one { 1 Word } other { { count } Words }}",
+	"wordcount.footerselector.charactercount": "Character Count",
+	"wordcount.footerselector.menulabel": "Type of count displayed in editor footer",
```
grant-cleary, line 26: I have serious doubts about this lang term but I couldn't think of anything better. I'm definitely open to suggestions here.

components/wordcount.js
```diff
@@ -0,0 +1,366 @@
...
+		return [ selectStyles, css`
+			:host {
+				display: block;
+			}
+			:host([hidden]) {
+				display: none;
+			}
+			[slot="footer"] {
+				display: flex;
+				flex-wrap: wrap;
```
grant-cleary, line 173: Editors opened in dialogs can be shrunk to the point that we need to care about this, especially since the `Footer Displays: Character Count` lang term is relatively long (and could be longer in some translations). I figure this is very much an edge case, but I don't think it hurts to account for it here.

components/wordcount.js
dbatiste: Yeah I wasn't sure if that would happen - if this works then that's nice and clean.  I thought you might end up needing an extra nested div to prevent that:
```html
<div style="display: flex;">
    <div style="flex: auto;">
        <d2l-button></d2l-button>
    </div>
    ...
</div>
```

lang/en.js
dbatiste: Ug yeah.  Maybe "Foot count type".  We can also leave it to Design to decide - in order for them to provide a good suggestion though, they will need to know how it's sounds to the screen reader user in context.

grant-cleary: I'll run through it with NVDA to see exactly what it spits out and then I can pass the buck to design. :P

grant-cleary:
Ok, I _think_ with that last lang term, we're good except for the icons (which I feel comfortable merging without).

dbatiste (approved):

PR #49, opened 2020-12-07
US120079 - Add support for @mentions plugin

grant-cleary:
Enables client-side support for the @mentions plugin in the new HTML editor component.


htmleditor.js
```diff
@@ -277,8 +280,24 @@ class HtmlEditor extends ProviderMixin(Localizer(RtlMixin(LitElement))) {
 				language: tinymceLang,
 				language_url: `${baseImportPath}/tinymce/langs/${tinymceLang}.js`,
 				menubar: false,
+				mentions_fetch: (query, success) => {
+					setTimeout(() => D2L.LP.Web.UI.Rpc.Connect(
```
grant-cleary, line 284: So as of now I just opted to copy what the LMS does here, but I'm unsure about this. We need to auth against the LMS to make this work right now (and currently @mentions doesn't function in FRAs so this is the only way it's done), but I'm not sure if there's a better way to make the actual RPC. I toyed around with `d2lfetch` (with `d2l-fetch-auth) but couldn't get it working and it wasn't entirely clear that was the right direction anyways.

htmleditor.js
```diff
@@ -277,8 +280,24 @@ class HtmlEditor extends ProviderMixin(Localizer(RtlMixin(LitElement))) {
...
+						D2L.LP.Web.UI.Rpc.Verbs.GET,
+						new D2L.LP.Web.Http.UrlLocation('/d2l/lp/htmleditor/tinymce/mentionsSearchQuery'),
+						{ searchTerm: query.term, orgUnitId: context.orgUnitId },
```
grant-cleary, line 287: This is admittedly making the assumption that if `D2L.LP` exists, `context` also exists. I don't know how valid that assumption is. If we think it's not, I can add an explicit `context` check to loading the plugin as well.

htmleditor.js
dbatiste: I think this is fine for now.  Pretty soon we need to adapt these LMS calls to work through iFrau.

dbatiste (approved):

PR #67, opened 2021-01-05
US122517 - Fire custom event when blur event is fired from the editor; expose editor.isDirty() method on component

grant-cleary:
This makes a couple changes to the component that are likely necessary to align with the old LMS editor's JS APIs. There are two main changes here:
1. Fire a custom event that bubbles up to consumers when the editor itself fires a blur event. This is necessary to handle some change tracking we do in some of our legacy UIs (specifically, on editor blur is when we check whether changes have occurred and update the UI accordingly).
2. Expose TinyMCE's `isDirty()` editor API to consumers of the component. This is the method some of our legacy UI code uses to determine whether an editor has changes or not (different from the change tracking mentioned above).

### DIFF
--- a/htmleditor.js
+++ b/htmleditor.js
@@ -333,6 +333,16 @@ class HtmlEditor extends ProviderMixin(Localizer(RtlMixin(LitElement))) {
 				setup: (editor) => {
 					addIcons(editor);
 
+					editor.on('blur', () => {
+						if (this.pasteLocalImages) editor.uploadImages();
+
+						this.dispatchEvent(new CustomEvent(
+							'd2l-htmleditor-blur', {
+								bubbles: true
+							}
+						));
+					});
+
 					if (this.pasteLocalImages) editor.on('blur', () => editor.uploadImages());
 
 					const createSplitButton = (name, icon, tooltip, cmd, items) => {
@@ -419,6 +429,11 @@ class HtmlEditor extends ProviderMixin(Localizer(RtlMixin(LitElement))) {
 
 	get initializationComplete() {
 		return this._initializationComplete;
+	}
+
+	get isDirty() {
+		const editor = tinymce.EditorManager.get(this._editorId);
+		return editor.isDirty();
 	}
 
 	_getToolbarConfig() {
--- a/htmleditor.js
+++ b/htmleditor.js
@@ -433,7 +433,7 @@ class HtmlEditor extends ProviderMixin(Localizer(RtlMixin(LitElement))) {
 
 	get isDirty() {
 		const editor = tinymce.EditorManager.get(this._editorId);
-		return isShadowDOMSupported ? editor.isDirty() : false;
+		return isShadowDOMSupported && editor.isDirty();
 	}
 
 	_getToolbarConfig() {
--- a/htmleditor.js
+++ b/htmleditor.js
@@ -433,7 +433,7 @@ class HtmlEditor extends ProviderMixin(Localizer(RtlMixin(LitElement))) {
 
 	get isDirty() {
 		const editor = tinymce.EditorManager.get(this._editorId);
-		return isShadowDOMSupported && editor.isDirty();
+		return (editor && editor.isDirty());
 	}
 
 	_getToolbarConfig() {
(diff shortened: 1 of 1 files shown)


htmleditor.js
```diff
@@ -333,6 +333,16 @@ class HtmlEditor extends ProviderMixin(Localizer(RtlMixin(LitElement))) {
 				setup: (editor) => {
 					addIcons(editor);
 
+					editor.on('blur', () => {
+						if (this.pasteLocalImages) editor.uploadImages();
+
+						this.dispatchEvent(new CustomEvent(
+							'd2l-htmleditor-blur', {
```
grant-cleary, line 340: I figured a generic name would be better here, though I'm also definitely open to suggestions otherwise. :)

htmleditor.js
```diff
@@ -421,6 +431,11 @@ class HtmlEditor extends ProviderMixin(Localizer(RtlMixin(LitElement))) {
 		return this._initializationComplete;
 	}
 
+	get isDirty() {
+		const editor = tinymce.EditorManager.get(this._editorId);
+		return editor.isDirty();
```
grant-cleary, line 436: An example of how this is used can be found [here](https://search.d2l.dev/xref/lms/lp/framework/web/D2L.Web/ui/RichEdit/JavaScript/RichEdit.js?r=7cfef204#55).

htmleditor.js
grant-cleary: Also worth mentioning... I tried to get around the need for this by just listening to `focusout` events from the editor component in the MVC code, but I found that it still didn't work as I expected. e.g. Clicking into the editor text area and clicking out didn't seem to trigger a `focusout` event, which I found odd.

dbatiste: FYI, we might eventually need to add an event for focus as well if we don't already have one.  Some consumers (a couple of the rubrics cases) _might_ need to keep track of whether the editor has focus or not (including focus on something inside the editor but not the text area).

grant-cleary: Ah, true. I can add that now if you'd prefer (to keep things more complete), or if we want to leave it until we need it that's fine. I'm not 100% sure on the specifics of some of the FRA/external app cases.

htmleditor.js
dbatiste: Be careful of the case where `isShadowDOMSupported` is `false`.  I think if the consumer tried to call this API in such case it would produce a JavaScript error.  In such case where we don't have an editor, we should probably just return `false`.

grant-cleary: Good call. I forgot about the addition of that check.

htmleditor.js
dbatiste: Yeah same here.  We'd just need to make sure that the focus event matches (target-wise) the blur event.  So if in Rubrics case they want that blue outer boarder to show up when it has focus, would it show when the buttons or one of the dialogs has focus, or not.

grant-cleary: Makes sense. TinyMCE's documentation says that they do support the `focus` event, so I would hope (maybe naïve of me) that they would both use the same target. I imagine it wouldn't be too tough to test that though. The editor seems to fire the `blur` event whenever focus leaves anything within the editor (but not, for example, when focus moves between editor components), so if `focus` behaves the same way, that would be nice.

htmleditor.js
```diff
@@ -421,6 +431,11 @@ class HtmlEditor extends ProviderMixin(Localizer(RtlMixin(LitElement))) {
 		return this._initializationComplete;
 	}
 
+	get isDirty() {
+		const editor = tinymce.EditorManager.get(this._editorId);
+		return isShadowDOMSupported && editor.isDirty();
```
dbatiste, line 436: If there were any other reason the editor didn't get initialized, this would also yield a JS error.  Might be acceptable assumption that that would not come up but I was thinking it would just check...
```
return (editor && editor.isDirty());
```

grant-cleary: Ooh, I see. I assume the editor wouldn't be initialized at all if the browser didn't support shadowDOM? i.e. That explicit check is unnecessary?

dbatiste (approved):

PR #68, opened 2021-01-07
Prevent editor from resizing below 300px or its initialized height, whichever is smaller

grant-cleary:
Not too much to say on this one. TinyMCE provides a `min_height` config to control how small the editor height can get. We should set this so users can't accidentally resize the editor to be too small to interact with.

### DIFF
--- a/htmleditor.js
+++ b/htmleditor.js
@@ -310,6 +310,7 @@ class HtmlEditor extends ProviderMixin(Localizer(RtlMixin(LitElement))) {
 				link_context_toolbar: true,
 				link_default_protocol: 'https',
 				menubar: false,
+				min_height: this.height < 300 ? this.height : 300,
 				mentions_fetch: (query, success) => {
 					setTimeout(() => D2L.LP.Web.UI.Rpc.Connect(
 						D2L.LP.Web.UI.Rpc.Verbs.GET,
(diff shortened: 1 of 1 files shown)


htmleditor.js
```diff
@@ -343,8 +344,6 @@ class HtmlEditor extends ProviderMixin(Localizer(RtlMixin(LitElement))) {
 						));
 					});
 
-					if (this.pasteLocalImages) editor.on('blur', () => editor.uploadImages());
```
grant-cleary, line 346: This is duplicated above; I just forgot to remove this line in my last PR.

htmleditor.js
```diff
@@ -310,6 +310,7 @@ class HtmlEditor extends ProviderMixin(Localizer(RtlMixin(LitElement))) {
 				link_context_toolbar: true,
 				link_default_protocol: 'https',
 				menubar: false,
+				min_height: this.height < 300 ? this.height : 300,
```
dbatiste, line 313: Is this correct?  Seems like it should be the other way around.

dbatiste: Oh I see.  If a provided height is less, then we use that as the min.

grant-cleary: Ah, my idea here was that if the initialized height was `< 300`, we should use that as our minimum height instead, otherwise users could never resize the editor back down to its initial height. It's not 100% clear to me whether that's how we want to handle this edge case.

dbatiste (approved):

PR #71, opened 2021-01-12
Update README with missing property descriptions

grant-cleary:
First shot at the missing property descriptions. Some of them don't feel fantastic to me, so I'm for sure open to suggestions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,24 +38,24 @@ Types of editors (toolbar features):
 
 | Property | Type | Description |
 |--|--|--|
-| `attached-images-only` | Boolean | Need description. Defaults to `false`. |
-| `auto-save` | Boolean | Need description. Defaults to `false`. |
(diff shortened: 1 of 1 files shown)


dbatiste (approved):

dbatiste:
I think we will need to document the `isDirty` property, and also the `d2l-htmleditor-blur` event.  Sounds like components (such as in FACE) will likely use the blur event as a signal to update their state-store with the updated HTML.

grant-cleary:
> I think we will need to document the `isDirty` property, and also the `d2l-htmleditor-blur` event. Sounds like components (such as in FACE) will likely use the blur event as a signal to update their state-store with the updated HTML.

Cool, I can take care of that too. The other property I wonder about is the `isInitialized` thing or whatever. If any other application  needs to focus on the editor for whatever reason, they'll need to check that first I think.

PR #73, opened 2021-01-13
Parse units from provided height property and ensure we're setting the minimum height accordingly

grant-cleary:
When we compare the default minimum height of `300px` against the height of the editor to determine which value to use, we should make sure to compare appropriately against whatever units the consumer provides. We can't really convert `%` well because we don't have access to the height of the bounding container, but we can ensure we're converting `px` and `rem` properly by parsing out the units and either using the values directly (in the case of `px`) or converting to `px` and then comparing (in the case of `rem`).

### DIFF
--- a/htmleditor.js
+++ b/htmleditor.js
@@ -446,11 +446,13 @@ class HtmlEditor extends ProviderMixin(Localizer(RtlMixin(LitElement))) {
 		const heightValue = heightParts[0];
 		const heightUnits = heightParts[1];
 
+		const fontSizeValue = rootFontSize.replace('px','');
+
 		switch (heightUnits) {
 			case 'px':
 				return heightValue < defaultMinHeight ? heightValue : defaultMinHeight;
 			case 'rem':
-				return (heightValue * 20) < defaultMinHeight ? (heightValue * 20) : defaultMinHeight;
+				return (heightValue * fontSizeValue) < defaultMinHeight ? (heightValue * fontSizeValue) : defaultMinHeight;
 			default:
 				return defaultMinHeight;
 		}
--- a/htmleditor.js
+++ b/htmleditor.js
@@ -446,7 +446,7 @@ class HtmlEditor extends ProviderMixin(Localizer(RtlMixin(LitElement))) {
 		const heightValue = heightParts[0];
 		const heightUnits = heightParts[1];
 
-		const fontSizeValue = rootFontSize.replace('px','');
+		const fontSizeValue = rootFontSize.replace('px', '');
 
 		switch (heightUnits) {
 			case 'px':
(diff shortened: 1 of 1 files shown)


htmleditor.js
```diff
@@ -437,6 +437,25 @@ class HtmlEditor extends ProviderMixin(Localizer(RtlMixin(LitElement))) {
 		return (editor && editor.isDirty());
 	}
 
+	_getMinHeight() {
+		const defaultMinHeight = 300;
+		const heightParts = this.height.split(/(?<=\d)(?=\D)/);
```
grant-cleary, line 442: Split the string based on the pattern `[digit, not-a-digit]`, e.g. `200px -> [200, 'px']`. If we don't end up with exactly two parts, the format we have is malformed and we return the default `300px` instead.

htmleditor.js
```diff
@@ -437,6 +437,25 @@ class HtmlEditor extends ProviderMixin(Localizer(RtlMixin(LitElement))) {
...
+		if (heightParts.length !== 2) return defaultMinHeight;
+
+		const heightValue = heightParts[0];
+		const heightUnits = heightParts[1];
+
+		switch (heightUnits) {
+			case 'px':
+				return heightValue < defaultMinHeight ? heightValue : defaultMinHeight;
+			case 'rem':
+				return (heightValue * 20) < defaultMinHeight ? (heightValue * 20) : defaultMinHeight;
```
grant-cleary, line 453: If we get the height in `rem`, just convert it to pixels and compare that to the default minimum height. TinyMCE always expects `min_height` to be a number in `px` so we also return the converted value.

dbatiste: FYI, at the top of this file we have the root font-size.  Unfortunately it contains the `px` units so you'd have to chop those off in order to use it.  Should be a matter of `rootFontSize.replace('px', '')`... yields 20.

dbatiste (approved):

htmleditor.js
grant-cleary: Do you think we should be accounting for the root font-size somehow?

grant-cleary: Sorry if I'm misunderstanding your FYI there.

dbatiste: The 20 you have hard-coded is the root font-size.  Instead of hard-coding that, we can get it from `rootFontSize.replace('px', '')`.  It's unlikely to change anytime soon, but if we ever change our typography, there's a good chance we may change the root font-size.

grant-cleary: Oh of course, yes. I'll change this to be safe.

grant-cleary:
You know, I always say to myself: "I'm just adding one line of code, I don't need to run the linter". Then it turns out I miss some whitespace.

dbatiste (approved):

PR #74, opened 2021-01-14
Implement SkeletonMixin on editor component

grant-cleary:
Implement the skeleton mixin so consumers don't have to, with a minimum of mucking about with styling native TinyMCE elements. Caveat: Because the editor has a margin and the skeleton styles are applied to an inner div, the overall skeletized animation is slightly smaller than the editor appears to be. It's hard to get a feel just by looking at it, but here's an example:

Not skeletized:
HtmlEditorNoSkeleton

Skeletized:
HtmlEditorSkeleton


htmleditor.js
```diff
@@ -156,6 +160,9 @@ class HtmlEditor extends ProviderMixin(Localizer(RtlMixin(LitElement))) {
 				background-color: #ffffff;
 				z-index: 1000;
 			}
+			:host([skeleton]) .tox .tox-editor-header {
+				z-index: 0;
```
grant-cleary, line 164: This is hopefully temporary, pending https://github.com/BrightspaceUI/core/pull/1038, but is required to stop the menu from clipping over the skeleton. It occurs to me that I can also just increase the z-index on the `d2l-skeletize` class. If the referenced PR isn't merged, I think that would be better than playing around with TinyMCE's styles.

htmleditor.js
```diff
@@ -415,10 +422,18 @@ class HtmlEditor extends ProviderMixin(Localizer(RtlMixin(LitElement))) {
...
+		const skeletonClasses = {
+			'd2l-skeletize': this.skeleton
+		};
+
 		//if (this.type === editorTypes.INLINE || this.type === editorTypes.INLINE_LIMITED) {
 		//	return html`<div id="${this._editorId}" .innerHTML="${this._html}"></div>`;
 		//} else {
-		return html`<textarea id="${this._editorId}" class="${classMap(textAreaClasses)}" aria-hidden="true" tabindex="-1">${this._html}</textarea>
+		return html`
+			<div class="${classMap(skeletonClasses)}">
```
grant-cleary, line 434: Wrapping this in a div avoids the need for us to apply skeleton styles to TinyMCE CSS classes, since we have no means to do so outside of overwriting it directly and making this component very fragile.

htmleditor.js
grant-cleary: Actually, re-reading Dave L's comment, it looks like maybe the idea is that we should fix this in here and not alter the z-index of the `d2l-skeletize` class... In which case I think I would want to switch this over to override the z-index on `d2l-skeletize` for this component instead of relying on TinyMCE's junk. I'm confirming with Dave.

grant-cleary:
TODO: Add visual diff tests for skeleton mode.

htmleditor.js
```diff
@@ -135,6 +136,9 @@ class HtmlEditor extends ProviderMixin(Localizer(RtlMixin(LitElement))) {
 			:host([hidden]) {
 				display: none;
 			}
+			:host([skeleton]) {
+				border: none;
```
dlockhart, line 140: Is this needed? The skeleton mixin should apply a `sylvite` border colour, which is the same colour as the rest of the skeleton and will override the mica one here. That way it doesn't shrink by 1px all around.

htmleditor.js
```diff
@@ -169,7 +176,7 @@ class HtmlEditor extends ProviderMixin(Localizer(RtlMixin(LitElement))) {
 				display: none;
 			}
 			/* stylelint-enable selector-class-pattern */
-		`;
+		`];
 	}
```
dlockhart, line 181: Is there any plan to build a `label="label for the editor"` into the component? Mostly just curious, but I noticed Maragret's example had a label. If it's part of the component (like all our other `<d2l-input-*>` components), we can wire it up to internal form elements using `for` or `aria-labelledby`, vs. external labels cannot do that. Also that would let HTML editor skeletize the label whenever itself is skeletized.

htmleditor.js
dlockhart: Ah I see, the `skeletize` CSS class is being applied internally to the `<div>`, not to the host. One option would be to move this border onto that `<div>`.

htmleditor.js
grant-cleary: So this is something we chatted about previously. In the LMS editor, the label/title are applied by taking them and attaching them to TinyMCE's internal iframe (the LMS editor frequently uses external labels outside the component). What I did for this component was to take the `title` attribute and apply it the same way. Is your thought here that we could create an explicit label from the label text within the component and then attach a `labelledby` to the iframe instead?

htmleditor.js
grant-cleary: Good thought. I'll give that a shot. The border issue is something that was bugging me, but I didn't know if there was a clean solution. I didn't think to move the border itself. :facepalm:

htmleditor.js
dlockhart: Right, so if you look at [input-text](https://github.com/BrightspaceUI/core/blob/master/components/inputs/input-text.js) for example, we use a combination of `label` (string) and `label-hidden` (boolean) to render either a `<label for="inputId">label text</label>` pointing at the text input, or if `label-hidden` is set we put `aria-label` right on the input.

Then in the LMS, since you're right the labeling is a bit weirder there we plan to just always use `label-hidden` but set `label` to the value from the LMS label.

Your case is slightly different since you're not dealing with a native input that MUST be labelled for accessibility... but I think in theory the same principle should apply? Otherwise you're going to get folks trying to link labels to the `<d2l-htmleditor>` and it not working the way they expect.

grant-cleary:
Apologies if the visual diff stuff ends up a bit noisy. For whatever reason I can't generate goldens locally (puppeteer always times out), so I can't test that piece easily.

htmleditor.js
grant-cleary: Cool, gotcha. I'm fine with aligning the component to be what we would expect in other components. I'll take a poke at it with a screenreader to see if the title on the iframe becomes a problem, but I'll figure it out.

grant-cleary:
> Apologies if the visual diff stuff ends up a bit noisy. For whatever reason I can't generate goldens locally (puppeteer always times out), so I can't test that piece easily.

Blech, I mucked it up enough that it would be way too noisy to fix here. I'll continue debugging puppeteer locally to try to figure out why it always times out for me.

grant-cleary:
Ah... Hilariously the reason the skeleton golden is broken is because the editor is still fullscreened from the previous test. 😆

dbatiste:
> Ah... Hilariously the reason the skeleton golden is broken is because the editor is still fullscreened from the previous test. 😆

Yeah sometimes we need a reset such as...
```javascript
beforeEach(async() => {
    await helper.reset(page, '#confirm');
});
```
Where we put the fixture back in its original state.  The alternative is that we reload the page between tests, but we haven't been doing that because we don't want hundreds of page reloads for our tests.

grant-cleary:
> Yeah sometimes we need a reset such as...
> 
> ```js
> beforeEach(async() => {
>     await helper.reset(page, '#confirm');
> });
> ```
> 
> Where we put the fixture back in its original state. The alternative is that we reload the page between tests, but we haven't been doing that because we don't want hundreds of page reloads for our tests.

Ah, makes sense. I sort of just cheated by running the skeleton tests first, but I can also look into this.

test/htmleditor.visual-diff.js
```diff
@@ -8,7 +9,7 @@ describe('d2l-htmleditor', () => {
 	let browser, page;
 
 	before(async() => {
-		browser = await puppeteer.launch();
+		browser = await puppeteer.launch({ args: ['--no-sandbox'] });
```
grant-cleary, line 12: ... I didn't mean to commit this. Whoops.

test/htmleditor.visual-diff.js
```diff
@@ -37,4 +42,17 @@ describe('d2l-htmleditor', () => {
 
 	});
 
+	describe('skeleton', () => {
+
+		afterEach(async() => {
+			await helper.reset(page, '#skeleton');
+		});
+
+		it('normal', async function() {
```
dbatiste, line 51: I think I'd be inclined to put this in the other describe block and just always reset all the fixtures.  Originally I was thinking along the lines of `full`, `inline`, `inline-limited`, and maybe a block of tests that cover all the things that are common.  We may choose to organize the tests differently as we expand on them however.  nbd though.

dbatiste (approved):

test/htmleditor.visual-diff.html
```diff
@@ -17,14 +17,15 @@
...
-			.visual-diff {
+			d2l-htmleditor {
 				margin-bottom: 30px;
 			}
 		</style>
 	</head>
 	<body class="d2l-typography">
 		<div class="visual-diff">
 			<d2l-htmleditor id="full"></d2l-htmleditor>
+			<d2l-htmleditor id="skeleton" skeleton="skeleton"></d2l-htmleditor>
```
dbatiste, line 28: I think it should be possible to just use 
```html
<d2l-htmleditor id="skeleton" skeleton></d2l-htmleditor>
```

dlockhart (approved):

test/htmleditor.visual-diff.js
grant-cleary: Ah, I see what you mean. I can get behind that. In the interests of unblocking Margaret I think I'll merge this as is now, but I'll start work on another set of changes to the tests as well.

PR #79, opened 2021-01-15
Add disabled state that renders html-block instead of tinymce.

dbatiste:


grant-cleary (approved):

dbatiste:
We're going to update this eventually to use some expand-collapse (mixin from core) or the more-less component, but there's still some design discussion that needs to happen to perhaps change the style of the more-less toggle.  Until then, I'll merge this to get this in, and we'll enhance soon for that.

PR #82, opened 2021-01-18
Switch to using label on web component instead of only updating the iframe title

grant-cleary:
These changes should help align the control with our other input controls. The way we're applying the label is a bit odd, because iframes aren't labelable. Done this way, screenreaders will ignore the span element written with the label in it, but will still read out the label from the iframe's title. The Interwebs were uncertain when it comes to whether it would be acceptable and/or better to use an actual `label` element here instead of a span, but from what I can see, the broad idea is "no, don't apply labels to unlabelable elements". I don't personally feel strongly about this though, so I'm open to changing that.

### DIFF
--- a/htmleditor.js
+++ b/htmleditor.js
@@ -29,6 +29,7 @@ import { css, html, LitElement, unsafeCSS } from 'lit-element/lit-element.js';
 import { addIcons } from './generated/icons.js';
 import { classMap } from 'lit-html/directives/class-map.js';
 import { getUniqueId } from '@brightspace-ui/core/helpers/uniqueId.js';
+import { inputLabelStyles } from '@brightspace-ui/core/components/inputs/input-label-styles.js';
 import { isfStyles } from './components/isf.js';
 import { Localizer } from './lang/localizer.js';
 import { ProviderMixin } from '@brightspace-ui/core/mixins/provider-mixin.js';
@@ -116,11 +117,12 @@ class HtmlEditor extends SkeletonMixin(ProviderMixin(Localizer(RtlMixin(LitEleme
 			fullPageFontSize: { type: String, attribute: 'full-page-font-size' },
 			height: { type: String },
 			html: { type: String },
+			label: { type: String },
+			labelHidden: { type: Boolean, attribute: 'label-hidden' },
 			mentions: { type: Boolean },
 			noFilter: { type: Boolean, attribute: 'no-filter' },
 			noSpellchecker: { type: Boolean, attribute: 'no-spellchecker' },
 			pasteLocalImages: { type: Boolean, attribute: 'paste-local-images' },
-			title: { type: String },
 			type: { type: String },
 			width: { type: String },
 			wordCountInFooter: { type: Boolean, attribute: 'word-count-in-footer' },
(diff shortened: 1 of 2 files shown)


htmleditor.js
```diff
@@ -116,11 +117,12 @@ class HtmlEditor extends SkeletonMixin(ProviderMixin(Localizer(RtlMixin(LitEleme
 			fullPageFontSize: { type: String, attribute: 'full-page-font-size' },
 			height: { type: String },
 			html: { type: String },
+			label: { type: String },
+			labelHidden: { type: Boolean, attribute: 'label-hidden' },
 			mentions: { type: Boolean },
 			noFilter: { type: Boolean, attribute: 'no-filter' },
 			noSpellchecker: { type: Boolean, attribute: 'no-spellchecker' },
 			pasteLocalImages: { type: Boolean, attribute: 'paste-local-images' },
-			title: { type: String },
```
grant-cleary, line 123: This makes this a breaking API change, but would also align us with other input elements that already use `label` and `labelHidden` to fulfill these requirements.

htmleditor.js
```diff
@@ -424,28 +427,22 @@ class HtmlEditor extends SkeletonMixin(ProviderMixin(Localizer(RtlMixin(LitEleme
 	render() {
 
 		if (this.disabled) {
-			const htmlBlockClasses = {
-				'd2l-skeletize': this.skeleton
-			};
 			return html`
-				<d2l-html-block class="${classMap(htmlBlockClasses)}">
+				<d2l-html-block class="d2l-skeletize">
```
grant-cleary, line 431: It turns out that the styles for `d2l-skeletize` are only applied when the skeleton attribute is present on the host, so we don't actually need to use class mapping for these. We can just always apply the class.

htmleditor.js
```diff
@@ -424,28 +427,22 @@ class HtmlEditor extends SkeletonMixin(ProviderMixin(Localizer(RtlMixin(LitEleme
...
-			'd2l-htmleditor-container': true,
-			'd2l-skeletize': this.skeleton
-		};
 
 		//if (this.type === editorTypes.INLINE || this.type === editorTypes.INLINE_LIMITED) {
 		//	return html`<div id="${this._editorId}" .innerHTML="${this._html}"></div>`;
 		//} else {
 		return html`
-			<div class="${classMap(containerClasses)}">
+		${this.label && !this.labelHidden ? html`<span class="d2l-input-label d2l-skeletize" aria-hidden="true">${this.label}</span>` : ''}
```
grant-cleary, line 444: The net result of this is to show a label (styled like our standard input labels) right above the editor component that's invisible to screenreaders (which rely on the iframe title).

htmleditor.js
dbatiste: Cool yeah, I noticed that as well.

dbatiste (approved):
Just need to update the readme for the new properties.  Be sure to indicate the label is required.

dbatiste (approved):
Oh wow... that is really gross eh?

grant-cleary:
> Oh wow... that is really gross eh?

... Yeah, it's pretty bad. Bit disappointing that there isn't a more pleasant way to format this. =/

PR #83, opened 2021-01-28
increase precedence of style override

mdulat:
Fixes https://trello.com/c/sYrL3tsh/303-new-html-editor-cannot-use-emoji-picker-math-editor-etc-in-the-expanded-html-editor, which states that in expanded view, the emoji, special character, and equation picker dialogs show up behind the editor. The css that is supposed to be overridden in the the tinymce skin styles, [e.g. this one](https://github.com/BrightspaceUI/htmleditor/blob/master/tinymce/skins/ui/oxide/skin.css#L1933), was not specific enough to take precedence over the skin.css one.

### DIFF
--- a/htmleditor.js
+++ b/htmleditor.js
@@ -163,7 +163,7 @@ class HtmlEditor extends SkeletonMixin(ProviderMixin(Localizer(RtlMixin(LitEleme
 				background: none; /* snow */
 			}
 			.tox-tinymce-aux,
-			.tox-tinymce.tox-fullscreen {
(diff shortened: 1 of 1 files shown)


dbatiste (approved):
This seems fine to me, but @grant-cleary has also been looking at fullscreen problems lately.

grant-cleary:
Yeah, I think more is broken than just the styling, based on some defects Jeannette found last week. I'll be looking into those this week, so if you find any other weirdness there, let me know.

PR #84, opened 2021-02-02
fix: find height parts without lookbehind assertions

mdulat:
The new editor does not display at all in mac/safari, with console error "Invalid regular expression: invalid group specifier name". Looks like it's because of [the Lookbehind assertion used here](https://github.com/BrightspaceUI/htmleditor/blob/master/htmleditor.js#L469) which are not supported by safari. This fix gets the height value and units without using these assertions.


dbatiste (approved):

grant-cleary (approved):
Ugh, Safari... Thanks for catching and fixing this!

PR #85, opened 2021-02-02
Updates to support FRA integration.

dbatiste:
This PR is the first of several PRs to support using the editor in FRAs.

* get context using ifrau if necessary
* override some tinyMCE styles for usage in FRA iframes

### DIFF
--- a/htmleditor.js
+++ b/htmleditor.js
@@ -61,7 +61,7 @@ const isShadowDOMSupported = !(window.ShadyDOM && window.ShadyDOM.inUse);
 let context;
 const getContext = () => {
 	if (context) return context;
-	context = new Promise(async (resolve) => {
+	context = new Promise(async resolve => { // eslint-disable-line no-async-promise-executor
 		if (window.ifrauclient) {
 			const ifrauClient = await window.ifrauclient().connect();
 			const ifrauEditorService = await ifrauClient.getService('htmleditor', '0.1');
--- a/htmleditor.js
+++ b/htmleditor.js
@@ -58,7 +58,20 @@ const editorTypes = {
 
 const isShadowDOMSupported = !(window.ShadyDOM && window.ShadyDOM.inUse);
 
-const context = JSON.parse(document.documentElement.getAttribute('data-he-context'));
+let context;
+const getContext = () => {
+	if (context) return context;
+	context = new Promise(async (resolve) => {
+		if (window.ifrauclient) {
+			const ifrauClient = await window.ifrauclient().connect();
+			const ifrauEditorService = await ifrauClient.getService('htmleditor', '0.1');
+			resolve(JSON.parse(await ifrauEditorService.getContext()));
+		} else {
+			resolve(JSON.parse(document.documentElement.getAttribute('data-he-context')));
+		}
+	});
+	return context;
+};
 
 const rootFontSize = window.getComputedStyle(document.documentElement, null).getPropertyValue('font-size');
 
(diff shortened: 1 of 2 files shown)


htmleditor.js
```diff
@@ -58,7 +58,20 @@ const editorTypes = {
 
 const isShadowDOMSupported = !(window.ShadyDOM && window.ShadyDOM.inUse);
 
-const context = JSON.parse(document.documentElement.getAttribute('data-he-context'));
+let context;
```
dbatiste, line 61: When the `ifrauclient` is detected we will use the new `htmleditor` service to get the context. 

This `getContext` method returns a promise for the context because the ifrau service methods are async. It is designed to get the context once regardless of how many editors are on the page.

htmleditor.js
```diff
@@ -179,6 +193,15 @@ class HtmlEditor extends SkeletonMixin(ProviderMixin(Localizer(RtlMixin(LitEleme
 			.tox-tinymce.tox-fullscreen .tox-statusbar__resize-handle {
 				display: none;
 			}
+			:host([fra-context]:not(.tox-fullscreen)) {
```
dbatiste, line 196: tinyMCE uses `position: fixed`, `display: flex`, and `align-items: center` to position it's dialogs vertically in the center of the view-port.  However, FRA iframes can be very tall because iframe resizer sizes the iframe to fit the contents allowing the host page to control the scrolling.  This results in the tinyMCE dialog being positioned out of view (ex. further down the page). This sol'n relies on this behaviour of fixed positioning to cause the dialog to be positioned above the editor component:

> It is positioned relative to the initial containing block established by the viewport, except when one of its ancestors has a transform, perspective, or filter property set to something other than none (see the CSS Transforms Spec), in which case that ancestor behaves as the containing block.

htmleditor.js
```diff
@@ -203,27 +226,12 @@ class HtmlEditor extends SkeletonMixin(ProviderMixin(Localizer(RtlMixin(LitEleme
 		this.width = '100%';
 		this.wordCountInFooter = false;
 		this._editorId = getUniqueId();
+		this._fraContext = !!window.ifrauclient;
 		this._html = '';
 		this._initializationComplete = new Promise((resolve) => {
 			this._initializationResolve = resolve;
 		});
 		this._uploadImageCount = 0;
-		if (context) {
```
dbatiste, line 211: The `provideInstance` methods are moved because getting the context is now asynchronous and we therefore need to `await` it, however `constructor` methods cannot be `async` (because they have to return an object by def'n, not a Promise). So, this block of code is moved down to `firstUpdated` where we can wait for the context before doing any tinyMCE initialization.

htmleditor.js
```diff
@@ -441,11 +465,11 @@ class HtmlEditor extends SkeletonMixin(ProviderMixin(Localizer(RtlMixin(LitEleme
 		//	return html`<div id="${this._editorId}" .innerHTML="${this._html}"></div>`;
 		//} else {
 		return html`
-		${this.label && !this.labelHidden ? html`<span class="d2l-input-label d2l-skeletize" aria-hidden="true">${this.label}</span>` : ''}
+			${this.label && !this.labelHidden ? html`<span class="d2l-input-label d2l-skeletize" aria-hidden="true">${this.label}</span>` : ''}
```
dbatiste, line 468: Formatting only.

index.html
```diff
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html lang="en">
 <head>
-	<link rel="stylesheet" href="components/demo/styles.css" type="text/css">
+	<link rel="stylesheet" href="/node_modules/@brightspace-ui/core/components/demo/styles.css" type="text/css">
```
dbatiste, line 4: This fixes 404 errors on the index.html page.

grant-cleary (approved):

PR #86, opened 2021-02-03
DE42315 - a11ychecker More Info links should point to WCAG documentation instead of TinyMCE's docs

grant-cleary:
At some point between the a11ychecker version we had in the old editor and this new one, TinyMCE updated their More Info links to point to their documentation instead of WCAG's (presumably so it was easier for them to keep the docs up to date). We would like to go back to just using WCAG documentation. Thankfully, TinyMCE offers up a way to customize this now. Unfortunately, this does require that we keep an internal mapping of WCAG rules to the actual documentation URLs.

### DIFF
--- a/htmleditor.js
+++ b/htmleditor.js
@@ -305,6 +305,29 @@ class HtmlEditor extends SkeletonMixin(ProviderMixin(Localizer(RtlMixin(LitEleme
 
 			tinymce.init({
 				a11ychecker_allow_decorative_images: true,
+				a11ychecker_issue_url_callback: ruleId => {
+					const ruleUrlMap = {
+						'D1': 'https://www.w3.org/WAI/WCAG21/Techniques/html/H42.html',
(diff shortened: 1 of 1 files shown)


htmleditor.js
```diff
@@ -305,6 +305,29 @@ class HtmlEditor extends SkeletonMixin(ProviderMixin(Localizer(RtlMixin(LitEleme
 
 			tinymce.init({
 				a11ychecker_allow_decorative_images: true,
+				a11ychecker_issue_url_callback: ruleId => {
+					const ruleUrlMap = {
+						'D1': 'https://www.w3.org/WAI/WCAG21/Techniques/html/H42.html',
```
grant-cleary, line 310: Worth noting: This rule is a bit ambiguous, in the sense that TinyMCE actually links out to [two different sources in their docs](https://www.tiny.cloud/docs/plugins/premium/a11ychecker/#a11ycheckerruledetails-d1). This one seemed the most relevant for now, but it's something we might want to run by Sam.

dbatiste (approved):

PR #88, opened 2021-02-04
lms adapter

dbatiste:
This PR introduces an adapter for integration in the LMS either directly or via `ifrau`. It also moves the context resolution from `htmleditor.js`, and updates the ISF plugin to use it for launching the legacy dialog.

### DIFF
--- a/components/lms-adapter.js
+++ b/components/lms-adapter.js
@@ -1,0 +1,76 @@
+
+let contextPromise;
+
+export function getContext() {
+	if (contextPromise) return contextPromise;
+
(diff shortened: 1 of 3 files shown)


components/lms-adapter.js
```diff
@@ -0,0 +1,76 @@
...
+		} else {
+			resolve(JSON.parse(document.documentElement.getAttribute('data-he-context')));
+		}
+	});
+
+	return contextPromise;
+}
+
+export function hasLmsContext() {
+	return D2L.LP || window.ifrauclient;
```
grant-cleary, line 21: Nice!

grant-cleary (approved):

PR #89, opened 2021-02-04
DE42313 - Don't query outdent menu button state on select (it always returns true if the content can be outdented)

grant-cleary:
The outdent command actually has a valid state response to `queryCommandState`... Specifically, it returns true iff the current selected content (or the location of the caret) can be outdented. This leads to the outdent option always showing as active in our split button menu, provided any indentation has been applied (either through explicit indentation or one of the list options).

### DIFF
--- a/htmleditor.js
+++ b/htmleditor.js
@@ -394,7 +394,7 @@ class HtmlEditor extends SkeletonMixin(ProviderMixin(Localizer(RtlMixin(LitEleme
 							tooltip: tooltip,
 							onAction: () => editor.execCommand(cmd),
 							onItemAction: (api, value) => editor.execCommand(value),
-							select: value => editor.queryCommandState(value),
+							select: value => value !== 'outdent' && editor.queryCommandState(value),
 							fetch: callback => callback(items)
 						});
 					};


htmleditor.js
```diff
@@ -394,7 +394,7 @@ class HtmlEditor extends SkeletonMixin(ProviderMixin(Localizer(RtlMixin(LitEleme
 							tooltip: tooltip,
 							onAction: () => editor.execCommand(cmd),
 							onItemAction: (api, value) => editor.execCommand(value),
-							select: value => editor.queryCommandState(value),
+							select: value => value !== 'outdent' && editor.queryCommandState(value),
```
grant-cleary, line 397: I find it annoying that TinyMCE assumes that we're going to want to apply the same `select` function to every item in the menu.

dbatiste:
Interesting.  So `queryCommandState('indent')` will return true when the text _is_ indented, but `queryCommandState('outdent')` returns true when the text _can be_ outdented?  I suppose it makes sense that one can't query for an `outdent` state, because the text is either not indented, or it is.

dbatiste (approved):

grant-cleary:
> Interesting. So `queryCommandState('indent')` will return true when the text _is_ indented, but `queryCommandState('outdent')` returns true when the text _can be_ outdented? I suppose it makes sense that one can't query for an `outdent` state, because the text is either not indented, or it is.

I was poking around at the internal state mapping TinyMCE uses to determine what `queryCommandState` returns, and it actually looks like `indent` doesn't have a mapping in that table (and thus `queryCommandState('indent')` always returns false). In contrast, the mapping for outdent calls a function `canOutdent` which does what you would think it does.

It looks like this is used to figure out whether to enable the toolbar button when it's not used in a split-button, since in their demo, the `outdent` button is disabled when nothing has indented the text, while the indent button is always enabled. Sure enough, when I indent (or manually delete indentation), I can see `queryCommandState('outdent')` being invoked before the button is enabled/disabled. Not exactly the behaviour I would have expected, but 🤷, I guess it does make sense in hindsight.

PR #93, opened 2021-02-05
Switch colours and background colours to align with WCAG criteria (and Brightspace design in general)

grant-cleary:
This... Isn't great. This effectively functions as a stop-gap for now while we figure out how best to address these issues. If this is still too gross for us to do even in the short-term, we can also make the decision on whether we want a slightly higher-effort solution.


htmleditor.js
```diff
@@ -150,6 +150,40 @@ class HtmlEditor extends SkeletonMixin(ProviderMixin(Localizer(RtlMixin(LitEleme
 			.tox .tox-toolbar__group {
 				padding: 0 4px 0 8px; /* snow */
 			}
+			/* Contrast a11y corrections for Snow skin */
+			.tox .tox-collection--list .tox-collection__item--active:not(.tox-collection__item--state-disabled) {
+				color: var(--d2l-color-celestine);
+			}
+			.tox .tox-collection--list .tox-collection__item--active:not(.tox-collection__item--state-disabled) * {
+				color: var(--d2l-color-celestine) !important;
+				text-decoration-color: var(--d2l-color-celestine) !important;
```
grant-cleary, line 159: The formatting menu uses its actual relevant elements (e.g. `<h1>` for "Heading 1"), but these are all hard-coded with an element style that sets them to be coloured black regardless of focus, hover, etc.

grant-cleary:
Looks like we're going to use a different skin instead, so I'm going to close this out.

PR #94, opened 2021-02-09
Update insert image to use the lms-adapter.

dbatiste:
This change needs to be made in conjunction with LMS change.

### DIFF
--- a/components/lms-adapter.js
+++ b/components/lms-adapter.js
@@ -55,7 +55,7 @@ export async function openDialogWithParam(opener, location, params, settings) {
 
 	} else {
 
-		const result = await (new Promise(resolve => {
+		return new Promise(resolve => {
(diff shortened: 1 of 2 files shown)


components/lms-adapter.js
```diff
@@ -55,7 +55,7 @@ export async function openDialogWithParam(opener, location, params, settings) {
 
 	} else {
 
-		const result = await (new Promise(resolve => {
+		return new Promise(resolve => {
```
dbatiste, line 58: Just cleaning tidying up the promise.

components/lms-adapter.js
```diff
@@ -100,7 +99,7 @@ export async function openLegacyDialog(opener, location, settings) {
...
```
dbatiste, line 102: Just tidying up the promise.

grant-cleary (approved):

PR #95, opened 2021-02-10
explicitly add required styles to host

mdulat:
The emoji and special char picker is cut off on safari when viewed in fullscreen mode:

mac3

The reason is because [the skin styles](https://github.com/BrightspaceUI/htmleditor/blob/master/tinymce/skins/ui/snow/skin.shadowdom.css) are not being applied to the htmleditor component when displayed within a FACE page. An ideal fix would allow these styles to be applied, but I have not been able to figure out how to do that, so am explicitly adding the styles that I know are required to fix this bug.

### DIFF
--- a/htmleditor.js
+++ b/htmleditor.js
@@ -140,7 +140,7 @@ class HtmlEditor extends SkeletonMixin(ProviderMixin(Localizer(RtlMixin(LitEleme
 				position: fixed;
 			}
 			:host(.tox-shadowhost.tox-fullscreen) {
-				z-index: 1200;
+				z-index: 1000;
 			}
 			.d2l-htmleditor-container {
 				border: 1px solid var(--d2l-color-mica); /* snow */
--- a/htmleditor.js
+++ b/htmleditor.js
@@ -135,6 +135,12 @@ class HtmlEditor extends SkeletonMixin(ProviderMixin(Localizer(RtlMixin(LitEleme
 			:host([hidden]) {
 				display: none;
 			}
+			:host(.tox-fullscreen) {
+				position: fixed;
+			}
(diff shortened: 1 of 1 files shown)


htmleditor.js
```diff
@@ -135,6 +135,13 @@ class HtmlEditor extends SkeletonMixin(ProviderMixin(Localizer(RtlMixin(LitEleme
 			:host([hidden]) {
 				display: none;
 			}
+			/* stylelint-disable selector-class-pattern */
+			:host(.tox-fullscreen) {
+				position: fixed;
+			}
+			:host(.tox-shadowhost.tox-fullscreen) {
+				z-index: 1200;
+			}
```
grant-cleary, line 144: If it turns out we need to do this, that's fine, but I would prefer poking around a little bit to see if there's a solution to the underlying issue. This sort of stuff can end up being pretty fragile when we inevitably need to update TinyMCE, since these stylings could arbitrarily change (TinyMCE keeps no contract around its CSS classes, with very rare exceptions).

dbatiste: We already override some tinyMCE style for fullscreen, I think in order to make the stacking order compatible with our other components.  I don't recall the exact use case, but I seem to recall there being a very narrow margin of z-index that will work. https://github.com/BrightspaceUI/htmleditor/blob/master/htmleditor.js#L163

grant-cleary: Yeah, I mostly just find it weird that this specific case works for the monolith editor, but not for FACE. That's the part that's concerning to me... I would expect this behaviour to be consistent. It sounds like the `z-index` isn't even being applied when this isn't explicitly included in the component CSS, which worries me.

grant-cleary (approved):

PR #96, opened 2021-02-10
US120090 - Add functional support for inline/inline-limited modes to editor

grant-cleary:
This introduces the functional capacity for inline/inline-limited modes within the editor proper. How the toolbars render currently doesn't match design spec, but that will come in a future PR.


htmleditor.js
```diff
@@ -388,6 +403,11 @@ class HtmlEditor extends SkeletonMixin(ProviderMixin(Localizer(RtlMixin(LitEleme
 								bubbles: true
 							}
 						));
+
+						if (this.type === editorTypes.INLINE || this.type === editorTypes.INLINE_LIMITED) {
+							this._isEditing = false;
```
dbatiste, line 408: What's the gist of these here - won't these already be initialized?

htmleditor.js
```diff
@@ -123,12 +125,14 @@ class HtmlEditor extends SkeletonMixin(ProviderMixin(Localizer(RtlMixin(LitEleme
 			width: { type: String },
 			wordCountInFooter: { type: Boolean, attribute: 'word-count-in-footer' },
 			_editorId: { type: String },
-			_fraContext: { type: Boolean, attribute: 'fra-context', reflect: true }
+			_fraContext: { type: Boolean, attribute: 'fra-context', reflect: true },
+			_isEditing: { type: Boolean, attribute: 'is-editing' },
+			_isHtmlBlockButtonFocusing: { type: Boolean, attribute: 'is-html-block-button-focusing' }
```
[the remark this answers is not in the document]
grant-cleary: Ah of course, I only need this to ensure we re-render on the property change, but it doesn't need to bind the attribute to do that. Good thought on renaming the variable. I'll do that.

htmleditor.js
grant-cleary: `this._isEditing` is flipped to `true` when focus is placed into the editor. Flipping it to false in the blur handler ensures that we re-render and trigger the CSS changes to flip back into read-only mode. `this._html` may not be reflective of what's currently in the editor when we do that focus switch, since it's only initialized when the editor is first loaded. If the user has entered content since the editor was first initialized, that won't be reflected in the property. So in this case, when I switch into readonly mode, I set the `this._html` to reflect the current state of the editor so that consumers that use it (incl. the readonly HTML block) get the most up-to-date content.

dbatiste: I see.  Makes sense.  Thanks.

dbatiste (approved):

PR #97, opened 2021-02-11
Switch TinyMCE to a new custom skin instead of the existing Snow skin

grant-cleary:
As per Design, we're going to drop the Snow skin in favour of a custom skin generated through TinyMCE's skinning tool (to bypass existing a11y issues with the Snow skin).


htmleditor.js
```diff
@@ -142,29 +142,12 @@ class HtmlEditor extends SkeletonMixin(ProviderMixin(Localizer(RtlMixin(LitEleme
...
-			}
-			.tox .tox-pop__dialog .tox-toolbar-nav-js {
-				margin-bottom: 0; /* snow */
-				margin-top: 0; /* snow */
-				min-height: auto; /* snow */
-				padding: 0; /* snow */
-			}
-			.tox .tox-pop__dialog .tox-toolbar {
-				background: none; /* snow */
-			}
```
grant-cleary, line 167: Things seem to look fine without any of these adjustments, but I'm also not 100% sure what they were all implemented for. My thinking is that we can drop them for now and then run this by Noah/Jeff again to see what (if anything) we want to customize with CSS overrides. That said, if I'm missing context on why we need these, then let me know and I'll add them back as appropriate. :)

htmleditor.js
```diff
@@ -142,29 +142,12 @@ class HtmlEditor extends SkeletonMixin(ProviderMixin(Localizer(RtlMixin(LitEleme
 			:host(.tox-shadowhost.tox-fullscreen) {
 				z-index: 1000;
 			}
-			.d2l-htmleditor-container {
-				border: 1px solid var(--d2l-color-mica); /* snow */
-				border-radius: 6px;
-				padding: 4px; /* snow */
-			}
```
grant-cleary, line 149: I imagine we still want the editor border to use the mica colour, but that also might look a bit weird with the overall look of the skin. I'll check it out.

grant-cleary: Oh, maybe not actually. I didn't realize how close mica already is to this colour. Never mind, I think this has already been accounted for in the skin.

d2l-skin-params.json
```diff
@@ -0,0 +1,39 @@
+{
```
grant-cleary, line 1: Not necessarily sure if we want this to live here, but I think we should keep it in source control so that it's easy to modify the skin in the future.

dbatiste: Maybe we should store this in the `tinymce` folder and call tis `skin-params.json`?

grant-cleary: The reason I wasn't sure about that is because right now we include the entire `tinymce` folder in the NPM package. My initial thought is that we don't want/need this to be bundled with everything else, though we could also work around that by just specifying exactly which directories in `tinymce` we want published. I wasn't sure whether we would want to go that route though, since it could mean we would need to change what we publish if TinyMCE changes that folder structure.

d2l-skin-params.json
```diff
@@ -0,0 +1,39 @@
...
+  "backgroundColor": "#ffffff",
+  "skinName": "D2L",
+  "contentCss": "body {\n  font-family: -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Oxygen, Ubuntu, Cantarell, 'Open Sans', 'Helvetica Neue', sans-serif;\n  line-height: 1.4;\n  margin: 1rem;\n}\n\ntable {\n  border-collapse: collapse;\n}\n\ntable th,\ntable td {\n  border: 1px solid #ccc;\n  padding: .4rem;\n}\n\nfigure {\n  display: table;\n  margin: 1rem auto;\n}\n\nfigure figcaption {\n  color: #999;\n  display: block;\n  margin-top: .25rem;\n  text-align: center;\n}\n\nhr {\n  border-color: #ccc;\n  border-style: solid;\n  border-width: 1px 0 0 0;\n}\n\ncode {\n  background-color: #e8e8e8;\n  border-radius: 3px;\n  padding: .1rem .2rem;\n}\n\n.mce-content-body:not([dir=rtl]) {\n  blockquote {\n    border-left: 2px solid #ccc;\n    margin-left: 1.5rem;\n    padding-left: 1rem;\n  }\n}\n\n// RTL\n.mce-content-body[dir=rtl] {\n  blockquote {\n    border-right: 2px solid #ccc;\n    margin-right: 1.5rem;\n    padding-right: 1rem;\n  }\n}\n",
```
dbatiste, line 4: Are these fonts intentional?  I also see a font file being included - is it actually needed?

d2l-skin-params.json
```diff
@@ -0,0 +1,39 @@
...
+    "@color-black": "rgba(0, 0, 0, 1)",
+    "@color-white": "#fff",
+    "@color-tint": "rgba(0, 111, 191, 1)",
+    "@color-error": "rgba(205, 32, 38, 1)",
+    "@color-success": "rgba(70, 166, 97, 1)",
+    "@base-value": "16px",
+    "@control-border-radius": "5px",
+    "@panel-border-radius": "5px",
+    "@tinymce-border-radius": "5px",
+    "@font-stack": "-apple-system,BlinkMacSystemFont,\"Segoe UI\",Roboto,Oxygen-Sans,Ubuntu,Cantarell,\"Helvetica Neue\",sans-serif",
```
dbatiste, line 20: Here too - are we sure about this?

d2l-skin-params.json
grant-cleary: Unknown. This was what Noah sent to me. I assumed it was good, but now I'm less sure because I didn't really notice the font additions. I'll remove them and see if anything horrific breaks.

d2l-skin-params.json
```diff
@@ -0,0 +1,39 @@
...
+    "@toolbar-background-color": "@background-color",
+    "@menubar-background-color": "@background-color",
+    "@border-color": "rgba(227, 233, 241, 1)",
+    "@color-black": "rgba(0, 0, 0, 1)",
+    "@color-white": "#fff",
+    "@color-tint": "rgba(0, 111, 191, 1)",
+    "@color-error": "rgba(205, 32, 38, 1)",
+    "@color-success": "rgba(70, 166, 97, 1)",
+    "@base-value": "16px",
+    "@control-border-radius": "5px",
```
dbatiste, line 17: Is this the border radius that is being used to replace the `6px` that we were using before?

grant-cleary: It's annoyingly hard to tell. Changing these border-radius values in their skinning tool doesn't seem to do much. I dislike this tool.

d2l-skin-params.json
dbatiste: Also, we probably don't want to publish this file to NPM so I think we should to change the `package.json` `files` to...

```json
  "files": [
    "htmleditor.js",
    "/components",
    "/generated",
    "/lang",
    "/tinymce",
    "/wordcount/wordcount.js",
    "!/tinymce/skin-params.json"
  ],
```

d2l-skin-params.json
grant-cleary: Ok, it seems doubtful that this is necessary to put into the skinning tool, but it also produces output that's identical to the other two skins (at least for the fonts). What I'm less keen on is that using their skinning tool produces content that goes into `skins/content/d2l` as well, whereas it looks like for both the oxide and snow cases we were just using TinyMCE's defaults. I'm going to try removing those to see if there's any impact there.

d2l-skin-params.json
grant-cleary: Yes, if we moved this into the `tinymce` folder, we would need to do this. I can do that (move it in and add this exclusion).

d2l-skin-params.json
grant-cleary: I think I see what's happening here. Because we override TinyMCE's content CSS anyways, this ends up being irrelevant. I'm pretty sure this would only apply if we weren't specifying `content_css` in TinyMCE's init. I'll remove this regardless because it's redundant and we can just let TinyMCE fall back to the defaults since we know they won't be used anyways. I'll also update the README to indicate that we can just ignore that file.

grant-cleary: As for what's in the actual skinning tool JSON... The tool doesn't seem to display things properly if the `font-stack` property isn't set. I assume it's mandatory even though we don't use its output. Sigh.

tinymce/skin-params.json
```diff
@@ -0,0 +1,39 @@
+{
+  "backgroundColor": "#ffffff",
+  "skinName": "d2l",
```
grant-cleary, line 3: I lowercase'd this to make it work out with the standard casing of our directories; otherwise the skin tool creates a folder called "D2L" which may or may not overwrite "d2l" depending on OS.

tinymce/skin-params.json
```diff
@@ -0,0 +1,39 @@
...
+    "@color-black": "rgba(0, 0, 0, 1)",
+    "@color-white": "#fff",
+    "@color-tint": "rgba(0, 111, 191, 1)",
+    "@color-error": "rgba(205, 32, 38, 1)",
+    "@color-success": "rgba(70, 166, 97, 1)",
+    "@base-value": "16px",
+    "@control-border-radius": "5px",
+    "@panel-border-radius": "5px",
+    "@tinymce-border-radius": "5px",
+    "@font-stack": "-apple-system,BlinkMacSystemFont,\"Segoe UI\",Roboto,Oxygen-Sans,Ubuntu,Cantarell,\"Helvetica Neue\",sans-serif",
```
grant-cleary, line 20: If this is left empty, their skinning tool breaks. We don't actually implement this now that I've nuked `/tinymce/skins/content/d2l` and specified that we should only include the `ui` folder in the README, so in this case I think it's okay.

d2l-skin-params.json
dbatiste: 👍 Sounds good.

dbatiste (approved):

PR #99, opened 2021-02-12
make sure picker dialog is always on top in Safari

mdulat:
The latest issue in Safari looks like this:

Safari_picker_cutoff

It is when the editor is NOT in fullscreen mode. TinyMCE sets these picker positions as element styles, so the only way to overwrite that is to use `!important` as far as I can tell. Though maybe there is a better way if one digs deeper into how TinyMCE decides what to set it to.


dbatiste (approved):
Thanks Margaret.  Maybe Grant or I can inquire with tinyMCE about this behavior and show them what we sometimes observe we can get an explanation, and perhaps a fix from them.

PR #106, opened 2021-02-16
Update TinyMCE to 5.7.0 and get rid of now-defunct fullscreen background color hack

grant-cleary:
There are some fixes we really want in 5.7.0:
- Loss-of-focus fix in symbols/emoticons dialog search bar
- Transparent fullscreen background causing grief
- Unexpected event propagation when using ESC to close an open dialog

I didn't notice anything untoward during some light testing, but we're going to be doing a bit of a deeper dive on testing the editor, so it will be interesting to see if we catch anything surprising.


htmleditor.js
```diff
@@ -150,7 +150,6 @@ class HtmlEditor extends SkeletonMixin(ProviderMixin(Localizer(RtlMixin(LitEleme
 			}
 			.tox-tinymce-aux,
 			.tox.tox-tinymce.tox-fullscreen {
-				background-color: #ffffff;
```
grant-cleary, line 153: No longer necessary since the container background-color appears to default to white now.

tinymce/skins/ui/oxide-dark/content.css
```diff
@@ -322,11 +322,15 @@ pre[class*="language-"] {
 .mce-content-body div.mce-resizehandle:nth-of-type(4) {
```
grant-cleary, line 322: In general, all of these oxide and oxide-dark skin changes are largely irrelevant to us, since we're not using the oxide or oxide-dark skins. That said, I figure we may wish to keep them up to date regardless in case we ever change our minds on that.

dbatiste (approved):
I downloaded and updated the tinyMCE lang files.  Not sure why GitHub seems to think they are empty when looking at the diff in the PR.  They look correct I think, when looking at the files in this branch.

grant-cleary:
> I downloaded and updated the tinyMCE lang files. Not sure why GitHub seems to think they are empty when looking at the diff in the PR. They look correct I think, when looking at the files in this branch.

Ah, thanks Dave. I forgot about the lang files.

PR #108, opened 2021-02-20
Inject the fonts.css file into fullpage document if lato is being used.

dbatiste:
This required a lot of playing around, trying to get the various scenarios working. It turned out to be not too bad, but the process of arriving at this code was a little frustrating. This code effectively ensures that when we get the HTML from the editor, it will contain the [Lato stylesheet](https://s.brightspace.com/lib/fonts/0.5.0/fonts.css) reference if the HTML contains a reference to Lato.  I think this change covers the cases where:

* user selects Lato from toolbar dropdown
* user pastes content containing Lato
* user starts a new document with the full page font specifying Lato
* user adds Lato font in source code view

The content that is saved is updated via the `d2l-fullpage` plugin that wires up to tinyMCE's `GetContent` event.  The design view shows Lato if used because the stylesheet is always included in the tinyMCE content styles configuration.

This does not detect the scenario where Lato is used in an external stylesheet, but I think in that case, that external stylesheet should just include an `@import` statement to include the font.

**Important**: by allowing the user to select Lato specifically, we are effectively committing to supporting Lato forever in user-authored content. If we ever change fonts, we will have a new fonts.css url (ex. 0.6.0), we would be adding another check like this one (not removing this one).

### DIFF
--- a/components/equation.js
+++ b/components/equation.js
@@ -1,5 +1,5 @@
 /* eslint no-useless-escape: 0 */
-
+import 'tinymce/tinymce.js';
 import { css, LitElement } from 'lit-element/lit-element.js';
 import { hasLmsContext, openDialogWithParam } from './lms-adapter.js';
 import { getComposedActiveElement } from '@brightspace-ui/core/helpers/focus.js';
--- a/components/fullpage.js
+++ b/components/fullpage.js
@@ -1,21 +1,23 @@
 import 'tinymce/tinymce.js';
 
+const latoRegEx = /font-family: lato/i;
+
 tinymce.PluginManager.add('d2l-fullpage', function(editor) {
 
-	if (!editor.settings.fullpage_default_font_family.toLowerCase().includes('lato')) return;
-
+	// specifically pull in 0.5.0 containing Lato font
 	const fontsUrl = 'https://s.brightspace.com/lib/fonts/0.5.0/fonts.css';
 
-	editor.on('SetContent', () => {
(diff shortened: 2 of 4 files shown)


components/equation.js
```diff
@@ -1,5 +1,5 @@
 /* eslint no-useless-escape: 0 */
-
+import 'tinymce/tinymce.js';
```
dbatiste, line 2: This was just a missing import that I uncovered as a result of sorting the imports in htmleditor.js

components/fullpage.js
```diff
@@ -0,0 +1,23 @@
+import 'tinymce/tinymce.js';
+
+tinymce.PluginManager.add('d2l-fullpage', function(editor) {
```
dbatiste, line 3: Sadly, tinymce does not provide a config to support this automatically.  This is something we should ask them to add for the fullpage plugin.

htmleditor.js
```diff
@@ -79,6 +80,10 @@ if (!tinymceLangs.includes(documentLang)) {
 const pathFromUrl = url => url.substring(0, url.lastIndexOf('/'));
 const baseImportPath = pathFromUrl(import.meta.url);
 
+const fullPageStyles = css`
```
dbatiste, line 83: This includes the Lato font in design view regardless of whether Lato is specified as the fullpage font, because the user can select Lato from the font selection in the toolbar, or they could also specify Lato in the source code view.

grant-cleary (approved):

PR #111, opened 2021-02-22
Add ifrau pass-through for ImageTools image fetching

grant-cleary:
Paired with [this LMS PR](https://github.com/Brightspace/lms/pull/6848) - Allows us to fetch images via ifrau when using ImageTools in FRAs.


components/image.js
```diff
@@ -19,6 +19,33 @@ class FileData {
...
+		const fileName = path.replace('/shared/', '');
+		return await getSharedFile(src, fileName);
+	} else if (path.startsWith('/d2l/lp/files/temp/')) {
+		const fileId = path.replace('/d2l/lp/files/temp/', '').replace('/View', '');
+		return await getTempFile(src, fileId);
+	} else {
+		return window.fetch(src).then((response) => {
+			if (!response.ok) return new Blob();
+			return response.blob();
+		});
```
grant-cleary, line 45: This should cover external resources. If someone links out to an image outside of the LMS, TinyMCE still lets them try to edit it. Of course, if it's cross-origin and requires cookies or something, this will just fail... But it would fail even if we let TinyMCE do its own thing, so :shrug:.

htmleditor-mixin.js
```diff
@@ -311,6 +311,13 @@ export const HtmlEditorMixin = superclass => class extends Localizer(RtlMixin(Pr
 			documentConfig.document_base_url = this._context.host;
 		}
 
+		const imageToolsConfig = {};
+		if (this._fraContext) {
```
grant-cleary, line 315: This is technically unnecessary, since the `getImage` fallbacks already cover this case. We would just be doing the `fetch` to get the image instead of TinyMCE. In that regard, I'm also okay if we omit this check and just always include this function. I guess it depends on whether we want to delegate to TinyMCE in the case that we're not in an FRA. :)

components/image.js
```diff
@@ -19,6 +19,33 @@ class FileData {
...
+	// bail if no LMS context (local image fetching relies on LMS context for now)
+	if (!hasLmsContext()) return;
+
+	const orgUnitPath = requestInstance(editor, 'orgUnitPath');
+	const host = requestInstance(editor, 'host');
+
+	const path = host ? new URL(src, host).pathname : new URL(src).pathname;
+
+	if (path.startsWith(orgUnitPath)) {
+		const fileName = path.replace(orgUnitPath, '');
```
dbatiste, line 33: If the file is within nested folders on Content, this won't really be just the file name per se.  I think that's fine for this code, but the way the variable is named... just want to make sure that's what you're expecting.

grant-cleary: Mm, true. I didn't think of that case. I _think_ in this case it still works out fine, because the LMS side of things basically spits out `{orgUnitPath}{fileName}` at the moment (so the fragment will still be correct). I'll test to verify, but either way I'll change the variable name to be something more accurate.

components/image.js
```diff
@@ -19,6 +19,33 @@ class FileData {
 	}
 }
 
+export async function getImage(editor, src) {
```
dbatiste, line 22: Note: I don't think this `async` (or the `await`s) below are necessary.  For example, in the code below, it can just be:
```javascript
return getContentFile(src, orgUnitPath, fileName);
```
Because `getContentFile` (and everything else) is already returning a Promise.

grant-cleary: Yeah, I think you're right. I got lost in the async/await weeds trying to debug this stuff, so I imagine I added some where I don't need them.

components/lms-adapter.js
```diff
@@ -158,3 +158,65 @@ export async function uploadFile(orgUnitId, fileName, file, maxFileSize) {
...
+
+export async function getContentFile(src, orgUnitPath, fileName) {
+	if (window.ifrauclient) {
+
+		if (!fetchFilesService) {
+			const ifrauClient = await window.ifrauclient().connect();
+			fetchFilesService = await ifrauClient.getService('fetch-files', '0.1');
+		}
+
+		return await fetchFilesService.getContentFile(orgUnitPath, fileName);
```
dbatiste, line 172: Almost the same thing here... this can just be:
```javascript
return fetchFilesService.getContentFile(orgUnitPath, fileName);
```
Because I think `fetchFilesService` is returning a Promise.  The `async` is still needed though since we need to await the ifrau stuff above.

components/lms-adapter.js
```diff
@@ -158,3 +158,65 @@ export async function uploadFile(orgUnitId, fileName, file, maxFileSize) {
...
+
+export async function getSharedFile(src, fileName) {
+	if (window.ifrauclient) {
+
+		if (!fetchFilesService) {
+			const ifrauClient = await window.ifrauclient().connect();
+			fetchFilesService = await ifrauClient.getService('fetch-files', '0.1');
+		}
+
+		return await fetchFilesService.getSharedFile(fileName);
```
dbatiste, line 192: Ditto.

components/lms-adapter.js
```diff
@@ -158,3 +158,65 @@ export async function uploadFile(orgUnitId, fileName, file, maxFileSize) {
...
+
+export async function getTempFile(src, fileId) {
+	if (window.ifrauclient) {
+
+		if (!fetchFilesService) {
+			const ifrauClient = await window.ifrauclient().connect();
+			fetchFilesService = await ifrauClient.getService('fetch-files', '0.1');
+		}
+
+		return await fetchFilesService.getTempFile(fileId);
```
dbatiste, line 212: Ditto.

components/image.js
grant-cleary: Kind of weird, but after removing the `async` keyword here, VS Code really wants me to add it again. Seems fine to just remove it and return the Promise though.

dbatiste (approved):

dbatiste (approved):

PR #112, opened 2021-02-24
Ensure focus goes to content iframe after D2L dialogs are closed

grant-cleary:
Dependent on this LMS PR: https://github.com/Brightspace/lms/pull/7021

This PR adds logic to our custom dialogs to put focus back on the editor after they're closed. We also need to specify the `byPassOpenerFocus` setting on the dialogs when we open them, otherwise when the dialog close event finishes processing (after our code has already executed), focus gets placed back on the opener (which is the toolbar button).

### DIFF
--- a/components/equation.js
+++ b/components/equation.js
@@ -50,11 +50,13 @@ tinymce.PluginManager.add('d2l-equation', function(editor) {
 			dialog.mathML = decodeURIComponent(contextNode.attributes.getNamedItem('data-d2l-mathml').value);
 		}
 
+		if (editor.selection) dialog.bookmark = editor.selection.getBookmark();
 		dialog.type = editorType;
 		dialog.opened = true;
 		dialog.addEventListener('d2l-htmleditor-equation-dialog-close', (e) => {
 			const html = e.detail.html;
 			if (html) editor.execCommand('mceInsertContent', false, html);
+			root.host.focus();
 		}, { once: true });
 
 	};
@@ -236,7 +238,8 @@ class EditorDialog extends LitElement {
 			const result = await openDialogWithParam(
 				getComposedActiveElement(),
 				'/d2l/lp/math/createeditor',
-				{ mathml: this.mathML, editorType: this.type }
+				{ mathml: this.mathML, editorType: this.type },
+				{ byPassOpenerFocus: true }
 			);
 
 			this.opened = false;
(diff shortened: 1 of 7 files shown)


components/equation.js
```diff
@@ -55,6 +55,7 @@ tinymce.PluginManager.add('d2l-equation', function(editor) {
 		dialog.addEventListener('d2l-htmleditor-equation-dialog-close', (e) => {
 			const html = e.detail.html;
 			if (html) editor.execCommand('mceInsertContent', false, html);
+			root.host.focus();
```
grant-cleary, line 58: I can confirm that this respects the editor's bookmarking and selection state, so this behaves as you would expect. It would have been easier here to just invoke `editor.focus()` and accomplish the same thing, but in general I like the idea of focusing it through the component's focus function so we don't get unexpected surprises if we ever add additional logic there.

components/lms-adapter.js
```diff
@@ -82,6 +82,7 @@ export async function openLegacyDialog(opener, location, settings) {
 			const result = await dialogService.openLegacy(
 				location,
 				{
+					byPassOpenerFocus: settings.byPassOpenerFocus,
```
grant-cleary, line 85: Turns out this was already a supported setting in the legacy dialog service; just not in the LMS JS.

components/equation.js
grant-cleary, line 58: It does occur to me that I could use the provider stuff to pass around the `focus()` method instead of invoking it on root. I don't know if one method is better than the other.

dbatiste: Yeah I'm not sure how much I like the provider stuff within the context of the editor.  I think what you're doing here is fine.

dbatiste (approved):
Looks good.  I'm not 100% sure where focus should be returned to in the case of Preview or WordCount.  I suppose being consistent in the behaviour is a good thing.

grant-cleary:
> Looks good. I'm not 100% sure where focus should be returned to in the case of Preview or WordCount. I suppose being consistent in the behaviour is a good thing.

Yeah, in the current editor, closing the preview dialog does return focus in this manner, so I opted to remain consistent with both that behaviour and the rest of the dialogs. Word count has no analog in the existing editor, but it feels to me (sort of at least) that word count should probably behave like preview. It's also slightly more complicated, since you can open that dialog with an active selection to get its specific word count, so in that case I think we probably want to return focus to the selection regardless.

PR #114, opened 2021-03-01
Refactor tinyMCE code to better support toolbar refactoring.

dbatiste:
For the most part this is just moving the tinyMCE code out to the mixin.  Some additional change will follow to clean this up further once I create the toolbars and remove the plugin dependencies on tinyMCE.

### DIFF
--- a/htmleditor-mixin.js
+++ b/htmleditor-mixin.js
@@ -1,0 +1,487 @@
+import '@brightspace-ui/core/components/alert/alert.js';
+import 'tinymce/tinymce.js';
+import 'tinymce/icons/default/icons.js';
+import 'tinymce/plugins/autosave/plugin.js';
+import 'tinymce/plugins/autolink/plugin.js';
+import 'tinymce/plugins/charmap/plugin.js';
+import 'tinymce/plugins/code/plugin.js';
+import 'tinymce/plugins/directionality/plugin.js';
+import 'tinymce/plugins/emoticons/plugin.js';
+import 'tinymce/plugins/emoticons/js/emojis.js';
+import 'tinymce/plugins/fullpage/plugin.js';
+import 'tinymce/plugins/fullscreen/plugin.js';
+import 'tinymce/plugins/hr/plugin.js';
+import 'tinymce/plugins/image/plugin.js';
+import 'tinymce/plugins/imagetools/plugin.js';
+import 'tinymce/plugins/link/plugin.js';
+import 'tinymce/plugins/lists/plugin.js';
+import 'tinymce/plugins/preview/plugin.js';
+import 'tinymce/plugins/quickbars/plugin.js';
+import 'tinymce/plugins/table/plugin.js';
+import 'tinymce/plugins/textpattern/plugin.js';
+import 'tinymce/themes/silver/theme.js';
+import { css, html, unsafeCSS } from 'lit-element/lit-element.js';
+import { addIcons } from './generated/icons.js';
+import { classMap } from 'lit-html/directives/class-map.js';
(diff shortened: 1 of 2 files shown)


htmleditor-mixin.js
```diff
@@ -0,0 +1,478 @@
...
+			quickbars_insert_toolbar: false,
+			relative_urls: false,
+			resize: true,
+			setup: editor => {
+				addIcons(editor);
+
+				editor.on('blur', () => {
+					if (this.pasteLocalImages) editor.uploadImages();
+
+					this._html = editor.getContent();
```
dbatiste, line 382: Previously this line was only being run for inline.  I _think_ it should be ok to run this regardless of the mode.  In this mixin, we don't really want to have any notion of the type of editor (full, inline, inline-limited) so the change for isEditing is done in the consumer of this mixin.

htmleditor-mixin.js
```diff
@@ -0,0 +1,478 @@
...
+	if (cultureIndex !== -1) tinymceLang = tinymceLang.substring(0, cultureIndex);
+	if (!tinymceLangs.includes(tinymceLang)) {
+		tinymceLang = tinymceLangs.find(lang => {
+			return lang.startsWith(tinymceLang);
+		});
+		if (!tinymceLang) tinymceLang = 'en';
+	}
+}
+
+const a11yCheckerRuleUrlMap = {
```
dbatiste, line 56: These were moved out of the component/mixin since they didn't need to be defined over and over.

htmleditor-mixin.js
grant-cleary: Agreed, this should be safe to do in practice. When the LMS submits a form that includes an editor, we more or less do this anyways. I see no reason why we wouldn't want this property to be kept in sync.

htmleditor-mixin.js
```diff
@@ -0,0 +1,478 @@
...
+		this.provideInstance('attachedImagesOnly', this.attachedImagesOnly);
+		this.provideInstance('fileUploadForAllUsers', this.fileUploadForAllUsers);
+		this.provideInstance('fullPage', this.fullPage);
+		this.provideInstance('noFilter', this.noFilter);
+		this.provideInstance('wordCountInFooter', this.wordCountInFooter);
+		this.provideInstance('localize', this.localize.bind(this));
+
+		requestAnimationFrame(() => {
+			// eventually we can lazy load, but not until the plugins have been updated for toolbar refactoring
+			this._initializeTinymce();
```
dbatiste, line 242: So yeah, eventually the idea is that this `_initializeTinyMCE` can be called on demand, but currently, the existing toolbar plugins rely on tinyMCE so it's not yet possible to lazy load/init tinyMCE.

htmleditor-mixin.js
```diff
@@ -0,0 +1,478 @@
...
+					this._html = editor.getContent();
+
+					this.dispatchEvent(new CustomEvent(
+						'd2l-htmleditor-blur', {
+							bubbles: true
+						}
+					));
+				});
+
+				const createSplitButton = (name, icon, tooltip, cmd, items) => {
```
dbatiste, line 391: This custom UI stuff will eventually get removed from here once we have our own toolbar components.

htmleditor-mixin.js
grant-cleary: Ah yeah, that makes sense. I like this. Once the toolbars are our own, we shouldn't run into any further problems lazy-loading. I would love to see a case where we only actually loaded the plugins we needed for a given editor configuration.

grant-cleary (approved):

PR #115, opened 2021-03-01
When in an FRA context, specify the document base URL as the Valence host and ensure we're using that to construct images

grant-cleary:
Inserting an image in an FRA context currently fails because the current workflow assumes the image src is going to match the FRA's host domain (the CDN). We should set `document_base_url` in TinyMCE's init to match the Valence host if we're in an FRA context to get around this problem.

### DIFF
--- a/components/image.js
+++ b/components/image.js
@@ -141,14 +141,10 @@ tinymce.PluginManager.add('d2l-image', function(editor) {
 					)
 				);
 
-				const tempImg = document.createElement('img');
-				tempImg.setAttribute('src', src);
-				tempImg.setAttribute('alt', e.detail.IsDecorative ? '' : e.detail.ImageAlt);
-				tempImg.setAttribute('title', e.detail.IsDecorative ? '' : e.detail.ImageAlt);
-				tempImg.setAttribute('data-d2l-editor-default-img-style', 'true');
-				tempImg.style.maxWidth = '100%';
+				const altText = e.detail.IsDecorative ? '' : e.detail.ImageAlt;
+				const imgHtml = `<img src="${src}" alt="${altText}" title="${altText}" data-d2l-editor-default-img-style>`;
 
-				editor.execCommand('mceInsertContent', false, tempImg.outerHTML);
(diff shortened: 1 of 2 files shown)


components/image.js
```diff
@@ -141,14 +141,10 @@ tinymce.PluginManager.add('d2l-image', function(editor) {
...
-				const tempImg = document.createElement('img');
-				tempImg.setAttribute('src', src);
-				tempImg.setAttribute('alt', e.detail.IsDecorative ? '' : e.detail.ImageAlt);
-				tempImg.setAttribute('title', e.detail.IsDecorative ? '' : e.detail.ImageAlt);
-				tempImg.setAttribute('data-d2l-editor-default-img-style', 'true');
-				tempImg.style.maxWidth = '100%';
-
-				editor.execCommand('mceInsertContent', false, tempImg.outerHTML);
+				const altText = e.detail.IsDecorative ? '' : e.detail.ImageAlt;
+				const imgHtml = `<img src="${src}" alt="${altText}" title="${altText}" data-d2l-editor-default-img-style style="max-width: 100%;">`;
```
grant-cleary, line 145: When constructing the `tempImg` element above, the browser tries to render it, which fails because it interprets the `src` as relative to the host iframe (the CDN) - This doesn't impact the image rendering in the editor, since at that point it does resolve properly relative to the `document_base_url` setting. Nonetheless, we probably want to avoid a JS error every time someone inserts an image.

htmleditor-mixin.js
```diff
@@ -306,6 +306,11 @@ export const HtmlEditorMixin = superclass => class extends Localizer(RtlMixin(Pr
 			autosave_retention: '0s'
 		};
 
+		const documentConfig = {};
+		if (this._fraContext && this._context && this._context.host) {
```
dbatiste, line 310: I think this `this._fraContext` check here shouldn't be necessary?

dbatiste (approved):

htmleditor-mixin.js
grant-cleary: I think you're right. I was thinking about some eventual use case where this_context.host could be specified while this._fraContext was false... But in that case I think we still want to set the base URL.

PR #117, opened 2021-03-03
Container divs should inherit the height of their parents when the HTML editor has a percentage height

grant-cleary:
When the editor is given a height in percentage, it (by default) will size down to near zero (50-something pixels in my testing), since its container divs don't have a fixed height. If this is the case, we want to set the container divs to inherit their parents' height to ensure the editor has something to size itself against.

### DIFF
--- a/htmleditor-mixin.js
+++ b/htmleditor-mixin.js
@@ -479,7 +479,7 @@ export const HtmlEditorMixin = superclass => class extends Localizer(RtlMixin(Pr
 			'd2l-htmleditor-tinymce': true,
 			'd2l-htmleditor-tinymce-percentage-height': this.height.includes('%')
 		};
-		
+
 		const textAreaClasses = {
 			'd2l-htmleditor-no-tinymce': !isShadowDOMSupported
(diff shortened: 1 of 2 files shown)


htmleditor-mixin.js
```diff
@@ -128,7 +128,9 @@ export const HtmlEditorMixin = superclass => class extends Localizer(RtlMixin(Pr
 			.d2l-htmleditor-no-tinymce {
 				display: none;
 			}
-
+			.d2l-htmleditor-tinymce-percentage-height {
+				height: inherit;
```
dbatiste, line 132: Is there any harm in always setting this?

dbatiste (approved):

htmleditor-mixin.js
grant-cleary: My only thought is that it might be unexpected if the editor's height is set in `px` and its containers start trying to size themselves based on the nearest parent div. I'm not sure if that's a problem in practice or if it actually matters though. :shrug:

PR #121, opened 2021-03-11
Add attributes and custom color picker to the new editor; rejig toolbars to reflect latest design

grant-cleary:
Okay, so this is a bunch of changes required to get the editor going with the current toolbars. There are three major components here:
1. Switch to using D2L's LMS color picker dialog if context is available.
2. Add the "Insert Attributes" functionality (copy of LMS functionality, but ultimately net-new).
3. Localize any toolbar lang terms that haven't yet been localized.


components/emoji.js
```diff
@@ -0,0 +1,616 @@
...
+ * Licensed under the LGPL or a commercial license.
+ * For LGPL see License.txt in the project root for license information.
+ * For commercial licenses see https://www.tiny.cloud/
+ *
+ * Version: 5.7.0 (2021-02-10)
+ */
+/**
+ * Modified by: Grant Cleary
+ * Modification date: March 11, 2021
+ */
```
grant-cleary, line 14: Appears to be required as per the LGPL license. Hopefully we won't need this work-around for long and we can just remove this whole thing.

components/emoji.js
```diff
@@ -0,0 +1,616 @@
...
+
+    var register = function (editor, database) {
+      var onAction = function () {
+        return open(editor, database);
+      };
+      /* Modified code begins here */
+      editor.addCommand('mceEmoticons', function () {
+        onAction();
+      });
+      /* Modified code ends here */
```
grant-cleary, line 590: Above is the modified code.

htmleditor-mixin.js
```diff
@@ -347,6 +346,7 @@ export const HtmlEditorMixin = superclass => class extends Localizer(RtlMixin(Pr
 			font_formats: `Arabic Transparent=arabic transparent,sans-serif; Arial=arial,helvetica,sans-serif; Comic Sans=comic sans ms,sans-serif; Courier=courier new,courier,sans-serif; Ezra SIL=ezra sil,arial unicode ms,arial,sans-serif; Georgia=georgia,serif; Lato (${this.localize('font.family.recommended')})=Lato,sans-serif; SBL Hebrew=sbl hebrew,times new roman,serif; Simplified Arabic=simplified arabic,sans-serif; Tahoma=tahoma,sans-serif; Times New Roman=times new roman,times,serif; Traditional Arabic=traditional arabic,serif; Trebuchet=trebuchet ms,helvetica,sans-serif; Verdana=verdana,sans-serif; 돋움 (Dotum)=dotum,arial,helvetica,sans-serif; 宋体 (Sim Sun)=simsun; 細明體 (Ming Liu)=mingliu,arial,helvetica,sans-serif`,
 			fontsize_formats: '8pt 10pt 12pt 14pt 18pt 24pt 36pt',
 			height: this.height,
+			image_uploadtab: false,
```
grant-cleary, line 349: It turns out that the base Image plugin allows you to upload images through their workflow... We don't want this, so this config turns that off.

lang/en.js
```diff
@@ -1,16 +1,44 @@
...
+	"indent": "Increase Indent",
 	"inline.button": "Edit {name}",
+	"insert": "Insert",
 	"insertstuff.tooltip": "Insert Stuff",
+	"list.bullets": "Bulleted List",
+	"list.numbers": "Numbered List",
+	"outdent": "Decrease Indent",
 	"preview.tooltip": "Preview",
 	"quicklink.tooltip": "Insert Quicklink",
+	"symbols": "Symbols",
```
grant-cleary, line 41: I have yet to check whether some of these lang terms exist in the LMS, but figured it would be worth putting this up for review while I check that and fill in any that I can find.

components/colors.js
```diff
@@ -0,0 +1,150 @@
...
+		timerId = setTimeout(() => {
+			let contextNode = (editor.selection ? editor.selection.getNode() : null);
+			if (contextNode) {
+				if (contextNode.nodeType === Node.DOCUMENT_NODE) contextNode = contextNode.body;
+				const currentForegroundColor = getComputedStyle(contextNode, 'color');
+
+				const colorIcon = root.querySelector('#tox-icon-text-color__color');
+				colorIcon.style.fill = currentForegroundColor;
+			}
+		}, 100);
```
grant-cleary, line 91: Debounced update of the icon color. Downside of this is that it does override the hover/focus styles on the paintbrush bit of the SVG whenever this event fires (which is often). I'm unsure if that's what we want, but if we don't want it, we could do something like check if the fill is currently the hover/focus color and not change it if it is.

components/colors.js
```diff
@@ -0,0 +1,150 @@
...
+		icon: 'textcolor',
+		tooltip: localize('colors.tooltip'),
+		onAction: () => {
+			let dialog = root.querySelector('d2l-htmleditor-color-picker-dialog');
+			if (!dialog) dialog = root.appendChild(document.createElement('d2l-htmleditor-color-picker-dialog'));
+
+			dialog.opener = root.host;
+			dialog.opened = true;
+
+			let contextNode = (editor.selection ? editor.selection.getNode() : null);
```
dbatiste, line 63: I'm ot sure if you notice anything rerender... maybe it is ok because Lit is batching the updates and the dialog will not have opened yet (🤞 ).  I'd be tempted to move this background property assignment before opening the dialog.

components/colors.js
```diff
@@ -0,0 +1,150 @@
...
+	editor.on('selectionchange', () => {
+		clearTimeout(timerId);
+
+		timerId = setTimeout(() => {
+			let contextNode = (editor.selection ? editor.selection.getNode() : null);
+			if (contextNode) {
+				if (contextNode.nodeType === Node.DOCUMENT_NODE) contextNode = contextNode.body;
+				const currentForegroundColor = getComputedStyle(contextNode, 'color');
+
+				const colorIcon = root.querySelector('#tox-icon-text-color__color');
```
dbatiste, line 88: I was sort of expecting to see this swatch update above when the dialog closes and a new color has been picked.  When does `selectionchange` get dispatched?

components/emoji.js
dbatiste: Haha... so that's one wa to deal with it.  ;)

htmleditor-mixin.js
dbatiste: We'll want to make sure this change doesn't get forgotten in the new toolbar branch.

components/emoji.js
dbatiste: Granted, we don't use their LGPL license... their LGPL license was the primary reason we ended up getting an enterprise license (LGPL is unacceptable for us).

components/colors.js
grant-cleary: Oh, good point. I think that makes sense. There's no reason this can't compute prior.

components/emoji.js
grant-cleary: Yeah, I don't actually have a fantastic understanding of how modifying this code works given our enterprise license. Maybe this is unnecessary. I'll ask Jeannette.

components/emoji.js
grant-cleary: Yeah, I imagine this will be temporary, but navigating licensing is rough. 😅

grant-cleary:
There's definitely something breaking in the visual diff stuff. I didn't see JS errors when I was testing this locally, but it wouldn't be the first time the visual diff tests found an error I didn't see.

grant-cleary:
Urgh, that last merge from this branch messed up the goldens.

grant-cleary:
Blech, never mind, I'm stupid. Those were the changes made when the last visual diff PR was merged in here. 🤦

dbatiste (approved):

PR #128, opened 2021-03-15
Fix inverted logic in word count dialog calculations

grant-cleary:
Turns out the logic for calculating the word count in the dialog was a bit backwards, leading to trailing spaces being interpreted as an extra word. I've switched over the logic to be the same as the specific word count function to avoid any future confusion. I also added a new test case to capture this specific scenario. I can confirm that it fails before this change and passes afterwards.


grant-cleary:
Closing for now. Will rebase onto master and re-open once the custom toolbars work is merged.

PR #129, opened 2021-03-15
Add attributes, forked emoji plugin, image upload disable config, and save function to custom toolbars

grant-cleary:
This should pull into this branch the rest of the work from master merged in the last week or so that we actually want to keep (that wasn't pulled in during the last merge of master). Some modifications were necessary to work with the newer custom toolbars work.


components/toolbar/toolbar-full.js
```diff
@@ -4,6 +4,8 @@ import './button-color.js';
 import './button-menu.js';
 import './button-toggle.js';
 import '../plugins/a11ychecker.js';
+import '../plugins/attributes.js';
+import '../plugins/emoji.js';
```
grant-cleary, line 8: We don't use anything in this import, but we still need to import the plugin to ensure it runs.

htmleditor-mixin.js
```diff
@@ -7,7 +7,6 @@ import 'tinymce/plugins/autolink/plugin.js';
 import 'tinymce/plugins/charmap/plugin.js';
 import 'tinymce/plugins/code/plugin.js';
 import 'tinymce/plugins/directionality/plugin.js';
-import 'tinymce/plugins/emoticons/plugin.js';
 import 'tinymce/plugins/emoticons/js/emojis.js';
```
dbatiste, line 10: It looks like we should be moving this emojis file over as well eh?

dbatiste (approved):
Just that one question about the other empoji import.

htmleditor-mixin.js
grant-cleary: We could. Do we need to though? I guess I'm being a little optimistic that we're eventually going to be able to switch back to using their plugin lol.

PR #138, opened 2021-03-30
Use custom toolbars for HtmlEditor.

dbatiste:


components/toolbar/table-create.js
```diff
@@ -0,0 +1,225 @@
...
+	constructor() {
+		super();
+		this._columnCount = 1;
+		this._rowCount = 1;
+		this._selectorId = getUniqueId();
+	}
+
+	connectedCallback() {
+		super.connectedCallback();
+		// can't use Localizer mixin due to crazy issue with ResizeObserver in HierarchicalView & Localize
```
grant-cleary, line 86: Lol, now I have questions.

htmleditor-mixin.js
```diff
@@ -348,8 +340,6 @@ export const HtmlEditorMixin = superclass => class extends Localizer(RtlMixin(Pr
 				'mentions': `${baseImportPath}/tinymce/plugins/mentions/plugin.js`,
 				'powerpaste': `${baseImportPath}/tinymce/plugins/powerpaste/plugin.js`
 			},
-			font_formats: `Arabic Transparent=arabic transparent,sans-serif; Arial=arial,helvetica,sans-serif; Comic Sans=comic sans ms,sans-serif; Courier=courier new,courier,sans-serif; Ezra SIL=ezra sil,arial unicode ms,arial,sans-serif; Georgia=georgia,serif; Lato (${this.localize('font.family.recommended')})=Lato,sans-serif; SBL Hebrew=sbl hebrew,times new roman,serif; Simplified Arabic=simplified arabic,sans-serif; Tahoma=tahoma,sans-serif; Times New Roman=times new roman,times,serif; Traditional Arabic=traditional arabic,serif; Trebuchet=trebuchet ms,helvetica,sans-serif; Verdana=verdana,sans-serif; 돋움 (Dotum)=dotum,arial,helvetica,sans-serif; 宋体 (Sim Sun)=simsun; 細明體 (Ming Liu)=mingliu,arial,helvetica,sans-serif`,
-			fontsize_formats: '8pt 10pt 12pt 14pt 18pt 24pt 36pt',
```
grant-cleary, line 352: Was removing these intentional?

grant-cleary: Oh, probably right, since I imagine we've wrested control over TinyMCE for this functionality?

htmleditor-mixin.js
```diff
@@ -384,105 +374,31 @@ export const HtmlEditorMixin = superclass => class extends Localizer(RtlMixin(Pr
...
-				{ title: 'Code', format: 'code' }
-			],
 			target: textarea,
-			toolbar: this._getToolbarConfig(),
-			//toolbar: false,
-			toolbar_mode: 'sliding',
+			toolbar: false,
 			valid_elements: '*[*]',
-			width: this.width,
+			width: '100%',
```
grant-cleary, line 401: Is this intentional? Are we handling this elsewhere?

htmleditor.js
```diff
@@ -138,26 +197,40 @@ class HtmlEditor extends HtmlEditorMixin(SkeletonMixin(LitElement)) {
...
 		const containerClasses = {
 			'd2l-htmleditor-container': true,
 			'd2l-skeletize': true,
-			'd2l-htmleditor-percentage-height': super.height.includes('%')
+			'd2l-htmleditor-percentage-height': super.height.includes('%'),
+			'd2l-htmleditor-fullscreen': this._fullscreen
 		};
 
+		const containerStyles = {};
+		containerStyles.width = this.width;
```
grant-cleary, line 213: Oh I see now. We're not setting TinyMCE's width because we're applying it to the container and setting the editor's internal width to 100% to size to the container. Got it.

components/toolbar/table-create.js
dbatiste: Yeah, I don't fully understand why our Localize mixin has problems with components that also wire-up ResizeObserver.  I would have used it but I didn't want to invest too much time in it given the simple alternative sol'n.

htmleditor-mixin.js
dbatiste: Yeah, I don't think these serve any purpose in the tinyMCE config anymore.  We can certainly add them back if needed, but I think they were just being used to populate tinyMCE's toolbar menus.

htmleditor.js
dbatiste: You got it.  We need to apply the width to the container (parent) so that it limits the width of the toolbar as well as the editor's container.

htmleditor-mixin.js
grant-cleary: Yup, that's right. That was their only purpose.

htmleditor.js
```diff
@@ -220,9 +221,12 @@ class HtmlEditor extends HtmlEditorMixin(SkeletonMixin(LitElement)) {
 		const containerStyles = {};
 		containerStyles.width = this.width;
 
+		const toolbarStyles = {};
+		if (this.toolbarHorizontalOffset) toolbarStyles.right = this.toolbarHorizontalOffset;
+
 		return html`
 			<div class="${classMap(containerClasses)}" style="${styleMap(containerStyles)}">
-				<div class="d2l-htmleditor-toolbar-container">${toolbar}</div>
+				<div class="d2l-htmleditor-toolbar-container" style="${styleMap(toolbarStyles)}">${toolbar}</div>
```
grant-cleary, line 228: Oh neat, I didn't know `styleMap` was also a thing.

htmleditor-mixin.js
```diff
@@ -112,6 +112,9 @@ export const HtmlEditorMixin = superclass => class extends Localizer(RtlMixin(Pr
 			.tox-tinymce {
 				border: none !important;
 			}
+			.tox-tinymce-aux {
+				z-index: 990; /* needs to be less than our dropdowns */
```
grant-cleary, line 116: I love z-index. 😅

grant-cleary (approved):

htmleditor.js
dbatiste: Yep... usually we try to avoid inline styles for the most part, because they are impossible to override without replacing them or using `!important`.  With CSS it is much easier to override by specifying a selector with high specificity or simply by changing the order of the selectors.

PR #143, opened 2021-03-31
US126098 - Add support for fr-on lang pack to HTML editor

grant-cleary:
Does what it says on the tin. :)


lang/fr-on.js
```diff
@@ -0,0 +1,87 @@
+/* eslint quotes: 0 */
```
grant-cleary, line 1: Copied from `fr.js` as per instructions.

grant-cleary:
... Alright, I think at this point there just must be something wrong with the current goldens. This literally doesn't touch anything visual.

dbatiste:
Do we need to do anything with [htmleditor.serge.json](https://github.com/BrightspaceUI/htmleditor/blob/master/htmleditor.serge.json)?

dbatiste:
> ... Alright, I think at this point there just must be something wrong with the current goldens. This literally doesn't touch anything visual.

Yah seems like there's a timing issue.

grant-cleary:
> Do we need to do anything with [htmleditor.serge.json](https://github.com/BrightspaceUI/htmleditor/blob/master/htmleditor.serge.json)?

Sending you a Slack link offline, but no, I don't believe so.

grant-cleary:
As discussed offline, I'm disabling the visual diff tests for now while we look into these timing issues.

dbatiste (approved):

PR #147, opened 2021-04-02
Add focus trap and shortcut for fullscreen.

dbatiste:
* adds focus trap while in fullscreen
* adds shortcut for toggling fullscreen (Cmd+Shift+F / Ctrl+Shift+F)
* limit the width in fullscreen to 1230px as per design


components/plugins/attributes.js
```diff
@@ -19,7 +19,8 @@ const name = 'd2l-attributes';
 export function addAttributesPlugin() {
 	tinymce.PluginManager.add(name, function(editor) {
 
-		const root = editor.getElement().getRootNode();
+		const context = requestInstance(editor.getElement(), 'context');
```
dbatiste, line 22: I originally made this change to support the idea that I would need to manage the placement of the dialogs within the focus trap, but so far it has not been necessary.  Nevertheless, I would like to tidy up method for getting the component, as well as the `requestInstance` stuff so I decided to keep these changes.

components/plugins/fullscreen.js
```diff
@@ -1,28 +1,43 @@
...
+		const context = requestInstance(editor.getElement(), 'context');
+		const component = context.component;
 
-		editor.addQueryStateHandler(cmds.fullscreen, () => {
-			return state;
-		});
+		let state = false;
 
-		editor.addCommand(cmds.fullscreen, () => {
+		const toggleFullscreen = () => {
```
dbatiste, line 14: This is the same logic as before - it's just been factored out so that it can be invoked both when the the user clicks the fullscreen toggle button, or when they use the shortcut key.

htmleditor-mixin.js
```diff
@@ -228,6 +228,10 @@ export const HtmlEditorMixin = superclass => class extends Localizer(RtlMixin(Pr
 		this.provideInstance('wordCountInFooter', this.wordCountInFooter);
 		this.provideInstance('localize', this.localize.bind(this));
 
+		this.provideInstance('context', { ...{
```
dbatiste, line 231: As mentioned in my previous comment I would like to tidy this stuff up to make it less chatty. So my plan it to move these other items into this context and spread the private context into this as well.  I figured I would leave that for another PR as to not confuse this PR.

components/plugins/fullscreen.js
```diff
@@ -1,28 +1,43 @@
...
+			if (e.keyCode !== 70) return;
+			if (!(e.metaKey && e.shiftKey) && !(e.ctrlKey && e.shiftKey)) return;
+			toggleFullscreen();
+		};
+
+		editor.addQueryStateHandler(cmds.fullscreen, () => state);
+
+		editor.addCommand(cmds.fullscreen, toggleFullscreen);
+
+		// inside iframe (note: metaKey combination only works on keydown)
```
grant-cleary, line 36: Interesting. I didn't know that about metaKey.

htmleditor-mixin.js
grant-cleary: Makes sense to me.

grant-cleary (approved):

components/plugins/fullscreen.js
dbatiste: Yeah, on Mac it is the Cmd (or otherwise know as the 🍎 key), and I think on Windows is can be the Windows key.  Though I think there can be some inconsistency with it.  This logic however, doesn't do any OS detection, so on Mac, either combination will work.  Hopefully that does not cause any conflicts.  The Shift key effectively mitigates the conflict with the standard Cmd/Ctrl+F for "Find".

grant-cleary: Cool yeah, makes sense to me.

PR #149, opened 2021-04-07
minor fixes

dbatiste:
* remove unnecessary import for standard menu-item-separator
* fix for background when forcing focusable focus
* make grant a code owner


CODEOWNERS
```diff
@@ -1 +1 @@
-*       @dbatiste
+*       @dbatiste @grant-cleary
```
grant-cleary, line 1: I thought I was on here already. O.o

grant-cleary (approved):

CODEOWNERS
grant-cleary: > I thought I was on here already. O.o

Oh crap, now I wonder if the PR where I aimed to add myself in here was one of the ones that I closed out without merging...

PR #150, opened 2021-04-07
Make sure font size text reflects the text shown in the editor

grant-cleary:
The font size text in the toolbar wasn't updating appropriately when changing selection or moving the carat within the editor. This PR makes sure we're accurately reflecting the size of the current text in the editor.


components/toolbar/button-menu.js
```diff
@@ -112,7 +112,8 @@ class ButtonMenu extends DropdownOpenerMixin(ToolbarItemMixin(FocusVisiblePolyfi
 				item.selected = (item.value === this._value);
 				if (item.selected) {
 					hasSelected = true;
-					this._valueText = item.textContent;
+					if (item.textContent) this._valueText = item.textContent;
+					else this._valueText = item.text;
```
grant-cleary, line 116: For the font size menu items, we're setting `text={fontSize}` on the menu-item component, so `item.textContent` isn't populated. In contrast, the font selection creates a span with the font being used as its text content, so in that case this does get set correctly.

grant-cleary: Thinking on it, I think it might be better to set `text={font}` on the font selection menu items as well. That way we only ever have to check `item.text`. I'm a bit worried about other menu items that might rely on the previous behaviour though. I'll poke around.

components/toolbar/toolbar-full.js
```diff
@@ -68,7 +68,7 @@ export function getPlugins(component) {
 		component.pasteLocalImages ? 'imagetools' : '',
 		'link',
 		'lists',
-		component.mentions && D2L.LP ? 'mentions' : '',
+		component.mentions && component._context ? 'mentions' : '',
```
grant-cleary, line 71: @mentions has an ifrau pass-through, so there's no reason to just limit it to `D2L.LP`.

components/toolbar/toolbar-full.js
```diff
@@ -303,7 +303,7 @@ class Toolbar extends Localizer(ToolbarMixin(LitElement)) {
 								<d2l-htmleditor-menu-item-separator></d2l-htmleditor-menu-item-separator>
 								<d2l-htmleditor-menu-item cmd="${cmds.tableProps}" text="${this.localize('table.properties')}"></d2l-htmleditor-menu-item>
 								<d2l-htmleditor-menu-item cmd="${cmds.tableDelete}" icon="table-delete-table" text="${this.localize('table.delete')}"></d2l-htmleditor-menu-item>
-							<d2l-menu>
+							</d2l-menu>
```
grant-cleary, line 306: Everything here and below is just formatting stuff. As far as I can tell none of this was actually broken.

dbatiste (approved):

PR #151, opened 2021-04-07
DE43115 - URL encode selected text before allowing it to be embedded in a quicklink

grant-cleary:
The selected text for inserting a quicklink can be HTML (e.g. an image), in which case Chrome expects it to be properly URL encoded when making the request to launch the dialog.


components/plugins/quicklink.js
```diff
@@ -129,18 +129,17 @@ class QuicklinkDialog extends RequesterMixin(LitElement) {
...
-						&selectedText=${this.text}
-						&parentModuleId=0
-						&canChangeType=true
-						&showCancelButton=true
-						&urlShowTarget=true
-						&urlShowCancelButtonInline=false
-						&contextId=
-					`);
+					let selectUrl = new D2L.LP.Web.Http.UrlLocation(`/d2l/lp/quicklinks/manage/${this._orgUnitId}/createdialog`)
+						.WithQueryString('typeKey', '')
```
grant-cleary, line 133: `WithQueryString` URL encodes and serializes its contents for us.

components/plugins/quicklink.js
```diff
@@ -129,18 +129,17 @@ class QuicklinkDialog extends RequesterMixin(LitElement) {
...
+						.WithQueryString('initialViewType', 'Default')
+						.WithQueryString('outputFormat', 'html')
+						.WithQueryString('selectedText', this.text)
```
grant-cleary, line 136: This is the actual problematic piece. `this.text` in this case can be HTML and Chrome does not like unencoded HTML showing up in query parameters (it refuses to make the request).

dbatiste (approved):

PR #152, opened 2021-04-08
DE43118 - Encode orgUnitId replace strings from ISF on content insert and decode them on post-process

grant-cleary:
The old editor did this by hooking scripts from a custom plugin into the editor's `SetContent` and `PostProcess` events. Because we can embed this in ISF, we can do the encoding directly in the ISF return, though decoding still needs to be done via the `PostProcess` event.


components/plugins/isf.js
```diff
@@ -376,12 +404,37 @@ export function addISFPlugin() {
 			}
 		};
 
+		const decodeReplaceStrings = (html) => {
+			const decode = (match) => {
+				const tempNode = new DOMParser().parseFromString(match, 'text/html');
+
+				const img = tempNode.firstChild;
```
grant-cleary, line 411: Whoops, I missed changing this in the way the `encode` method was changed. This will never return anything useful.

grant-cleary: I'll also push a change shortly to return early if the document body doesn't exist or has no children.

dbatiste (approved):

components/plugins/isf.js
```diff
@@ -35,6 +35,35 @@ export function addISFPlugin() {
 		const component = context.component;
 		const wmodeOpaque = context.wmodeOpaque;
 
+		const encodeReplaceStrings = (html) => {
```
grant-cleary, line 38: Damn it. I accidentally wiped out these changes.

components/toolbar/toolbar-inline-limited.js
```diff
@@ -32,7 +32,7 @@ export function getPlugins(component) {
 		'emoticons',
 		'link',
 		'lists',
-		component.mentions && D2L.LP ? 'mentions' : '',
+		component.mentions && component._context ? 'mentions' : '',
```
grant-cleary, line 35: Missed this when I switched it over in the full toolbar.

dbatiste (approved):

PR #156, opened 2021-04-09
more visual-diff tests

dbatiste:
This PR adds visual-diff tests for...

* height and width of for editor
* inline and inline-limited toolbars
* breakpoints for aggregating tables, biu, fonts, and other tools


htmleditor.js
```diff
@@ -42,9 +42,9 @@ class HtmlEditor extends HtmlEditorMixin(SkeletonMixin(LitElement)) {
 			noBorder: { type: Boolean, attribute: 'no-border' },
 			toolbarHorizontalOffset: { type: String, attribute: 'toolbar-horizontal-offset' },
 			type: { type: String },
-			_isEditing: { type: Boolean, attribute: 'is-editing' },
```
dbatiste, line 45: Just renaming these private properties to follow our general convention for naming properties on components.

grant-cleary (approved):

grant-cleary:
Weird... I wonder what's generating that error.

test/htmleditor.visual-diff.js
```diff
@@ -53,4 +74,38 @@ describe('d2l-htmleditor', () => {
 
 	});
 
+	const toggleEditing = async selector => {
+		await page.$eval(selector, editor => {
+			editor.shadowRoot.querySelector('button').click();
+			editor.focus();
+			return new Promise(resolve => requestAnimationFrame(resolve));
+		});
+		return hideCaret(selector);
```
dbatiste, line 83: So far the only way I have found to stop the blinking caret (which is the source of flakiness) is to make the caret transparent.  The timing of this is fragile, since we need to wait until tinyMCE has create the iframe and the document within. An alternative is to bake something into the editor component to do this when some special test variable is set, but I'd rather not pollute the component with test stuff.

dbatiste: Hopefully this doesn't become an issue for people who are consuming the component and have visual-diff tests.  I think it would only be an issue if they force focus into the editor, which they are unlikely to do.

PR #160, opened 2021-04-15
Add custom footer and reintroduce support for percentage heights

grant-cleary:
We need to rework the way we manage percentage heights in the editor now that we render our own toolbar. To accomplish this, we more or less render all of the inner containers (including TinyMCE itself) with `height: 100%` and then attach the provided height property to the outermost container only. This means we do need to ditch TinyMCE's resizing logic, since TinyMCE won't be in control of its own height anymore... Which also means we need our own footer instead of using TinyMCE's.


htmleditor.js
```diff
@@ -239,59 +312,91 @@ class HtmlEditor extends HtmlEditorMixin(SkeletonMixin(LitElement)) {
...
+			'd2l-htmleditor-editor-container-shim': true,
+			'd2l-htmleditor-editor-container-shim-resizing': this._resizing
 		};
 
 		const containerStyles = {};
+		if (!this._fullscreen) containerStyles.height = this.height;
 		containerStyles.width = this.width;
 
+		const flexContainerStyles = {};
+		if (!this._fullscreen) flexContainerStyles.height = this.height;
```
grant-cleary, line 332: I think it might be possible to just set this to 100% height all the time, now that I think about it. As long as the height is set on the outer container, this should just work.

grant-cleary:
Bah, I think my es-lint-config is out of date locally. Will update and fix these linting errors.

components/plugins/wordcount.js
```diff
@@ -127,33 +78,23 @@ export function addWordCountPlugin() {
...
+				default:
+					return countWords(text);
+			}
+		};
 
-				const countOption = countButton.getAttribute('data-d2l-count-option');
-				countButton.textContent = component.localize(getButtonLangTerm(countOption, isSelection), { count: getButtonCount(countOption, text) });
-			}, 100);
-		});
+		editor.addQueryStateHandler(cmds.wordCount, () => countType);
```
dbatiste, line 95: Could we, and does it make sense to, return the count and the type together as ...
```javascript
editor.addQueryValueHandler(cmds.wordCount, {
    count: getCount(),
    type: countType
});
```
so that the WordCount UI component doesn't need to make two queries, overloading the `queryCommandState`.  To be honest, I'm not sure which way makes the most sense to surface the type (value vs state).

components/footer/word-count-button.js
```diff
@@ -0,0 +1,124 @@
...
+		});
+	}
+
+	render() {
+		const buttonClasses = {
+			'd2l-htmleditor-word-count-button-hidden': this._countType === footerCountType.NOTHING
+		};
+
+		return html`
+			<button class="${classMap(buttonClasses)}" aria-haspopup="true" @click=${this._handleClick}>
```
dbatiste, line 95: What's the reason for `aria-haspopup="true"`?  I don't think our other buttons that launch dialogs include this.  Makes me 🤔 though.

components/plugins/wordcount.js
grant-cleary: The reason I didn't is because the dialog only fetches the type (not the count), so I wanted to avoid having the dialog compute counts unnecessarily. So I guess it's a question of whether we think that tradeoff is worthwhile. The dialog opener is likely to be invoked much less often compared to what the word count component does, so that's something to consider.

components/footer/word-count-button.js
grant-cleary: Originally I tried to use `d2l-button-subtle` (before discovering that the styles weren't fantastic) and I think I misread its README about accessible properties. I think this is only relevant if the button opens a _menu_.

components/plugins/wordcount.js
```diff
@@ -84,37 +61,11 @@ export function addWordCountPlugin() {
 				};
 			}
 
-			if (wordCountInFooter) dialog.countType = countButton.getAttribute('data-d2l-count-option');
+			dialog.countType = countType;
 
 			dialog.addEventListener('d2l-htmleditor-wordcount-select-option', (e) => {
 				if (!e.detail) return;
```
dbatiste, line 67: It looks like the detail will always exist... is that true?  If so it looks like this could just become:

```javascript
dialog.addEventListener('d2l-htmleditor-wordcount-select-option', e => countType = e.detail.countType);
```

components/footer/word-count-button.js
grant-cleary: Though [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Element/ariaHasPopup) implies that it should still be used for dialogs... So I'm with you on the 🤔.

grant-cleary: Oh, but it looks like it should be set to `aria-haspopup="dialog"` in that case.

components/plugins/wordcount.js
grant-cleary: Yeah, I think you're right. I copied this from some other component that was relying on a less-certain event.

grant-cleary: Same will be true in the component, so I'll change that there as well.

grant-cleary: Ah, never mind. The component already doesn't check it. lol

htmleditor.js
```diff
@@ -196,6 +236,41 @@ class HtmlEditor extends HtmlEditorMixin(SkeletonMixin(LitElement)) {
...
+		e.preventDefault();
+
+		const container = this.shadowRoot.querySelector('.d2l-htmleditor-flex-container');
+		const containerRect = container.getBoundingClientRect();
+
+		this._resizing = true;
+		this._originalHeight = containerRect.height;
+		this._originalMousePos = e.pageY;
+
+		window.addEventListener('mousemove', this._handleResize, false);
```
dbatiste, line 270: Just to make sure... are we 100% sure we need the shim if we're listening on the `window`?

grant-cleary: I'll test again, but this code was written before I added the shim, so I'm reasonably sure, yes.

grant-cleary: Confirmed, yes. Removing the shim puts us into forever-resizing mode if the mouseup event happens on the internal iframe doc.

components/plugins/wordcount.js
dbatiste: Your reasoning makes sense.

components/footer/word-count-button.js
dbatiste: Interesting.  That sounds like a bigger fix.

grant-cleary: I can remove it or switch it to `dialog`. Maybe the former for consistency? Unsure.

dbatiste: Does it get announced differently? Ideally we are consistent, but I also think that doing this for the other toolbar buttons and menu items that open a dialog will be tricker, and then there is the issue of this being a more systemic omission.

grant-cleary: Hmm, from https://www.w3.org/TR/wai-aria-1.2/#aria-haspopup:
```
If there is no visual indication that an element will trigger a popup, authors are advised to consider whether use of aria-haspopup is necessary, and avoid using it when it's not.
```
... So maybe not such a big deal? I just don't entirely understand why we would want to avoid using it in this case. Weird?

test/components.visual-diff.html
```diff
@@ -98,5 +99,9 @@
 			</d2l-htmleditor-button-menu>
 		</div>
 
+		<div class="visual-diff">
```
dbatiste, line 102: I usually don't condone setting private properties in tests, however I think i would be possible to capture the different count types by setting:
```html
<d2l-htmleditor-word-count-button id ="..." _count="..." or _countType="..."></d2l-htmleditor-word-count-button>
```
Arguably we'd set `attribute: '_count-type'` for the property.

grant-cleary: Oh that's a neat idea.

grant-cleary: Yeah, I didn't bind the properties to attributes because they're not used in the actual editor component in that way. But it would be nice to capture that in the tests.

PR #168, opened 2021-04-16
Fix percentage height in the editor and build custom footer to allow resizing to work properly given those changes

grant-cleary:
See [here](https://github.com/BrightspaceUI/htmleditor/pull/162) for the original abortive PR where I tried to figure out the visual diff issues (there's some context here). If they fail again (which they probably will), I'll comment them out while I look into them further to make sure this can get merged.


htmleditor.js
```diff
@@ -256,42 +336,66 @@ class HtmlEditor extends HtmlEditorMixin(SkeletonMixin(LitElement)) {
...
-				];
+				return html`
+					<div class="d2l-htmleditor-label-flex-container" style="${styleMap(flexContainerStyles)}">
+						<span class="d2l-input-label d2l-skeletize" aria-hidden="true">${this.label}</span>
+						${this._renderEditor(this._toolbar.render(), true)}
+					</div>
+				`;
 			} else {
-				return this._renderEditor(this._toolbar.render());
+				return this._renderEditor(this._toolbar.render(), true);
```
dbatiste, line 385: If we ever do want to make a footer component, I imagine this would becomes something like:
```javascript
return this._renderEditor(this._toolbar.render(), this._toolbar.renderFooter());
```
Or something along those lines.

dbatiste (approved):

htmleditor.js
grant-cleary: Agreed. I think it will probably end up this way (or maybe should end up this way to keep the main component cleaner).

PR #176, opened 2021-04-22
Don't try to set the filename on the blob TinyMCE passes to us when copy/pasting images into the editor

grant-cleary:
The File API's name property is read-only, so attempting to set the filename produces an error that then blocks the workflow from continuing. This seems to happen only on copy/paste (not on drag and drop) and only in the new editor. Given that we get the blob object directly from TinyMCE, I have to assume something changed in PowerPaste between 4.7.2 (or whatever we're on in the old editor) and the version we're on now and there's an inconsistency in the way these workflows behave.


components/lms-adapter.js
```diff
@@ -160,22 +160,20 @@ export async function openLegacyDialog(opener, location, settings) {
...
+export async function uploadFile(orgUnitId, file, maxFileSize) {
 	if (window.ifrauclient) {
 
 		if (!fileUploadService) {
 			const ifrauClient = await window.ifrauclient().connect();
 			fileUploadService = await ifrauClient.getService('file-upload', '0.1');
 		}
 
-		return fileUploadService.uploadFile(fileName, file, maxFileSize);
+		return fileUploadService.uploadFile(file, maxFileSize);
```
grant-cleary, line 171: This is going to require an LMS PR, as well as a manual bump of the editor version in the Lessons FRA (yay breaking changes to ifrau services). I figure this PR needs to come first, THEN the Lessons FRA bump and the LMS change need to follow relatively closely. I'll be putting the latter PR up shortly and I'll be coordinating with @ianlaidlaw on getting Lessons updated when we get there. :)

components/lms-adapter.js
```diff
@@ -160,22 +160,20 @@ export async function openLegacyDialog(opener, location, settings) {
...
 
 	} else {
 
 		return new Promise((resolve, reject) => {
 
-			file.name = fileName;
```
grant-cleary, line 177: I'm going to do a bit more testing on this (specifically around uploading multiple files at once), but I broke my local instance a bit so I need to rebuild first. TinyMCE should apply unique names if multiple images are pasted in at once and we overwrite them all anyways as soon as we upload them to temp, so this name-setting business is sort of useless. I can't really remember why we did it in the first place, to be honest.

grant-cleary:
Test failure is that flaky inline-limited test. I'm going to aim to fix that once and for all in this PR as well just to get it out of the way. :)

grant-cleary:
Mm, sometimes TinyMCE sends us a `Blob` and not a `File`, in which case its name isn't defined. I'm going to take another swing at this. I guess we were getting around this before by "faking" adding a name with `file.name = 'blah'`... Adding an arbitrary property instead of using the file API I guess. 🤷

components/lms-adapter.js
```diff
@@ -174,10 +174,10 @@ export async function uploadFile(orgUnitId, fileName, file, maxFileSize) {
...
+			const fileToUpload = file.name ? file : new File([file], fileName, { type: file.type });
```
grant-cleary, line 177: This addresses the Blob vs. File problem. If the file doesn't have a name, it's a blob, so we convert it to a file and give it a name in the process. If the name is empty or null, that's sort of a problem too, so I don't think we explicitly need to check `undefined`.

grant-cleary: Some Googling suggests that the new `const` just creates a pointer to the existing blob/file data in memory, so this should negligibly increase the browser's memory usage.

test/htmleditor.visual-diff.js
```diff
@@ -101,7 +101,7 @@ describe('d2l-htmleditor', () => {
 			].forEach(info => {
 				it(info.name, async function() {
 					if (info.action) await info.action(info.selector);
-					await page.$eval(info.selector, () => new Promise(resolve => requestAnimationFrame(resolve)));
+					await page.$eval(info.selector, () => new Promise(resolve => setTimeout(resolve, 50)));
```
grant-cleary, line 104: Slightly longer than two ideal frames, so hopefully that's enough to wipe out any flake. Visual diff tests passed this time, which is promising, but... Flake is flake, so who knows. :)

components/lms-adapter.js
dbatiste: Alternatively there is the [instanceof](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/instanceof) operator if you need to check the type.

dbatiste (approved):

components/lms-adapter.js
grant-cleary: Hmm, I do like that better. That said, I think we always need to have this name set properly and I'm not super convinced we can guarantee it will be though, given we don't control how the file is first defined. Bummer.

PR #179, opened 2021-04-26
DE43187 - Allowing file uploads for all users should bypass view permissions

grant-cleary:
This aligns the new editor's behaviour with the previous LMS editor.


grant-cleary:
Crap... 50 ms timeout for that inline-limited test didn't fix the flake. :(

grant-cleary:
It's also possible the golden is the flaky image here.

PR #182, opened 2021-04-28
DE43362 - Use correct host object when rendering equation editor images

grant-cleary:
~Haven't tested this one just yet; need to rebuild first because of some broken stuff in my local LMS. Will give this a test run tomorrow. :)~

This caused an issue where `host` was undefined and the URL snippets we were providing for reaching out to WIRIS and then trying to render the image itself were set to `undefined/wiris/...` and `undefined/lp/...` respectively. Because these aren't absolute URLs, the browser prepends the iframe src to it and it ends up creating an infinite render loop as we keep trying (and failing) to load the image to render it.


grant-cleary:
Hmm. Apparently the timeout change did _not_ fix the flaky test. I thought 100ms would be more than enough. =/

dbatiste (approved):

PR #187, opened 2021-04-29
Update table selector labels to show 'M columns by N rows' explicitly

grant-cleary:
This fix is meant to address a11y specifically. Without the visual indication of seeing the table select grid when moving your mouse around the table selector, the label "N by M" doesn't indicate which number corresponds to rows and which corresponds to columns. This fix addresses that.


lang/ar.js
```diff
@@ -75,7 +75,7 @@ export default {
 	"table.column.paste-before": "لصق العمود قبل",
 	"table.delete": "حذف جدول",
 	"table.insert": "إدراج جدول",
-	"table.insert.colums-by-rows": "{columns} مقابل {rows}",
+	"table.insert.colums-by-rows": "{columns} من الأعمدة في {rows} من الصفوف",
```
grant-cleary, line 78: These are just the base translations used in the LMS currently. They don't take advantage of the plural support, but we need something we can backport to 05 for now.

lang/da.js
```diff
@@ -75,7 +75,7 @@ export default {
 	"table.column.paste-before": "Indsæt kolonne før",
 	"table.delete": "Slet tabel",
 	"table.insert": "Indsæt tabel",
-	"table.insert.colums-by-rows": "{columns} af {rows}",
+	"table.insert.colums-by-rows": "{columns} kolonner med {rækker} rækker",
```
grant-cleary, line 78: This was wrong in the LMS translation and I didn't pay enough attention when copying/pasting. I'll look over these again.

dbatiste (approved):

PR #190, opened 2021-05-03
Only enable source code editor if that setting is provided

grant-cleary:
We should only display the source code editor if it's specifically set to be enabled. We do this by mimicking the old editor's behaviour: respect the `source-code` attribute if it's provided, or fall back to checking a provided context value if not.


components/toolbar/toolbar-full.js
```diff
@@ -436,6 +441,10 @@ class Toolbar extends ToolbarMixin(LitElement) {
 
 customElements.define('d2l-htmleditor-toolbar-full', Toolbar);
 
+function allowSourceEditing(component) {
+	return component.sourceCode || (component._context && component._context.editSource);
```
grant-cleary, line 445: This is equivalent to the existing logic, except there's going to be an LMS PR on the way that actually sets these things appropriately. :)

grant-cleary: (The LMS PR will need to merge before this one does to avoid just shutting off the source code editor entirely for a few builds.)

htmleditor-mixin.js
```diff
@@ -94,6 +94,7 @@ export const HtmlEditorMixin = superclass => class extends Localizer(RtlMixin(Pr
 			noFilter: { type: Boolean, attribute: 'no-filter' },
 			noSpellchecker: { type: Boolean, attribute: 'no-spellchecker' },
 			pasteLocalImages: { type: Boolean, attribute: 'paste-local-images' },
+			sourceCode: { type: Boolean, attribute: 'source-code' },
```
grant-cleary, line 97: I've gone back and forth a few times about how to name this property (and how to name the associated context object). I would prefer to figure it out in this PR before putting up the LMS PR because changing it there is more cumbersome.

grant-cleary:
Those visual diff results are interesting, to say the least. Going to investigate those!

grant-cleary:
... I'm dumb. I didn't add the `source-code` attribute to the visual diff test HTML snippets. 🤦

htmleditor-mixin.js
dbatiste: `source-code`, `source-editable` seem fine.  The thing I don't like about `edit-source` is that it makes the property `editSource`, which actually sounds like a method, not a property.

grant-cleary: Ah yeah, good point. I'll do that.

dbatiste: I suppose `source-code`sounds like a property, but not what it represents.  Given that, maybe `source-editable` is the most accurate.  🤷‍♂️

dbatiste (approved):

PR #206, opened 2021-05-14
Add clear() API to reset editor state

grant-cleary:
The quiz-question-authoring FRA uses similar functionality in the FRA editor to reset editor state under some conditions. To help support this, we can add the same API to the new editor to expose this functionality.


htmleditor-mixin.js
```diff
@@ -212,6 +212,14 @@ export const HtmlEditorMixin = superclass => class extends Localizer(RtlMixin(Pr
 		this.addEventListener('d2l-dropdown-close', () => this._dropdownOpen = false);
 	}
 
+	clear() {
+		const editor = tinymce.EditorManager.get(this._editorId);
+		if (!editor) return;
+
+		editor.undoManager.clear();
+		this.html = '';
```
grant-cleary, line 220: This will trigger a re-render, so do it last. Note that this has an interesting side effect. Because the editor contents don't propagate to the component's `html` property until the editor has fired a blur event, invoking this via the console before such a blur event has had time to fire can do some strange things. I could (in theory) fix this by checking `editor.getContent()` and forcing an update of `html` in the process, but this is technically already a problem that _could_ happen with the `html` getter/setter as is.

htmleditor-mixin.js
```diff
@@ -212,6 +212,14 @@ export const HtmlEditorMixin = superclass => class extends Localizer(RtlMixin(Pr
 		this.addEventListener('d2l-dropdown-close', () => this._dropdownOpen = false);
 	}
 
+	clear() {
+		const editor = tinymce.EditorManager.get(this._editorId);
+		if (!editor) return;
+
+		editor.undoManager.clear();
```
grant-cleary, line 219: One thing I'm sort of not clear on here is whether we want to use `clear()` or `reset()` (https://www.tiny.cloud/docs/api/tinymce/tinymce.undomanager/#reset) here. It's not super clear to me from the documentation what the value is of adding that one extra undo level.

htmleditor-mixin.js
dbatiste: Does setting `this.html` result in an undo level being created?  Just wondering if this lines should come before the undo clear.

htmleditor-mixin.js
dbatiste: Yeah I'm not sure either.  The `reset` says that is creates an initial level after clearing, which `clear` does not, and I am not sure what that means in terms of behavior.

htmleditor-mixin.js
grant-cleary: It doesn't appear to. When I go through this workflow I can't undo my removal. I can move it to be extra careful though.

htmleditor-mixin.js
grant-cleary: Yeah, when I go through the workflow using `clear()` everything looks like it works as I would expect. The old FRA editor does a `clear()` and then an `add()`, which mimics a `reset()`, but for the life of me I can't understand why...

dbatiste (approved):

PR #210, opened 2021-05-25
External plugins should not always be loaded

grant-cleary:
It looks like `external_plugins` is meant to be used *in place of* (not _in addition to_) `plugins` when loading external plugins, so we should reflect that in the editor. My goal here was to try to maintain separation of concerns... The toolbars should know which plugins to include, while the editor mixin should know how to map those plugins to their actual locations.


components/toolbar/toolbar-full.js
```diff
@@ -45,6 +45,17 @@ import { getEditorComponent } from '../editor-adapter.js';
...
 import { ToolbarMixin } from './toolbar-mixin.js';
 
+export function getExternalPlugins(component) {
+	return [
+		allowSourceEditing(component) ? 'advcode' : '',
+		'a11ychecker',
+		'advtable',
+		'formatpainter',
+		component.mentions && component._context && component._context.mentions ? 'mentions' : '',
+		'powerpaste'
```
grant-cleary, line 55: So technically all of these plugins were loading all of the time, regardless of whether we needed them. @mentions was the only one that had an actual impact because the others all either rely on toolbar buttons, which we (correctly) don't render if they're not needed, or are enabled everywhere (powerpaste is the only example of this).

Note that this isn't an issue prior to 20.21.6. While these plugins are all always enabled even in prior versions, the actual server-side @mentions query checks the feature flag/config, thus it returns no results without those conditions in place. In other words, we're not leaking the feature; we're just including it in more places than intended in 20.21.6+.

Trandrew17 (approved):

components/toolbar/toolbar-full.js
dbatiste: It's unfortunate that the toolbars need to know the difference between normal plugins and external plugins.  I really wish tinyMCE's API didn't need to differentiate between them.  Overall, I think this is a great net improvement though - to not load these plugins unnecessarily as was done previously.

dbatiste (approved):

components/toolbar/toolbar-full.js
grant-cleary: I was thinking the same thing. I had just assumed that `external_plugins` was meant to be _in addition to_ the default plugin list. Perhaps more reason why we might want to deploy the editor to the CDN. That way we could just merge these external plugins into the "normal" plugins directory and flip back to the way we were loading the plugins before.

PR #223, opened 2021-05-28
Prevent toolbar rendering code from exploding if it runs while the editor container is detached from the DOM

grant-cleary:
If a user's connection is very slow, it's possible that the resize code can trigger while the editor component's children are still detached from the DOM (mid-update). This causes the `allowSourceCodeEditing` check to fail because it's expecting the editor component to be present... But because the editor container is detached from the DOM, `findComposedAncestor` hits the editor container, finds no parent, and returns null.

From my testing, this only happens in an ephemeral state mid-loading, so if we just short-circuit that `allowSourceCodeEditing` check to `false` while we're in this scenario, we'll eventually re-render with the component's contents re-attached to the DOM and everything works out. i.e. I could reproduce this issue on every refresh, but with this change, every refresh leaves me with a functional source code editor button in the toolbar as I would expect.


components/toolbar/toolbar-full.js
```diff
@@ -123,7 +123,8 @@ class Toolbar extends ToolbarMixin(LitElement) {
 
 	render() {
 		const hasContext = hasLmsContext();
-		const showSourceCodeEditor = allowSourceEditing(getEditorComponent(this));
+		const component = getEditorComponent(this);
```
grant-cleary, line 126: This is where the issue arises. This can return `null` if we're at a point in the update where the editor container is detached from the editor's DOM context.

dbatiste: Does the toolbar rendering need to be delayed?  What ensures that this toolbar `render` gets rerun after the container is attached?

grant-cleary: Unsure, actually. In practice I just saw it was happening. The toolbar rendering code was being kicked off multiple times, seemingly because TinyMCE loading was kicking off a resize. I imagine we could similarly delay rendering the toolbar, but I'm not sure how that fits in with the way everything else is loading. e.g. There's other work we already push to the back of the queue via rAF, so I'm not entirely sure if that would work well with the toolbar delayed... Or if the toolbar rendering would trigger another resize that would detach the container from the DOM anyways.

grant-cleary: I can try rAFing the toolbar rendering and see if that has the same effect.

grant-cleary: Unsurprisingly, this doesn't work because we render it as part of the editor's render function.

grant-cleary: I think we end up with a weird chicken/egg scenario here because TinyMCE's initial size is dependent on the toolbar being in place before it renders.

grant-cleary: Never mind, it's not what I thought. Timing just looked that way because some async stuff happens in different orders depending on exactly how much I throttle my connection. I also don't think this solves the problem anyways; something else explodes down the line to prevent TinyMCE from rendering if I throttle enough.

htmleditor.js
```diff
@@ -373,17 +375,19 @@ class HtmlEditor extends HtmlEditorMixin(SkeletonMixin(LitElement)) {
...
-					</div>
-				`;
-			} else {
-				return this._renderEditor(this._toolbar.render(), true);
-			}
+			return html`
+				<div class="d2l-htmleditor-label-flex-container" style="${styleMap(flexContainerStyles)}">
+					<span class="${classMap(labelClasses)}" aria-hidden="true">${this.label && !this.labelHidden ? this.label : ''}</span>
+					${this._renderEditor(this._toolbar.render(), true)}
+				</div>
```
grant-cleary, line 389: If we always render this outer container, TinyMCE won't violently explode when we set the label after first render.

htmleditor.js
```diff
@@ -373,17 +375,19 @@ class HtmlEditor extends HtmlEditorMixin(SkeletonMixin(LitElement)) {
 		const flexContainerStyles = {};
 		if (!this._fullscreen && this.height.includes('%')) flexContainerStyles.height = this.height;
 
+		const labelClasses = {
+			'd2l-input-label': true,
+			'd2l-input-label-required': this.required,
```
grant-cleary, line 380: This is something we missed when first porting the editor over. Required labels should have this CSS applied. The unfortunate thing is that there's no equivalent for screenreader users. This problem isn't limited to the editor, however. Other fields also don't alert screenreaders to required fields, as far as I could tell.

grant-cleary:
Ah, looks like the span is occupying some height regardless of whether there's label text present. Will fix.

grant-cleary:
Merging master into this because I think we likely still want this change, even if we're no longer setting the label in a way that breaks things.

dbatiste:
I don't have any major objections here.  Although the addition of `required` makes me think that we should be trying to use the `FormElementMixin`.  ex. [d2l-input-text](https://github.com/BrightspaceUI/core/blob/master/components/inputs/input-text.js).  I'm not sure how feasible it will be to use that stuff.  There are a number of examples... but let me know if you'd like an overview of it, since it might not be obvious what all is required.  I think Allan put together [some documentation](https://github.com/BrightspaceUI/core/tree/master/components/form).

htmleditor-mixin.js
```diff
@@ -405,6 +444,12 @@ export const HtmlEditorMixin = superclass => class extends Localizer(RtlMixin(Pr
 					this._handleFocusIn();
 				});
 
+				let timerId = 0;
+				editor.on('input setcontent', () => {
+					clearTimeout(timerId);
+					timerId = setTimeout(() => this._validate(false), 250);
```
grant-cleary, line 450: Debounce so we're only running this calculation once every 250 ms. I think we could go higher here if we wanted to (or lower I suppose).

htmleditor-mixin.js
```diff
@@ -213,6 +227,25 @@ export const HtmlEditorMixin = superclass => class extends Localizer(RtlMixin(Pr
...
+				if (this.invalid) body.setAttribute('aria-invalid', true);
+				else body.removeAttribute('aria-invalid');
+			}
+		});
+	}
+
+	updated(changedProperties) {
+		super.updated(changedProperties);
+		changedProperties.forEach((_, propName) => {
+			if (propName === 'html') this._validate(false);
```
grant-cleary, line 247: I don't think this is ever likely to be necessary. Every invocation of the `html` setter will trigger a `setContent` event on the editor. This only serves a purpose if we end up into the `else` block of the `if (window.tinymce)` logic within the setter. I feel like that's incredibly unlikely to happen though; nor do I think we've been so careful in other places outside of that setter.

grant-cleary:
This may need changes depending on how we want to handle conflicts with the existing LMS validation logic.

grant-cleary:
Ok, this PR is good to go as is, @dbatiste. If we're hiding the MVC validation tooltip in the LMS, we don't need to change anything here.

dbatiste (approved):

htmleditor.js
```diff
@@ -360,14 +400,22 @@ class HtmlEditor extends HtmlEditorMixin(SkeletonMixin(LitElement)) {
 			'd2l-htmleditor-resize-handle-fullscreen': this._fullscreen
 		};
 
+		const invalidIconSide = (this.dir === 'rtl') ? 'left' : 'right';
+		const invalidIconStyles = {
```
dbatiste, line 404: Is there a reason we can't do this in CSS?

dbatiste: i.e. something along the lines of...
```javascript
.d2l-htmleditor-invalid-icon {
    right: 12px;
}
:host([dir="rtl"]) .d2l-htmleditor-invalid-icon {
    left: 12px;
    right: 0;
}
```

htmleditor.js
```diff
@@ -414,6 +464,11 @@ class HtmlEditor extends HtmlEditorMixin(SkeletonMixin(LitElement)) {
 		if (this._percentageBasedHeight()) htmlBlockStyles.height = this.height;
 		else htmlBlockStyles.minHeight = this.height;
 
+		const invalidIconSide = (this.dir === 'rtl') ? 'left' : 'right';
```
dbatiste, line 467: Same thing here... if moved to CSS, looks like this isn't needed.

htmleditor.js
grant-cleary: Oh, no. I was thinking that; I just didn't remember to implement it.

README.md
```diff
@@ -55,11 +55,13 @@ Types of editors (toolbar features):
 | `isDirty` | Boolean | Whether or not the editor is [dirty](https://www.tiny.cloud/docs/api/tinymce/tinymce.editor/#isdirty). If `auto-save` is enabled, set `isDirty` to false when triggering workflows where you do not want the autosave prompt. |
 | `label` | String, required | Label for the editor. |
 | `label-hidden` | Boolean | Hides the label visually. |
+| `max-length` | Number, default: null | Maximum length of HTML contents, including HTML control characters. Enables validation on this criterion when set. |
```
dbatiste, line 58: I suppose most people would assume that null means there is no maximum length.  🤷

grant-cleary: Yeah, I wasn't really sure what to choose as a default here. `0` is kind of counter-intuitive, so I didn't want to do that. I guess the alternative is declaring it `undefined`, but that seemed largely the same as `null`.

htmleditor-mixin.js
```diff
@@ -217,6 +231,25 @@ export const HtmlEditorMixin = superclass => class extends Localizer(RtlMixin(Pr
 			this._dropdownOpen = false;
 			this._handleFocusOut();
 		});
+		this.addEventListener('invalid-change', async() => {
+			if (this._initialized) {
```
dbatiste, line 235: Nit: we can reduce nesting here by writing this as `if (!this._initialized) return;`.

htmleditor-mixin.js
```diff
@@ -217,6 +231,25 @@ export const HtmlEditorMixin = superclass => class extends Localizer(RtlMixin(Pr
...
+				const editor = await this._getEditor();
+				if (!editor) return;
+
+				const body = editor.getBody();
+				if (!body) return;
+
+				if (this.invalid) body.setAttribute('aria-invalid', true);
```
dbatiste, line 242: Testing this out in Chrome, it seems ok... just something to be careful of, since attributes are meant to be strings, and `true` is not.

grant-cleary: Ah. Yeah, it worked, but you're right. I'll stringify this.

grant-cleary: Oh, I see. `setAttribute` implicitly stringifies its value argument if it's not already a string. Nonetheless, better to do this explicitly.

htmleditor-mixin.js
```diff
@@ -368,6 +402,11 @@ export const HtmlEditorMixin = superclass => class extends Localizer(RtlMixin(Pr
 				const iframe = editor.getContentAreaContainer().firstElementChild;
 				if (iframe) iframe.title = this.label;
 
+				if (this.required) {
+					const body = editor.getBody();
+					if (body) body.setAttribute('aria-required', true);
```
dbatiste, line 407: Same thing here - normally I am explicit passing a string, but assuming no browsers complain about passing Booleans to `setAttribute` then I am fine with it.

grant-cleary: I changed both of these to be explicit anyways. It looks like the spec suggests browsers _should_ just `toString()` whatever value gets passed in here, but I prefer being explicit.

README.md
dbatiste: Yeah I think `undefined` or `null` is fine.  I was just wondering if the readme needed it spelled out in effective terms.

grant-cleary: Oh, I see. Yeah, I mean there's no harm in being more explicit.

dbatiste (approved):

PR #232, opened 2021-06-04
US126879 - Switch to dynamic localization as a preface to adding OSLO support

grant-cleary:
This is a pre-requisite to make life easier when adding OSLO support. Switching to dynamic localization makes adding OSLO support quick and easy (just a one-line config change).


lang/localizer.js
```diff
@@ -1,77 +1,31 @@
-import { LocalizeMixin } from '@brightspace-ui/core/mixins/localize-mixin.js';
+import { LocalizeDynamicMixin } from '@brightspace-ui/core/mixins/localize-dynamic-mixin.js';
 
-export const Localizer = superclass => class extends LocalizeMixin(superclass) {
+// Map transformed languages to avoid 404s trying to fetch non-existent files.
+const langMappings = {
+	'ar-sa': 'ar',
```
grant-cleary, line 5: Don't ask me how copy-paste managed to produce different spaces vs. tabs on different lines in the same block. Sigh. Will fix.

lang/localizer.js
```diff
@@ -1,77 +1,31 @@
-import { LocalizeMixin } from '@brightspace-ui/core/mixins/localize-mixin.js';
+import { LocalizeDynamicMixin } from '@brightspace-ui/core/mixins/localize-dynamic-mixin.js';
 
-export const Localizer = superclass => class extends LocalizeMixin(superclass) {
+// Map transformed languages to avoid 404s trying to fetch non-existent files.
+const langMappings = {
```
grant-cleary, line 4: This isn't strictly necessary for localization to work properly. If we try to dynamically import `en-us`, for example, that file doesn't exist because we actually map it to `en`, but the localization code does fall back to `en` and load up that file regardless. However, trying to import the non-existent file produces a console error. This is ugly and also causes issues when integrated in the LMS because some of our UI tests will fail if pages start producing errors (even benign ones). So... May as well just map these langs here and avoid that issue altogether.

lang/localizer.js
```diff
@@ -1,77 +1,31 @@
...
-			}
-		}
-		translations = await import('./en.js');
+export const Localizer = superclass => class extends LocalizeDynamicMixin(superclass) {
+
+	static get localizeConfig() {
 		return {
-			language: 'en',
-			resources: translations.default
+			importFunc: async lang => {
```
dbatiste, line 25: What happens if we don't have a mapping, or a translation file for the specified `lang`?  Previously I think it would have fallen back to `en`, but now it looks like it would produce a 404 error, and maybe also a JavaScript error.

grant-cleary: It produces a 404 error (which does show up in the console), but it seems to still do the fallback properly. If I remove the mapping, everything still seems to work properly.

grant-cleary: I'm pretty sure there's a catch block in the localization code that handles this; I suspect Chrome just tosses network errors into the console regardless even if they're caught exceptions.

lang/localizer.js
```diff
@@ -1,77 +1,14 @@
...
-		translations = await import('./en.js');
+	static get localizeConfig() {
 		return {
-			language: 'en',
-			resources: translations.default
+			importFunc: async lang => {
+				lang = langMappings[lang] || lang;
+				return (await import(`./${lang}.js`)).default;
+			},
+			osloCollection: 'my-project\\htmleditor',
```
grant-cleary, line 11: Damn, this snuck in by accident.

dbatiste (approved):

lang/localizer.js
grant-cleary: Yeah, this mapping stuff we were doing before doesn't exist anymore. I believe I asked about this behaviour and once I confirmed that it only happened in dev, I stripped out the mapping.

grant-cleary: It's slightly unfortunate that there's no way to suppress the 404 in the console, since it does lead to a tiny bit of extra noise when hosting locally or when running tests, but I barely notice it anymore, so 🤷.

PR #234, opened 2021-06-07
Remove nullish coalescing operator from editor

grant-cleary:


htmleditor-mixin.js
```diff
@@ -18,7 +18,7 @@ const rootFontSize = window.getComputedStyle(document.documentElement, null).get
 const pathFromUrl = url => url.substring(0, url.lastIndexOf('/'));
 const baseImportPath = pathFromUrl(import.meta.url);
 
-const documentLang = (document.documentElement.getAttribute('lang') ?? 'en').replace('-', '_');
+const documentLang = (document.documentElement.getAttribute('lang') || 'en').replace('-', '_');
```
grant-cleary, line 21: I realize `||` is less specific, but I think it's okay here if we're also doing this for all false-y things.

dbatiste (approved):

grant-cleary:
... Strange. The failure suggests there's something not great about this syntax. I'll take a quick peek.

grant-cleary:
Uh, never mind. I think our linting is broken?

PR #241, opened 2021-06-09
US126879 - Convert lang terms to Oslo-friendly format (and fix some broken translations)

grant-cleary:
This is another pre-requisite for getting Oslo working. We need to convert all of our lang terms into Oslo-friendly formats (which means no `.` at all). I also needed to correct a couple broken translations that were causing explosions through the editor, preventing me from testing anything. I did so by correlating the brokenness with what was present in the LMS translations before. This isn't ideal and I'm following up to find an actual solution, but I need this to at least work without crashing Chrome with a billion and one console errors.


grant-cleary:
I also opted to remove the fr-on lang pack in this PR. I see no reason to keep it up to date knowing that I would be removing it in the next PR anyways.

htmleditor.js
```diff
@@ -410,7 +410,7 @@ class HtmlEditor extends HtmlEditorMixin(SkeletonMixin(LitElement)) {
 					class="d2l-offscreen d2l-htmleditor-inline-button"
 					@click="${this._onInlineEditClick}"
 					@focus="${this._onInlineEditButtonFocus}">
-					${this.localize('inline\\button', { name: this.label })}
+					${this.localize('inline:button:', { name: this.label })}
```
dbatiste, line 413: Is this one supposed to end with `:`?

components/toolbar/table-create.js
```diff
@@ -102,7 +102,7 @@ class TableCreateView extends ToolbarItemMixin(HierarchicalViewMixin(LitElement)
 		const renderColumns = rowIndex => {
 			const columns = new Array(TOTAL_COLUMNS);
 			for (let colIndex = 0; colIndex < TOTAL_COLUMNS; colIndex++) {
-				const text = this._localize('table.insert.colums-by-rows', 'columns', colIndex + 1, 'rows', rowIndex + 1);
+				const text = this._localize('table:insert:columsByRows', { 'columns': colIndex + 1, 'rows': rowIndex + 1 });
```
dbatiste, line 105: So this was being done incorrectly before?

grant-cleary: Possibly. This function definitely expects an `Object` as its second argument and I'm unsure whether the JS engine is clever enough to convert the parameter list as appropriate. I figured I wouldn't test setting it back, since this is the format used elsewhere in these sorts of substitutions.

dbatiste (approved):

PR #244, opened 2021-06-10
US126879 - Publish serge mapping for Oslo integration

grant-cleary:
We need to publish our serge mapping file to the NPM package for the Oslo integration to function properly (probably should have seen that coming given that it needs to communicate with the LMS 🤦).


grant-cleary:
This needs a change to the htmleditor to publish the serge mappings file to the actual package. I have a PR open for that here: https://github.com/Brightspace/brightspace-integration/pull/4057

dbatiste (approved):

PR #257, opened 2021-06-15
Make sure toolbar menus and options open as expected and don't close unexpectedly

grant-cleary:
This is a collection of changes aimed at fixing up some issues with toolbar buttons in the inline/inline-limited editor (raised internally by teams looking to use it). I'll comment more about specific fixes inline.


components/toolbar/toolbar-inline-limited.js
```diff
@@ -35,6 +36,7 @@ export function getPlugins(component) {
 	return [
 		addColorPlugin(),
 		component.fullPage ? addFullPagePlugin() : '',
+		addQuicklinkPlugin(),
```
grant-cleary, line 39: This plugin was just missing, so the button didn't do anything.

components/toolbar/toolbar-inline.js
```diff
@@ -69,7 +69,7 @@ class Toolbar extends ToolbarMixin(LitElement) {
 				<d2l-dropdown-menu align="start" min-width="232" no-pointer vertical-offset="-1">
 					<d2l-menu label="${this.localize('insert')}">
 						${hasContext ? html`
-							<d2l-htmleditor-menu-item icon="table" text="${this.localize('equation')}">
+							<d2l-htmleditor-menu-item icon="equation-graphical" text="${this.localize('equation')}">
```
grant-cleary, line 72: Speaks for itself I think; just a copy-paste error on the icon used.

htmleditor-mixin.js
```diff
@@ -209,7 +209,10 @@ export const HtmlEditorMixin = superclass => class extends Localizer(RtlMixin(Pr
 		this.addEventListener('focusin', () => this._handleFocusIn());
 		this.addEventListener('focusout', () => this._handleFocusOut());
 		this.addEventListener('d2l-dropdown-open', () => this._dropdownOpen = true);
-		this.addEventListener('d2l-dropdown-close', () => this._dropdownOpen = false);
+		this.addEventListener('d2l-dropdown-close', () => {
+			this._dropdownOpen = false;
+			this._handleFocusOut();
```
grant-cleary, line 214: This is the bit that may be controversial. The issue here is that the hierarchical view underlying our dropdowns removes the tabindex on its opener when the opener is clicked. This leaves the opener unfocusable, so a `focusout` event fires and triggers the inline editor to collapse.

We can relatively easily fix this by checking if a dropdown is open before triggering our `focusout` handler, but because the `d2l-dropdown-close` event fires _after_ the `focusout` event, this means clicking into a dropdown and then clicking out of the editor closes the dropdowns, but doesn't collapse the editor. So my "fix" here is to treat the dropdown close event as a _potential_ focusout event, allowing our existing "does this actually have focus" checks to protect us from surprise collapses.

htmleditor-mixin.js
```diff
@@ -264,14 +267,15 @@ export const HtmlEditorMixin = superclass => class extends Localizer(RtlMixin(Pr
...
-			if (focusedWithin) return;
+			if (focusedWithin || this._dropdownOpen) return;
 
 			if (this._initialized) {
 				const editor = await this._getEditor();
 				if (editor.hasFocus()) return;
 			}
 
 			if (document.activeElement.classList.contains('d2l-dialog-frame')) return;
+			if (this.shadowRoot.querySelector('tox-dialog')) return;
```
grant-cleary, line 278: I'm uncertain if this is exactly what we want. TinyMCE's dialogs create a shim that (usually) prevents the user from clicking out of the editor, thus losing focus and collapsing the inline editor. This does not appear to happen in FRAs, so as extra protection I think it would be good to avoid collapsing the editor if a TinyMCE dialog is currently open (as best we can at least). Without this, a user can click out of the editor in an FRA and collapse the editor while the dialog is still open. This leaves the dialog partially rendered but uninteractable, so the editor is basically broken.

As a sidenote, it's probably still worth looking into why the TinyMCE shim isn't set properly (possibly some kind of z-index issue or something specific to nested iframes), but I think this is a good change regardless for some extra protection.

htmleditor-mixin.js
dbatiste: I am a little confused by this:

>We can relatively easily fix this by checking if a dropdown is open before triggering our focusout handler, but because the d2l-dropdown-close event fires after the focusout event, this means clicking into a dropdown and then clicking out of the editor closes the dropdowns, but doesn't collapse the editor. 

I would have thought that if `focusout` fires first, the editor would toggle to read mode (collapses the editor).

grant-cleary: Right, it's paired with the actual fix here that this causes a problem: https://github.com/BrightspaceUI/htmleditor/pull/257/files#diff-4c87bd59fc7b2230dce410e378811fb29001a2b37c9561dfd8569ff64517f299R270

Testing locally, it appears that when the dropdown loses focus, it fires off a close event. If the dropdown losing focus also causes the editor to lose focus (clicking out of the editor, for example), then `this._dropdownOpen` still returns true (because `focusout` fires first) even though we're about to get a dropdown close event. So more or less, if the editor and dropdown simultaneously lose focus, we have no idea in the editor's `focusout` handler whether we should remain open (because there are active dropdowns open) or if we should close (because the dropdown is about to close).

dbatiste (approved):

PR #260, opened 2021-06-15
DE43725 - Dismiss open menus when clicking into the editor

grant-cleary:
Because the editor contents are a different document (inside an iframe), we don't register a `focusout` event on the menu dropdowns when someone clicks into the editor. Thus, this PR adds an `onclick` handler to the editor that closes any open menus (including propagating properties through our button/toolbar elements to allow that).

I was originally hoping to tie into the `focus` event, but we get stuck with the whole different document thing if we try that, since if someone is clicked into the editor and then opens a menu in the toolbar, TinyMCE doesn't register that the inner document has lost focus. I think this is fine, because in my testing, it didn't seem like it was possible to use keyboard nav to remove focus from a dropdown without either leaving the editor entirely or explicitly closing it.


.eslintrc.json
```diff
@@ -3,7 +3,7 @@
   "globals": {
     "tinymce": false
   },
-  "ignorePatterns": ["tinymce", "icons.js"],
+  "ignorePatterns": ["tinymce", "icons.js", "test/screenshots/*"],
```
grant-cleary, line 6: Unrelated to this change, but the new eslint lit plugin version doesn't like some formatting in the report file (HTML) that gets placed in this directory after running visual diff tests locally.

components/toolbar/toolbar-full.js
```diff
@@ -45,6 +45,12 @@ import { getEditorComponent } from '../editor-adapter.js';
 import { hasLmsContext } from '../lms-adapter.js';
 import { ToolbarMixin } from './toolbar-mixin.js';
 
+const elementName = 'd2l-htmleditor-toolbar-full';
+
+export function getElementName() {
+	return elementName;
+}
```
grant-cleary, line 52: This was my attempt to solve the whole "the mixin shouldn't know about the toolbar" thing. I don't know if we care about that separation anymore. :shrug:

components/toolbar/button-menu.js
```diff
@@ -91,6 +91,10 @@ class ButtonMenu extends DropdownOpenerMixin(ToolbarItemMixin(FocusVisiblePolyfi
 		this._hasSlottedIcon = false;
 	}
 
+	get isOpen() {
```
dbatiste, line 94: Normally these properties are just named without the is... `open`.  It stems from the fact that commonly, such properties also have corresponding attributes, and we wouldn't name the attribute `is-open`... but rather... there would just be an `open` attribute, or there wouldn't.

components/toolbar/toolbar-full.js
dbatiste: I think that's a good approach.  A super nit-picky thing would be... I think it would be more correct to refer to this as the tag name, rather than the element name.  It really is bike-shedding to the max though.

dbatiste (approved):

components/toolbar/button-menu.js
grant-cleary: Ah, gotcha. I can easily change this. I keep thinking of these getters as functions.

components/toolbar/toolbar-full.js
grant-cleary: Easy enough to change anyways. 👍

PR #261, opened 2021-06-16
Use correct query selector when determining if a TinyMCE dialog is open on focusout

grant-cleary:
Enough said, I think. :)


grant-cleary:
I'm going to ignore the visual diff tests for now. The Puppeteer update appears to have made them all flaky for reasons I don't understand.

dmd2l (approved):

PR #263, opened 2021-06-17
DE44038 - Pass orgUnitId to @mentions search query in query string

grant-cleary:
Just a bit of cleanup here, moving to a more accurate controller and passing in the org unit ID via query string instead of as part of the object.


klhuff (approved):

grant-cleary:
Visual diff tests are just broken at the moment and I haven't found the time to disable them yet. Will ignore again.

Trandrew17 (approved):

dbatiste (approved):

PR #265, opened 2021-06-18
Pass orgUnitId to @mentions search query in query string

grant-cleary:
I'm not in any particular rush to merge this. 20.21.7 lang term changes are going to be arriving at some point and I don't really want to cut a BSI release until those come in. I do want this change in there when we do cut one though.


test/htmleditor.visual-diff.js
```diff
@@ -1,4 +1,4 @@
-const puppeteer = require('puppeteer');
+/*const puppeteer = require('puppeteer');
```
klhuff, line 1: I think we decided it was ok before, but I don't know that I like to see a block of code commented out.

klhuff: I guess this is a reminder that we should look into the failures, or delete it.

klhuff (approved):

Trandrew17 (approved):

test/htmleditor.visual-diff.js
grant-cleary: This is more or less equivalent to `ExcludeFromBuild`. There's just no formal way to accomplish that in the visual diff framework. These tests are flaky and need to be looked at, but not right now.

PR #267, opened 2021-06-18
DE30771 - Add alt text to rendered equation images after render

grant-cleary:
When we render MathML images produced by the WIRIS service in the editor, we also want these images to have the math role and good accessible text so that assistive technologies can read the equations. We can use one of WIRIS' services to do this. Unfortunately, because TinyMCE doesn't process asynchronous event handlers in the way we would like (they won't pause rendering to wait for them to resolve), we need to wait for TinyMCE to render the DOM and then edit the image elements directly.


components/plugins/equation.js
```diff
@@ -80,6 +81,14 @@ export function addEquationPlugin() {
 		editor.addQueryStateHandler(cmds.equationMathML, () => isEquationSelected(editorTypes.MathML));
 		editor.addQueryStateHandler(cmds.equationChemistry, () => isEquationSelected(editorTypes.Chemistry));
 
+		const trimMml = (mml, removeAnnotation, removeTitle) => {
+			let trimmedMml = mml.replace(/>\s+</g, '><');
+			if (removeTitle) trimmedMml = trimmedMml.replace(/title=(\"|\')([^\"]*)(\"|\')/i, '');
```
grant-cleary, line 86: WIRIS' `mathml2accessible` API does not seem to like when a title attribute is included on the `math` tag. I figured it would also be worth still stripping out the annotation and any whitespace while we were in here since it's irrelevant for the API (thus splitting this out into its own function).

components/plugins/equation.js
```diff
@@ -164,10 +169,26 @@ export function addEquationPlugin() {
 			eventContext.content = eventContext.content.replace(/<img[^>]*\s+data-d2l-mathml=\"[^\"]*\"[^>]*>/gi, func);
 		};
 
+		const updateImageAltText = async() => {
+			const body = editor.getBody();
+			for (const imgNode of body.querySelectorAll('img[data-d2l-mathml]')) {
+				if (imgNode.getAttribute('alt')) continue;
```
grant-cleary, line 175: I figure if the image already has an alt tag, we probably don't want to wipe it out.

components/plugins/equation.js
```diff
@@ -164,10 +169,26 @@ export function addEquationPlugin() {
...
+
+				const decodedTrimmedMml = trimMml(decodeURIComponent(`${imgNode.getAttribute('data-d2l-mathml')}`), true, true);
```
grant-cleary, line 177: Having to decode and re-encode the math here sucks, but we need to do that for the trimming logic to work properly. Otherwise we would have to write a new regex to look for all of the URL encoded equivalents here. The alternative would be to keep a reference to all of the unencoded MathML objects when processing them at first in `BeforeSetContent` and then just iterating through them here.

components/plugins/equation.js
```diff
@@ -164,10 +169,26 @@ export function addEquationPlugin() {
...
+					if (response.ok) response.text().then(text => imgNode.setAttribute('alt', text));
+				});
+			}
+		};
+
 		editor.on('BeforeSetContent', (e) => {
 			convertToImages(e);
 		});
 
+		editor.on('SetContent', async() => {
```
grant-cleary, line 188: When the `SetContent` event fires, the editor contents have been fully rendered and attached to the DOM, so it's safe for us to do the kind of DOM manipulation we're doing here.

grant-cleary: Note that because TinyMCE doesn't block its lifecycle events when processing an async handler, this does come with a bit of danger. Specifically, if two `SetContent` events came in quickly enough that the second fired before the first finished, this handler would start running again and could potentially clobber the code running in the first handler. There's not really much we can do about that though and I feel like the risk is relatively low given how little this code actually does.

components/plugins/equation.js
```diff
@@ -80,6 +81,14 @@ export function addEquationPlugin() {
 		editor.addQueryStateHandler(cmds.equationMathML, () => isEquationSelected(editorTypes.MathML));
 		editor.addQueryStateHandler(cmds.equationChemistry, () => isEquationSelected(editorTypes.Chemistry));
 
+		const trimMml = (mml, removeAnnotation, removeTitle) => {
+			let trimmedMml = mml.replace(/>\s+</g, '><');
+			if (removeTitle) trimmedMml = trimmedMml.replace(/title=(\"|\')([^\"]*)(\"|\')/i, '');
+			if (removeAnnotation) trimmedMml = trimmedMml.replace(/<annotation[\s\S]*<\/annotation>/g, '');
```
klhuff, line 87: This is a nit, but, is it normal for us to do one liners in ifs like this (ie. without the braces)? For c#, I think we have an analyzer that will break compile on something like this.

components/plugins/equation.js
klhuff: Presumably this prevents this from running every time the HTML editor loads, since after the first time you'd expect the alt text to be saved?

components/plugins/equation.js
klhuff: How does this work if I add a new equation to the editor? I'm assuming the SetContent must happen every time you add "content" to the editor (with some sort of delay if I'm just typing text into it? Maybe when the editor comes out of focus?).

components/plugins/equation.js
grant-cleary: No, the alt text never gets saved. When the `PostProcess` event fires, we convert all of this back to MathML and that's what actually gets stored. This is only done while the images are in the editor proper. This exists in case this code runs twice so we don't send out the math for processing if we don't need to.

components/plugins/equation.js
grant-cleary: So `SetContent` is fired any time something _externally_ sets content in the editor (typically via TinyMCE's `setContent` API, but other APIs can trigger it). This happens on load as well as in this plugin when someone has added an equation.

components/plugins/equation.js
grant-cleary: This is normal in our JS stuff, yeah. Formatting/styling rules are managed via the various eslint plugins.

klhuff (approved):

dbatiste (approved):

components/plugins/equation.js
```diff
@@ -164,10 +169,26 @@ export function addEquationPlugin() {
 			eventContext.content = eventContext.content.replace(/<img[^>]*\s+data-d2l-mathml=\"[^\"]*\"[^>]*>/gi, func);
 		};
 
+		const updateImageAltText = async() => {
```
dbatiste, line 172: This is defined as an `async` method, which means it will return a promise and `await` can be used within, but in this case it will be resolved right away because we are not `await`ing anything here, and we are also not returning the promise that `fetch` returns.

It appears that this should not have `async`, and the promise from the `fetch` should be returned.  Thoughts?

grant-cleary: You're right; this is a mistake left over from when I was `await`ing the return from `fetch` instead of `then`ing it.

ca2hill (approved):
I was not useful in this review.

grant-cleary:
> I was not useful in this review.

... So the usual then? :trollface:

PR #276, opened 2021-06-24
Disabled default ifrau resizing behaviour

ronan-f:
This change is to explicitly disable the default ifrau resizing behaviour and instead leaving it up to each app to decide how it wants to resize. It was brought about as in Activity Feed, the default height calculation used doesn't work reliably for us. 

Grant mentioned that there are other teams which are using the html editor within an iframed app so this change _may_ have an impact on their functionality?


grant-cleary (approved):
LGTM. I've put out a call to other devs who may run into this, so we may want to leave it open until tomorrow or something for comment.

ronan-f:
> LGTM. I've put out a call to other devs who may run into this, so we may want to leave it open until tomorrow or something for comment.

Sounds good. I'll leave it open for a while then for other FRA teams to have a look.

mdgbayly:
Just as an aside, one thing we noticed when diagnosing this issue is that the way the ifrauclient is being used in the editor, every time you set up a new ifrauclient, I believe it is setting up a new iframe-resizer connection so the number of resizer messages being sent between the host and the iframe just keeps getting bigger and bigger. You can see this by monitoring the post message events in browser dev tools. 

`monitorEvents(window, ["message"])`

After setting up the monitoring, try opening/closing the quick link dialog a few times and see how the number of messages grows and grows.

It's already quite chatty with just one iframe-resizer. I don't think the underlying `iframe-resizer` library expects you to setup 
multiple connections for a single iframe.

grant-cleary:
@ronan-f I think it's been enough time, so I imagine you can probably merge this now. 😄

PR #277, opened 2021-06-24
DE30771 - Pass document language to WIRIS when requesting accessible text

grant-cleary:
WIRIS accepts ISO-639 lang codes as a parameter on this call, so we should pass the document language along so the accessible text can be translated into (hopefully) the right language. It looks like WIRIS falls back to `en` if it gets a language it doesn't know, so I think that's probably fine in the grand scheme of things.


grant-cleary:
This may need changes depending on how we want to handle conflicts with the existing LMS validation logic.

components/plugins/equation.js
```diff
@@ -175,7 +175,10 @@ export function addEquationPlugin() {
 				if (imgNode.getAttribute('alt')) continue;
 
 				const decodedTrimmedMml = trimMml(decodeURIComponent(`${imgNode.getAttribute('data-d2l-mathml')}`), true, true);
-				const response = await window.fetch(`${wirisPath}/mathml2accessible?mml=${encodeURIComponent(decodedTrimmedMml)}`);
+				const lang = document.documentElement.getAttribute('lang') || 'en';
```
klhuff, line 178: Does this come from the browser? Or are we setting it somehow?

klhuff (approved):

ca2hill (approved):

components/plugins/equation.js
grant-cleary: It comes from the parent document in the DOM.

klhuff: As an aside: I think there are going to be cases where this doesn't work well. But that's because I think the mml has chances to be big, and IIS isn't great at dealing with extremely long query strings.

grant-cleary: So "it depends". In Content, that may come from someone setting it manually on the HTML file. Outside of that, it comes (somewhere) from our localization code.

grant-cleary: > As an aside: I think there are going to be cases where this doesn't work well. But that's because I think the mml has chances to be big, and IIS isn't great at dealing with extremely long query strings.

Yes. This is generally a problem with WIRIS' APIs because they expect these details to be passed to them via the query string. That's partially why we trim the HTML. Notably, this (and the old editor) do the same thing for rendering the images themselves.

grant-cleary: I'm also not sure whether WIRIS respects the full four-character codes. Their documentation only shows the use of the more limited two-character codes. I did some testing with their API and the four-letter codes work, but I'm not sure if they're just ignoring the regional flags. So that might be a thing.

PR #280, opened 2021-06-25
LCS Insert Element Plugin

jschlumkoski:


tinymce/plugins/noneditable/index.js
```diff
@@ -0,0 +1,7 @@
+// Exports the "noneditable" plugin for usage with module loaders
```
grant-cleary, line 1: You don't need to include the `noneditable` folder here. Because it's one of their open source plugins, it's already pulled in via NPM and you'll just need to import it into the toolbar the way other open source plugins (e.g. link) are imported.

jschlumkoski: @grant-cleary The most recent commit should resolve this. I deleted the local noneditable files and imported it like you suggested.

components/plugins/lcs.js
```diff
@@ -0,0 +1,844 @@
...
+import { live } from 'lit-html/directives/live.js';
+import { requestInstance } from '@brightspace-ui/core/mixins/provider-mixin.js';
+import { getUniqueId } from '@brightspace-ui/core/helpers/uniqueId.js';
+
+
+const name = 'lcs-insertelement';
+
+export function addLCSInsertElementPlugin() {
+   tinymce.PluginManager.add(name, function (editor) {
+      if (!hasLmsContext()) return;
```
grant-cleary, line 32: This is going to need to change due to some recent changes in the editor, unfortunately. e.g. https://github.com/BrightspaceUI/htmleditor/blob/master/components/plugins/equation.js#L20

components/plugins/lcs.js
```diff
@@ -0,0 +1,844 @@
...
+         _properties: { type: Array },
+         _items: { type: Array },
+         _editHTML: { type: String },
+         _loadedHTML: { type: Boolean }
+      };
+   }
+
+   static get styles() {
+      return [inputLabelStyles, inputStyles, radioStyles, selectStyles, css`
+			.d2l-htmleditor-templates-field,
```
grant-cleary, line 178: I think linting is likely to get mad about the spaces as indents in some of these.

components/plugins/lcs.js
```diff
@@ -0,0 +1,844 @@
...
+      this.opened = false;
+      this._stepIndex = -1;
+      this._templateIndex = -1;
+      this._properties = [];
+      this._build = {};
+      this._items = [];
+      this._editHTML = '';
+      this._loadedHTML = false;
+      this._resizeObserver = new ResizeObserver(entries => {
+         for (let entry of entries) {
```
grant-cleary, line 286: I think this can be `const entry of entries`, since I don't think you're modifying `entity` in the loop.

components/plugins/lcs.js
```diff
@@ -0,0 +1,844 @@
...
+                  if (property.selector) {
+                     selector = property.value;
+                     break;
+                  }
+               }
+            }
+
+            templateHTML = this._buildHTML(template, selector);
+
+            let iframeStr = stylesStr + '<div class="container-fluid"><div class="row"><div class="col-12">' + templateHTML + '</div></div></div>' + scriptsStr + resizeScript;
```
grant-cleary, line 404: Consider using `const` here as well, and I would maybe use a templated string here. e.g. `${stylesStr}<div ...>${templateHtml} ...`

components/plugins/lcs.js
```diff
@@ -0,0 +1,844 @@
...
+
+   updated() {
+      const e = this.shadowRoot.querySelector('d2l-dialog').querySelector('.editor');
+      if (e !== null) {
+         this._resizeObserver.observe(e)
+      }
+
+      const d = this.shadowRoot.querySelector('d2l-dialog');
+      if (d !== null) {
+         this._resizeObserver.observe(d)
```
grant-cleary, line 442: I _think_ it's better to attach this in `connectedCallback()` and then detach it in `disconnectedCallback()`.

components/plugins/lcs.js
```diff
@@ -0,0 +1,844 @@
...
+            form.setAttribute('submitcheck', true)
+            form.addEventListener('d2l-form-submit', function () {
+
+               e._storeValues();
+
+               e._stepIndex += 1;
+            });
+         }
+
+         if (form.getAttribute('invalidcheck') === null) {
```
grant-cleary, line 786: What's this stuff for?

components/toolbar/toolbar-full.js
```diff
@@ -123,7 +127,8 @@ class Toolbar extends ToolbarMixin(LitElement) {
 
 	render() {
 		const hasContext = hasLmsContext();
-		const showSourceCodeEditor = allowSourceEditing(getEditorComponent(this));
+		const component = getEditorComponent(this);
+		const showSourceCodeEditor = allowSourceEditing(component);
```
grant-cleary, line 131: This stuff will also need to change to account for the change to `hasLmsContext`. If you take a look at what's done here in the main branch you can get an idea of what this should look like. You may also want to consider a helper function like `allowSourceEditing` just to clean up the checks you have in here and in the plugins list.

components/plugins/lcs.js
jschlumkoski: @grant-cleary So this is for form validation on the editor. We don't want the user progressing mostly if required fields are left empty. So in the properties step (line 778), which is the main editing view, I add the event listeners to the d2l-form and do a submit on it. There's also some logic to make sure I don't add it twice. If it's valid, we let the user move to the preview, otherwise we just need to do nothing. 

In hindsight, I might not need to add the event listener for an invalid form, as it's not actually handling anything.

grant-cleary: Yeah, that's what I'm wondering. I think the empty listener can just be removed.

components/toolbar/toolbar-full.js
jschlumkoski: Did both of these things in my most recent commit.

components/plugins/lcs.js
jschlumkoski: Good call. Removed it in my most recent commit!

components/toolbar/toolbar-full.js
```diff
@@ -67,6 +70,7 @@ export function getPlugins(component) {
 		addFullscreenPlugin(),
 		addImagePlugin(),
 		addISFPlugin(),
+		component.fullPage && allowInsertElement() ? addLCSInsertElementPlugin() : '',
```
dbatiste, line 73: The `allowInsertElement()` needs to be `allowInsertElement(component)`, otherwise `allowInsertElement` produces a JS error (observed in the demo page).

components/plugins/lcs.js
```diff
@@ -0,0 +1,839 @@
...
+         }
+
+         d2l-menu-item:before {
+            content: '';
+            height: 18px;
+            margin-right: 15px;
+            width: 18px;
+         }
+
+         d2l-menu-item.accordion:before {
```
dbatiste, line 199: @kathleenmurphy, @geurts - do you have a preference on which semantics should be used here (menu vs list)?  Currently `d2l-menu` is being used.  Either is viable, but in either case, some changes will be needed. For context...

image

Menu...

Using `d2l-menu` means that keyboard users will navigate with arrow-keys, and they will be announced to screen reader users as a menu.  This may be preferred, however the `d2l-menu-item` (intentionally) doesn't provide much flexibility to control the layout as @jschlumkoski discovered. Typically, we want consistent menu-items, which is why the built-in `d2l-menu-item` is fairly locked down.  

If a custom layout is required for a menu-item, the [MenuItemMixin](https://github.com/BrightspaceUI/core/blob/master/components/menu/menu-item-mixin.js) can be used as seen in [this example](https://github.com/BrightspaceUI/core/blob/master/components/menu/demo/custom-menu-item.js).  This approach gives the ultimate flexibility to lay out menu items as required, for example including an icon. One key requirement however, is that menu-items must never contain other focusable or interactive elements aside from a nested menu.

List...

Using `d2l-list` provides list semantics and will be announced as such to screen reader users. It provides a lot of options for layout, such as specifying an image to be displayed on the LHS, control over the list-item content, and many other things.  They can be configured to act like buttons in a similar way as menu-items. We have a related helper component, `d2l-list-item-content` that provides layout including primary, secondary, and supporting info.  Typically, a list is what we've used for this sort of thing (like the QuickLinks dialog). Here is a screenshot in case you aren't familiar.

image

components/plugins/lcs.js
```diff
@@ -0,0 +1,839 @@
...
+            }
+            if (requiresContentUpdate) e.target.setContent(editorDocument.documentElement.outerHTML);
+         }
+      });
+
+      editor.addCommand(cmds.insertElement, () => launchEditor());
+
+      editor.ui.registry.addButton('editlcs', {
+         icon: 'edit-block',
+         tooltip: 'Edit',
```
dbatiste, line 111: Looks like we'll need to add some things to be languified. This is where the terms get specified, and we'll need to use the localize method here instead.

https://github.com/BrightspaceUI/htmleditor/blob/master/lang/en.js

components/plugins/lcs.js
```diff
@@ -0,0 +1,839 @@
...
+      
+         .editor-add > d2l-button-subtle {
+            display: inline-block;
+         }
+
+			.editor-info {
+				margin: 0 15px;
+			}
+
+         .editor-info > h2 {
```
dbatiste, line 235: Note: our typography provides some shared styles that we use for styling headings.

https://github.com/BrightspaceUI/core/blob/master/components/typography/styles.js#L106

They are intended to allow using the correct semantic heading level, but a different class name can be applied to achieve the required "look" (size and margins).  For example, it would be possible to have `<h2 class="d2l-heading-3">some heading</h2>`. Either way, we should always be applying one of those heading level classes so that all our headings follow our typography.

components/plugins/lcs.js
```diff
@@ -0,0 +1,839 @@
...
+
+         .item-form > [slot="content"] {
+            flex-grow: 1;
+         }
+
+         .hidden {
+            display: none;
+         }
+
+         .sr-only {
```
dbatiste, line 262: [These offscreen styles](https://github.com/BrightspaceUI/core/blob/master/components/offscreen/offscreen.js#L4) should be imported, included, and `d2l-offscreen` used instead.

components/plugins/lcs.js
```diff
@@ -0,0 +1,839 @@
...
+      this._editHTML = '';
+      this._loadedHTML = false;
+      this._resizeObserver = new ResizeObserver(entries => {
+         for (const entry of entries) {
+            if (entry.target.tagName === 'D2L-DIALOG') {
+               const ifrm = this.shadowRoot.querySelector('#editor-preview');
+
+               if (ifrm !== null) {
+                  if (ifrm.offsetWidth >= 768) {
+                     this.shadowRoot.querySelector('.disclaimer').classList.add('hidden');
```
dbatiste, line 292: One of the golden rules with Lit is that the DOM should never be modified outside of `render`.  The UI should always be a function of the data (properties), and the inverse should always be true as well.  To apply the hidden class, a private property should be defined, and then conditionally applied in `render`. Lit automatically runs its render lifecycle whenever properties change.

components/plugins/lcs.js
```diff
@@ -0,0 +1,839 @@
...
+                  }
+                  else {
+                     this.shadowRoot.querySelector('.disclaimer').classList.remove('hidden');
+                  }
+               }
+            }
+            else {
+               const dialog = this.shadowRoot.querySelector('d2l-dialog');
+               dialog.shadowRoot.querySelector('dialog').style.overflow = 'hidden';
```
dbatiste, line 301: What's the rationale for hiding overflow for the dialog?  This would also fall into the category of not modifying the DOM outside of `render`.

components/plugins/lcs.js
```diff
@@ -0,0 +1,839 @@
...
+   render() {
+      let content;
+
+      if (this._editHTML !== '' && this._loadedHTML === false) {
+         this._parseHTML(this._editHTML);
+      }
+
+      if (this._templateIndex === -1) {
+         content = html`
+         	<d2l-menu>
```
dbatiste, line 317: Note: top-level menus should always have a `label` attribute.  It is not visible to sighted users, however it will be announced to screen reader users.

components/plugins/lcs.js
```diff
@@ -0,0 +1,839 @@
...
+         		${templates.map((template, index) => html`
+         			<d2l-menu-item data-template-index="${index}" @click="${this._handleMenuItemClick}" @keyup="${this._handleKeyboard}" text="${template.name}" hasChildView="true" class="${template.icon}">
```
dbatiste, line 319: The best event to wire up to here is `d2l-menu-item-select`. It will be dispatched for both mouse and keyboard users.  So there should not be any need to wire up to `click` and `keyup` separately.

Aside from the event wire-up, the `hasChildView` property is meant to be set internally only.  The menu-item looks for nested menus to impart some special behaviour for that case, and this property is used in conjunction with that.  This is solved depending on the chosen list/menu solution in my previous comment.

jschlumkoski: Not necessary anymore because menu was switched to list based on your other recommendation!

components/plugins/lcs.js
jschlumkoski: Not relevant anymore because the menu was switched to a list.

components/toolbar/toolbar-full.js
jschlumkoski: Added missing parameter as indicated.

components/plugins/lcs.js
jschlumkoski: I was having a double scrollbar issue earlier in development. I think this was just remaining, and not needed, so I removed it in my most recent commit.

components/plugins/lcs.js
jschlumkoski: Thank you for noting this! The disclaimer message this was for is something that we were testing but never fully approved, so I've opted to just remove it for the time being in the most recent commit.

components/plugins/lcs.js
jschlumkoski: Removed this style and switched the use of sr-only to use d2l-offscreen instead in the most recent commit.

components/plugins/lcs.js
jschlumkoski: Thanks for the suggestion. This styling wasn't really necessary, so I just opted to remove it in my most recent commit.

components/plugins/lcs.js
jschlumkoski: As we discussed, this will be tackled at a later date!

components/plugins/lcs.js
jschlumkoski: Switched the menu into a button list as per the recommendation. This is in the most recent commit.

components/toolbar/toolbar-full.js
jschlumkoski: Added missing parameter in the most recent commit.

components/plugins/lcs.js
jschlumkoski: Leaving as updated for the time being, as the mentioned callbacks only fire once.

components/plugins/lcs.js
jschlumkoski: Changed to const.

components/plugins/lcs.js
jschlumkoski: Changed to const.

components/plugins/lcs.js
jschlumkoski: Updated based on new context example.

components/lcs-templates.js
```diff
@@ -0,0 +1,266 @@
...
+		src: 'https://lcs-insertelement.s3.us-east-2.amazonaws.com/insertElement/_assets/thirdpartylib/bootstrap-4.3.1/js/bootstrap.min.js'		
+	},
+	{
+		src: 'https://lcs-insertelement.s3.us-east-2.amazonaws.com/insertElement/_assets/js/scripts.min.js'			
+	}
+];
+
+const templates = [
+	{
+		name: 'Accordion',
```
dbatiste, line 32: I know localization was brought up before, but I just wanted to make another comment how I envision this being done. Eventually I think these English values will need to be changed to terms, the actual values added to [en.js](https://github.com/BrightspaceUI/htmleditor/blob/master/lang/en.js), and the plugin code changes to use the localizer and call `this.localize(template.name)`.

components/lcs-templates.js
```diff
@@ -0,0 +1,266 @@
...
+		item: 'Accordion',
+		description: 'Accordions are useful when you need to present learners with a lot of information in a list of items. There may be a lot to learn about each item, but the first priority is for the learner to learn the list as a whole, then allow them to selectively open each section of the accordion to learn more. ',
+		icon: 'accordion.svg',
+		itemLabel: 'List Items and Related Content',
+		multi: true,
+		minimum: 1,
+		maximum: 25,
+		default: 2,
+		preview: {
+         numbered: {
```
dbatiste, line 42: I recommend running the linter locally to fix a bunch of issues (I will catch a lot of stuff for us).  Some nit-picky formatting things like this, but also some legit issues. `npm run lint` will run both eslint and stylelint with our configs.

components/lcs-templates.js
```diff
@@ -0,0 +1,266 @@
+const styleSheets = [
```
dbatiste, line 1: In the original PoC the scripts and stylesheets were specified as dependencies of each widget, with the idea that different widgets could have different dependencies, and we only really want to load the dependencies that are required.  What was the rationale for pulling them out... did you run into some issues?

components/lcs-templates.js
```diff
@@ -0,0 +1,266 @@
...
+		{
+			type: 'create'
+		}]
+	},
+	{
+		name: 'Tabs',
+		item: 'Tab',
+		description: 'Tabs are useful for providing a series of related pieces of content. For example, laying out the information from a scientific study that has a hypothesis, a method, and findings. Tabs are especially effective when information builds from one tab to the next.',
+		icon: 'tabs.svg',
+		multi: true,
```
dbatiste, line 82: I assume `multi`, `minimum`, `maximum`, and `default` represent the number of elements that are inserted... yes?  I might be inclined to make these slightly more descriptive.  `default` is the most ambiguous of them.  Just something that might help future unfamiliar readers.

icons/accordion.svg
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
dbatiste, line 1: Check out these [core guidelines for icons](https://github.com/BrightspaceUI/core/tree/master/components/icons#svg-format). This instruction isn't provided here since we were really expecting others to contribute icons, but this might be something we should add to our readme documentation.

When I run this icon through [svgomg](https://jakearchibald.github.io/svgomg/) it cleans up quite a bit and reduces the size by over 40%.  All the colours in here should probably also be changed to ferrite (`#494c4e`) instead of black.  Ferrite is our black and is used for all of our Daylight icons across the system.

components/plugins/lcs.js
```diff
@@ -0,0 +1,780 @@
...
+			if (!dialog) dialog = component.shadowRoot.appendChild(document.createElement('lcs-templates-dialog'));
+
+			const contextNode = getSelectedLCSElement();
+			if (contextNode) {
+				dialog._editHTML = contextNode.outerHTML;
+			}
+
+			dialog.opened = true;
+
+			dialog.addEventListener('lcs-templates-dialog-close', (e) => {
```
dbatiste, line 58: This is a stylistic thing we sometimes do in our core code to reduce indentation.
```javascript
dialog.addEventListener('lcs-templates-dialog-close', e => {
	if (!e.detail || !e.detail.html) return;
	if (contextNode) {
		editor.selection.select(contextNode);
	}
	editor.insertContent(`${e.detail.html}`);
});
```
Of course, there are other variants as well.

components/plugins/lcs.js
```diff
@@ -0,0 +1,780 @@
...
+				let requiresContentUpdate = false;
+
+				if (styleSheets) {
+					styleSheets.forEach(styleSheet => {
+						if (!editorDocument.head.querySelector(`link[href="${styleSheet.href}"]`)) {
+							requiresContentUpdate = true;
+							const elem = editorDocument.createElement('link');
+							elem.rel = 'stylesheet';
+							elem.href = styleSheet.href;
+							if (styleSheet.integrity) elem.integrity = styleSheet.integrity;
```
dbatiste, line 86: This is something that was originally included in the PoC data but appears excluded above.  Was it intentionally excluded, in which case can we remove this line, or is `integrity` still expected in some cases?

components/plugins/lcs.js
```diff
@@ -0,0 +1,780 @@
...
+		if (!context || !context.activeOrgEnrollment) return;
+
+		const getSelectedLCSElement = () => {
+			const contextNode = (editor.selection ? editor.selection.getNode() : null);
+			if (!contextNode) return null;
+			if (contextNode.tagName !== 'DIV' || !contextNode.classList.contains('lcs-element') || !contextNode.classList.contains('mceNonEditable')) return null;
+			return contextNode;
+		};
+
+		const launchEditor = () => {
```
dbatiste, line 44: I might be inclined to rename this to `launchElementEditor` to avoid any confusion about what editor this is referring to.

components/plugins/lcs.js
```diff
@@ -0,0 +1,780 @@
...
+					});
+				}
+				if (requiresContentUpdate) e.target.setContent(editorDocument.documentElement.outerHTML);
+			}
+		});
+
+		editor.addCommand(cmds.insertElement, () => launchEditor());
+
+		editor.ui.registry.addButton('editlcs', {
+			icon: 'edit-block',
```
dbatiste, line 111: Is this referring to, and using tinyMCE icons?  For everything else we have our own icons, and I suspect we will want to do the same here.  We have some icon generation logic in a CLI that gets run, and part of that also produces a function that adds our icons to tinyMCE's icon registry, so I think we will just need to add our icons to the `/icons` folder.

components/plugins/lcs.js
```diff
@@ -0,0 +1,780 @@
...
+			onAction: function () {
+				const node = editor.selection.getNode();
+
+				if (node.nextElementSibling) {
+					node.parentNode.insertBefore(node.nextElementSibling, node);
+				}
+			}
+		});
+
+		editor.ui.registry.addContextToolbar('modifylcs', {
```
dbatiste, line 151: This is one of the few tinyMCE UI controls we rely on (not just for this plugin).  I suspect the a11y may be an issue using them but I'm not 100% sure.  I wonder what it would take to do our own context toolbar.  I suspect it may be a heavy lift.  @grant-cleary food for thought.

components/plugins/lcs.js
```diff
@@ -0,0 +1,780 @@
...
+			},
+			items: 'moveuplcs|movedownlcs|editlcs|deletelcs',
+			position: 'node',
+			scope: 'editor'
+		});
+	});
+	return name;
+}
+
+class LCSTemplatesEditor extends LitElement {
```
dbatiste, line 163: Nit: I think we should be consistent with our terminology... templates vs elements.  For the class name itself, I would just call it `TemplateEditor` or `ElementEditor`.

components/plugins/lcs.js
```diff
@@ -0,0 +1,780 @@
...
+
+	static get properties() {
+		return {
+			opened: { type: html },
+			_stepIndex: { type: Number },
+			_templateIndex: { type: Number },
+			_properties: { type: Array },
+			_items: { type: Array },
+			_editHTML: { type: String },
```
dbatiste, line 172: It looks like this should be a public property, yes?

components/plugins/lcs.js
```diff
@@ -0,0 +1,780 @@
...
+
+		return html`
+			<d2l-dialog title-text="Insert Element" ?opened="${this.opened}" @d2l-dialog-close="${this._handleClose}" width="1200">
+            ${content}
+				<d2l-button slot="footer" data-dialog-action>Cancel</d2l-button>
+			</d2l-dialog>
+		`;
+	}
+
+	async updated(changedProperties) {
```
dbatiste, line 375: Looks like `async` can be removed from this lifecycle method.

We should set up the observer in one of two ways... either we wire it up once in `firstUpdated()` and disconnect it in `disconnectedCallback` to avoid possible memory leak (with the consequence that it doesn't get re-added if the element is reconnected), or we start observing when the dialog opens and stop observing when the dialog closes.

If I understand correctly, the observer is meant to resize the dialog to better fit the contents of the dialog, as they may change.  So observing the `.editor` content makes sense to me, but I'm not sure why we'd also wire up to observe the `d2l-dialog` also.

components/plugins/lcs.js
```diff
@@ -0,0 +1,780 @@
...
+		const d = this.shadowRoot.querySelector('d2l-dialog');
+		if (d !== null) {
+			this._resizeObserver.observe(d)
+		}
+	}
+
+	_create(step) {
+		const result = {};
+		const template = templates[this._templateIndex];
+		let items = '';
```
dbatiste, line 392: Looks like `items` and `templateHTML` variables are not needed in this method.

components/plugins/lcs.js
```diff
@@ -0,0 +1,780 @@
...
+		this._items.sort((a, b) => {
+			return a.index - b.index;
+		});
+
+		this.requestUpdate('items', []);
+	}
+
+	_buildField(property) {
+		let field;
+		let id = getUniqueId();
```
dbatiste, line 482: Looks like this could be a `const`, yes?

components/plugins/lcs.js
```diff
@@ -0,0 +1,780 @@
...
+
+		switch (property.type) {
+			case 'text':
+				return html`
+					<div class="d2l-htmleditor-templates-field">
+                  <label for="${id}" class="d2l-input-label ${property.required ? 'd2l-input-label-required' : ''} ${property.labelHidden ? 'd2l-offscreen' : ''}">
```
dbatiste, line 488: The `label` property should be specified directly on `d2l-input-text`.  This is because IDREFs cannot cross shadow boundaries, and the actual `input` element is within `d2l-input-text`'s shadowDOM.  Eventually browsers will support delegation of these references, but they don't yet.

components/plugins/lcs.js
```diff
@@ -0,0 +1,780 @@
...
+                  </label>
+                  ${property.tooltip ? html`<d2l-button-icon id="info_${property.key}" icon="tier1:help"></d2l-button-icon><d2l-tooltip position="right" for="info_${property.key}">${property.tooltip}</d2l-tooltip>` : ''}
+                  <d2l-input-text id="${id}" data-property-key="${property.key}" data-type="text" .value="${live(property.value)}" placeholder="${property.placeholder}" ?required="${property.required === true}" @change="${this._change}"></d2l-input-text>
+					</div>
+				`;
+
+			case 'textarea':
+				return html`
+					<div class="d2l-htmleditor-templates-field">
+                  <label for="${id}" class="d2l-input-label ${property.required ? 'd2l-input-label-required' : ''}">
```
dbatiste, line 499: Same thing for this one.

components/plugins/lcs.js
```diff
@@ -0,0 +1,780 @@
...
+		if (itemProperties !== undefined) {
+			for (const [index, itemProperty] of itemProperties.entries()) {
+				this.shadowRoot.querySelectorAll('[data-property-key="' + itemProperty.key + '"]').forEach((e, i) => {
+					this._items[i].properties[index].value = e.value;
+				});
+			}
+		}
+	}
+
+	_handleKeyboard(e) {
```
dbatiste, line 769: Is this referenced anywhere?

components/plugins/lcs.js
grant-cleary: I'm curious why we can't tackle this now. I don't think it would be a particularly difficult lift to wire this up for localization.

components/plugins/lcs.js
```diff
@@ -0,0 +1,780 @@
...
+							editor.selection.select(contextNode);
+						}
+
+						editor.insertContent(`${html}`);
+					}
+				}
+			}, { once: true });
+		};
+
+		editor.on('init change blur', function (e) {
```
grant-cleary, line 73: TinyMCE's `change` event fires a lot. I realize we're not actually doing a ton of work in here because of the nested `if` statements, but I'm curious what the justification is for using these events. Would `SetContent` not be sufficient?

jschlumkoski: @grant-cleary So initially this happened when the dialog was closed, because that's when we would need to add the dependencies if they were missing. The issue I was trying to solve here, was if someone accidentally removed one or all of the dependencies, stuff would stop working with no way to fix it other than to add a new element or edit one. If you have a better way to handle this, I'm all ears!

grant-cleary: So TinyMCE fires a `SetContent` event any time content is injected into the editor via the `mceInsertContent` function. Based on your use case it sounds like this would cover you, since it would fire on init, whenever someone inserted elements via the dialog, and also whenever someone saved in the source code viewer.

components/lcs-templates.js
jschlumkoski: @dbatiste The reason we changed this was because we reasonably expect most people to add more than one element on a page, and then it seemed more beneficial to just load the entire thing to save on request numbers. If we need to re-evaluate this in the future, we can.

components/lcs-templates.js
jschlumkoski: @dbatiste In my most recent commit, I updated the names to "allowMultipleItems", "minimumItems", "maximumItems", and "defaultItems", as they control how many items certain items can have (like accordions). If it still needs to be more descriptive, let me know.

components/plugins/lcs.js
jschlumkoski: @dbatiste Most recent commit updated the code to use this convention.

components/plugins/lcs.js
jschlumkoski: @dbatiste Integrity wasn't being used currently, so I removed it in the most recent commit.

components/plugins/lcs.js
jschlumkoski: @dbatiste Updated naming conventions across the board to make it more consistent in the most recent commit. I incorporated this change.

components/plugins/lcs.js
jschlumkoski: @dbatiste Updated naming conventions across the board to make it more consistent in the most recent commit. I incorporated this change.

components/plugins/lcs.js
jschlumkoski: @dbatiste Updated naming conventions across the board to make it more consistent in the most recent commit. I incorporated this change.

components/plugins/lcs.js
jschlumkoski: @dbatiste Updated to use public property convention in the most recent commit.

components/lcs-templates.js
dbatiste: That sounds great.  😄

components/plugins/lcs.js
jschlumkoski: @dbatiste Correct, I've fixed this in my most recent commit!

components/plugins/lcs.js
jschlumkoski: @dbatiste Removed these in my most recent commit!

components/plugins/lcs.js
jschlumkoski: @dbatiste Label moved to a property on d2l-text-input in most recent commit.

components/plugins/lcs.js
jschlumkoski: @dbatiste Label moved to a property on d2l-input-textarea in most recent commit.

components/plugins/lcs.js
jschlumkoski: @dbatiste It's not used anymore. Removed in my my most recent commit.

components/plugins/lcs.js
jschlumkoski: @grant-cleary Works great, thanks for the suggestion! 👍

components/plugins/lcs.js
grant-cleary: @jschlumkoski, thoughts? I'm wondering what we gain by delaying this change, given that it should be relatively little work.

jschlumkoski: @grant-cleary I think just timing and bandwidth mostly. I think we were also hoping that our internal dev testing might help shape some of the verbiage of the tool, and then it would need changed.

grant-cleary: Ah okay, if the verbiage isn't finalized here, that seems fine then. 👍

jschlumkoski:
@dbatiste The most recent commit adds the change from tooltips to dialogs.

components/d2l-templates.js
```diff
@@ -0,0 +1,267 @@
+import { html } from 'lit-element/lit-element.js';
+const styleSheets = [
+	{
+		href: 'https://lcs-insertelement.s3.us-east-2.amazonaws.com/insertElement/_assets/thirdpartylib/bootstrap-4.3.1/css/bootstrap.min.css'
+	},
+	{
+		href: 'https://lcs-insertelement.s3.us-east-2.amazonaws.com/insertElement/_assets/thirdpartylib/fontawesome-free-5.9.0-web/css/all.min.css'
```
dbatiste, line 7: I _think_ the licensing is ok here but it is worth double checking.  Is it just the fonts that we are relying on?  According to Google Fonts, it looks like Lato is also OFL, and is the font we use in BrightspaceUI.
 https://fontawesome.com/license/free

CC 4.0 is fine, as it MIT, but we currently don't list OFL. Ref: https://github.com/Brightspace/d2l-license-checker/blob/master/config/default_config.json

@cbellini FYI, since you've been looking into licensing for the editor recently.

components/d2l-templates.js
```diff
@@ -0,0 +1,267 @@
...
+	},
+	{
+		href: 'https://lcs-insertelement.s3.us-east-2.amazonaws.com/insertElement/_assets/css/styles.min.css'
```
dbatiste, line 10: I did not review the CSS in this file but to be honest it would not be a bad idea.  It is not specific to this PR, but it would not be a bad idea to have that reviewed if it hasn't already.  Nevertheless, one thing I noticed was the Lato import from Google.  We actually have the Lato font stored on our CDN which is a preferred endpoint from an SLA perspective.

Ref: 
https://s.brightspace.com/lib/fonts/0.5.0/fonts.css
https://github.com/BrightspaceUI/core/blob/master/components/typography/typography.js

components/d2l-templates.js
```diff
@@ -0,0 +1,267 @@
...
+	},
+	{
+		href: 'https://lcs-insertelement.s3.us-east-2.amazonaws.com/insertElement/_assets/css/custom.css'
```
dbatiste, line 13: It looks like this will be used by clients to override some CSS.  Just for my own awareness, what is the process for them to use this - do they engage with LCS to have custom CSS applied?

components/d2l-templates.js
```diff
@@ -0,0 +1,267 @@
...
+	{
+		src: 'https://lcs-insertelement.s3.us-east-2.amazonaws.com/insertElement/_assets/js/scripts.min.js'			
+	}
+];
+
+const templates = [
+	{
+		name: 'Accordion',
+		item: 'Accordion',
+		help: html`<p>Accordions are useful when you need to present learners with a lot of information in a list of items. There may be a lot to learn about each item, but the first priority is for the learner to learn the list as a whole, then allow them to selectively open each section of the accordion to learn more.</p>`,
```
dbatiste, line 35: I know the `html` template being used in here is convenient for constructing the help dialogs.  I think it is worth figuring out the localization sooner rather than later because my original thinking was that these were going to have to be changed to lang-terms, but that obviously won't work with this approach because lang-terms are ICU formatted text, not HTML.  To localize this, it would require us to pass in the `localize` method.  I know the localization was being deferred but I think it has bearing on how this stuff works.

components/d2l-templates.js
```diff
@@ -0,0 +1,267 @@
...
+		icon: 'accordion.svg',
```
dbatiste, line 36: Another option is to bake the `svg` into an `html` template here.  It would save a few HTTP requests.

icons/accordion.svg
```diff
@@ -0,0 +1,19 @@
+<svg width="18" height="18" viewBox="0 0 18 18" xmlns="http://www.w3.org/2000/svg">
```
dbatiste, line 1: Awesome.  These look much better.  We might want to contemplate baking them into the templates module to save requests.

components/plugins/lcs.js
```diff
@@ -0,0 +1,775 @@
...
+import { getUniqueId } from '@brightspace-ui/core/helpers/uniqueId.js';
+
+const name = 'd2l-element-editor-dialog';
+
+export function addElementEditorDialogPlugin() {
+	tinymce.PluginManager.add(name, function (editor) {
+		const context = requestInstance(editor.getElement(), 'context');
+
+		// bail if no LMS context
+		if (!context || !context.activeOrgEnrollment) return;
```
dbatiste, line 33: Does this technically require an org enrolment to use this feature?  In practice, I know this will only be turned on in a couple of placed and controlled via the `insert-element` attribute on the editor, but am wondering if this org check is required.

components/plugins/lcs.js
```diff
@@ -0,0 +1,775 @@
...
+		else {
+			const template = templates[this._templateIndex];
+			const step = template.steps[this._stepIndex];
+			const isCreateNext = (template.steps.length > this._stepIndex + 1 && template.steps[this._stepIndex + 1].type === 'create');
+
+			if (step.type === 'properties') {
+				const properties = (step.getProperties ? step.getProperties() : undefined);
+				const itemProperties = (step.getItemProperties ? step.getItemProperties() : undefined);
+
+				if (properties !== undefined) {
```
dbatiste, line 272: Nit: checking `if (properties)` is a bit more resilient compared to strictly checking `undefined` because it will account for `null` as well.  Not a big deal though.

components/plugins/lcs.js
```diff
@@ -0,0 +1,775 @@
...
+							break;
+						}
+					}
+				}
+
+				templateHTML = this._buildHTML(template, selector);
+
+				const iframeStr = `${stylesStr}<div class="container-fluid"><div class="row"><div class="col-12">${templateHTML}</div></div></div>${scriptsStr}${resizeScript}`;
+
+				content = html`<div class="editor-info"><h2>Preview</h2></div><iframe id="editor-preview" class="editor-preview" srcdoc="${iframeStr}"></iframe>`;
```
dbatiste, line 342: Note: usually we would expect headings to be styled with one of our typography classes.

Ref: https://github.com/BrightspaceUI/core/blob/master/components/typography/styles.js

components/plugins/lcs.js
```diff
@@ -0,0 +1,775 @@
...
+					}
+                  </div>
+                  </d2l-form>
+               </div>
+				`;
+			}
+			else if (step.type === 'preview') {
+				const stylesStr = this._buildPreviewStyles();
+				const scriptsStr = this._buildPreviewScripts();
+				const resizeScript = '<script>function resize() { const h = document.querySelector(\'.container-fluid\').offsetHeight; window.frameElement.setAttribute(\'style\' , \'height: \' + h + \'px;\'); window.frameElement.closest(\'d2l-dialog\').resize(); } const rO =  new ResizeObserver(entries => { resize(); }); rO.observe(document.querySelector(\'.container-fluid\')); resize();</script>';
```
dbatiste, line 324: Is the intention of the resize logic to resize the dialog when the iframe is sized?

components/plugins/lcs.js
```diff
@@ -0,0 +1,775 @@
...
+			buttons.push(html`
+				<d2l-button slot="footer" primary @click="${this._handleNextClick}">
+					${isCreateNext ? 'Insert' : 'Preview'}
+				</d2l-button>
+			`);
+		}
+
+		return html`
+			<d2l-dialog id="editor" title-text="Insert Element" ?opened="${this.opened}" @d2l-dialog-close="${this._handleClose}" width="1200">
+				<div class="editor" tabindex="0">
```
dbatiste, line 364: I think this will end up being a mysterious tab-stop.  I think focus would land here for keyboard users, but nothing would get announced.  For sighted users they might see an outline depending on the browser.  Sometimes it is advantageous to apply focus to heading.

components/d2l-templates.js
cbellini: Thanks @dbatiste!

I'll include this in the existing licensing thread.

components/d2l-templates.js
jschlumkoski: @dbatiste So this is a byproduct of me just including the files from our templates. This is intended to be a local file, so I've removed it in the most recent commit.

components/d2l-templates.js
jschlumkoski: @dbatiste What would be the preferred approach here. Is it better to pass in localize? Should I switch out the html template here in favor of using unsafeHTML in the plugin itself?

components/d2l-templates.js
jschlumkoski: @dbatiste Did this in the most recent commit.

icons/accordion.svg
jschlumkoski: @dbatiste Did this in the most recent commit.

components/plugins/lcs.js
jschlumkoski: @dbatiste I had this because some of the examples I looked at when building it included it. I've removed it in the most recent commit because I don't think it's required.

components/plugins/lcs.js
jschlumkoski: @dbatiste Changed it based on your suggestion.

components/plugins/lcs.js
jschlumkoski: @dbatiste Yes, the iframe is set to resize the dialog when something inside the iFrame changes height (e.g., opening an accordion) so it can change height accordingly.

components/plugins/lcs.js
jschlumkoski: @dbatiste The original reason for the tab index was because when I was using the tooltip button, when you edited an element and this was the first screen that opened, the button was the first thing to get focus, and the tooltip would be open. Switching to dialogs, this isn't an issue anymore. I just forgot to remove it when I pushed that update!

components/d2l-templates.js
grant-cleary: To me it feels like it would be ideal to do the latter, since it's likely easier/more straightforward to localize within the plugin definition vs. this template file.

components/d2l-templates.js
grant-cleary: I took a look at the CSS there. There's a lot going on and it's not super clear to me where we would expect some of it to apply, but nothing stands out to me as problematic given that these are ultimately going to be styles loaded into a Content document body.

components/plugins/lcs.js
jschlumkoski: @dbatiste Pushed a change to add this to the h2s.

components/d2l-templates.js
jschlumkoski: @dbatiste What would be the best way to incorporate this? If I add the CSS file to the style dependencies in the templates file, the version number shouldn't be static right?

grant-cleary: We already have a couple examples of this in the editor here: https://github.com/BrightspaceUI/htmleditor/blob/main/htmleditor-mixin.js#L60

I think the static version is fine.

components/plugins/lcs.js
grant-cleary: Yeah, the `activeOrgEnrollment` business is in many of our custom plugins because they rely on LMS dialogs launching. It shouldn't be necessary in your case. :)

components/d2l-templates.js
dbatiste: That is a good question about the version.  If we were to ever change out Brightspace system font to be something other than Lato, we would release a new version... ex. `https://s.brightspace.com/lib/fonts/0.6.0/fonts.css`, but we still want the user's HTML referencing `0.5.0` if they are using Lato, so that the required font is still downloaded and displayed correctly.  That is why the version is hard-coded in the case Grant linked.

For `styles.min.css` is the expectation that that file is not cached by the browser, so that LCS can apply updates without requiring some cache-busting?

components/d2l-templates.js
dbatiste: How would the code look (localized) when using the `unsafeHTML` directive?

Lang-terms can't contain HTML... so it seems like we'd still need the localize calls baked into the messages, which I suspect will require passing the `localize` method.

```javascript
const templates = [
   {
      ...
      help: `<p>this.localize('insertElement:accordionHelp')</p>`,
      ...
   }
];
```

```javascript
render() {
   return html`
      ...
      ${unsafeHTML(template.help)}
      ...
   `;
}
```

Originally, I was hoping to keep the data (template info) separate from the UI which would have looked like the following, but this does not give you much flexibility to construct rich help.  Ideally, the template data has no dependency on the tech used to construct the UI (in this case Lit, and our localize mixin).

```javascript
const templates = [
   {
      ...
      help: 'insertElement:accordionHelp', // the help lang-term (not containing any HTML)
      ...
   }
];
```

```javascript
render() {
   return html`
      ...
      <p>${this.localize(template.help)}</p>
      ...
   `;
}
```

I'm not really a fan of any of the options at this point.  We could perhaps define the help in the template as an array of lang-terms, each of which would get rendered as a `<p>` by the UI code.  That would keep the template data agnostic to the UI tech, and provide a tiny bit of flexibility.

Any other thoughts/ideas?  Or how would the `unsafeHTML` help us... maybe I am missing something?

components/d2l-templates.js
```diff
@@ -0,0 +1,264 @@
...
+		icon: html`<svg slot="illustration" width="18" height="18" viewBox="0 0 18 18" xmlns="http://www.w3.org/2000/svg"><g fill="none" fill-rule="evenodd" stroke="#494c4e" stroke-width=".5"><g transform="translate(1 1)"><rect x=".25" y=".25" width="15.5" height="4" rx=".5"/><path stroke-linecap="round" d="M2 2.25h6.5"/><path stroke-linecap="round" stroke-linejoin="round" d="M11.904 1.75l.914 1.102.914-1.102"/></g><g transform="translate(1 6.5)"><rect x=".25" y=".25" width="15.5" height="4" rx=".5"/><path stroke-linecap="round" d="M2 2.25h6.5"/><path stroke-linecap="round" stroke-linejoin="round" d="M11.904 1.75l.914 1.102.914-1.102"/></g><g transform="translate(1 12)"><rect x=".25" y=".25" width="15.5" height="4" rx=".5"/><path stroke-linecap="round" d="M2 2.25h6.5"/><path stroke-linecap="round" stroke-linejoin="round" d="M11.904 1.75l.914 1.102.914-1.102"/></g></g></svg>`,
```
dbatiste, line 33: The above conversation has bearing on this too.  If we did decide to try and avoid dependency on `html` here, this would need to become a string, and the consumer would indeed need to use `unsafeHTML` to render it.

components/d2l-templates.js
dbatiste: (so my current thinking is that we are going to need to pass `localize`, but certainly open to all other thoughts)

grant-cleary: Yeah, I wasn't really thinking about the multi-paragraph rich help situation. I was hoping these could be basic strings that we could localize and pass in, but that doesn't really work if we're talking about rich text that could even include other HTML tags. I too think passing in localize is probably the most feasible option, though I'm not keen on it either.

dbatiste (commented):
Hey Justin, I think things are looking pretty good.  At this point I think you'll just need to resolve the merge conflicts and fix up the linting errors and then it will be good to merge... localization to be done later.

components/toolbar/toolbar-full.js
```diff
@@ -136,10 +140,13 @@ class Toolbar extends ToolbarMixin(LitElement) {
 	}
 
 	render() {
-		const hasContext = this.context && this.context.activeOrgEnrollment;
-		const showSourceCodeEditor = this.sourceEditable || (this.context && this.context.sourceEditable);
+		const component = getEditorComponent(this);
```
grant-cleary, line 143: Unfortunately I think your merge resolution also wiped out some necessary changes. I would recommend checking against this version manually and re-aligning with it: https://github.com/BrightspaceUI/htmleditor/blob/main/components/toolbar/toolbar-full.js

Specifically, we can't fetch the editor component this way anymore. It turned out there were some craftily hidden errors that cropped up when we did this. Instead, we now pass down whatever we need (including the context object) to the toolbar itself instead.

components/toolbar/toolbar-full.js
```diff
@@ -459,6 +469,14 @@ class Toolbar extends ToolbarMixin(LitElement) {
 
 }
 
+function allowSourceEditing(component) {
+	return component.sourceEditable || (component._context && component._context.sourceEditable);
+}
+
+function allowInsertElement(component) {
+	return component.insertElement || (component._context && component._context.insertElement);
```
grant-cleary, line 477: Apologies; I didn't notice this before, but if you're intending to check this property on the component, you'll need to actually add it as a property to the `htmleditor` component and set its initial value in the constructor (check `htmleditor.js` for examples of how this is done).

jschlumkoski: @grant-cleary I've added an `insertElement `property in the HTML editor. I assume the value of that will be controlled by the backend feature flag stuff you're going to do?

I'm still unclear what else needs to happen here in order to control the visibility of the button in the toolbar. Previously I had checked for `component.fullPage` and also checked the `insertElement` property on the component, both of which don't seem possible now?

components/toolbar/toolbar-full.js
jschlumkoski: @grant-cleary  How would I go about doing this for fullPage and insertElement now? Having trouble finding an example.

htmleditor.js
```diff
@@ -215,6 +216,7 @@ class HtmlEditor extends HtmlEditorMixin(SkeletonMixin(LitElement)) {
 	constructor() {
 		super();
 		this.disabled = false;
+		this.insertElement = true;
```
grant-cleary, line 219: This should default to false, otherwise it won't be possible to set it false later (boolean attributes are considered `true` if they're present).

htmleditor.js
```diff
@@ -41,6 +41,7 @@ class HtmlEditor extends HtmlEditorMixin(SkeletonMixin(LitElement)) {
 	static get properties() {
 		return {
 			disabled: { type: Boolean },
+			insertElement: { type: Boolean },
```
grant-cleary, line 44:
```suggestion
			insertElement: { type: Boolean, attribute: 'insert-element' },
```
This ensures we can then set the attribute properly  on the element itself. :)

components/toolbar/toolbar-full.js
grant-cleary: Yeah, so you'll want to add it as a property on the toolbar component itself (e.g. https://github.com/BrightspaceUI/htmleditor/blob/main/components/toolbar/toolbar-full.js#L105) and then pass the component property into the toolbar like this: https://github.com/BrightspaceUI/htmleditor/blob/main/htmleditor.js#L466 (and on the render method: https://github.com/BrightspaceUI/htmleditor/blob/main/components/toolbar/toolbar-full.js#L464)

htmleditor.js
jschlumkoski: This should be fixed and pushed!

htmleditor.js
jschlumkoski: This should be added and pushed!

components/toolbar/toolbar-full.js
jschlumkoski: @grant-cleary Still a bit confused by this. Should I be doing this just to handle the fullPage stuff?

grant-cleary: Ah, you'll need to do it both to handle the `fullPage` stuff and to get your new `insertElement` property in here. So you should need to add two additional attributes/properties to the toolbar.

jschlumkoski: @grant-cleary Okay, my most recent commit does these things, and it seems to be working! (Had to set `insertElement` to true in htmleditor.js to test it though)

components/toolbar/toolbar-full.js
```diff
@@ -140,6 +144,7 @@ class Toolbar extends ToolbarMixin(LitElement) {
 	}
 
 	render() {
+		const allowFullPage = this.fullPage || (this.context && this.context.fullPage);
```
grant-cleary, line 147: This isn't needed. `fullPage` isn't attached to the context object, so the latter check will always be false. I would probably just wrap this into `showInsertElement` like this:
```
const showInsertElement = this.fullPage && (this.insertElement || (this.context && this.context.insertElement));
```

htmleditor.js
```diff
@@ -465,7 +465,7 @@ class HtmlEditor extends HtmlEditorMixin(SkeletonMixin(LitElement)) {
 			return html`
 				<div class="d2l-htmleditor-label-flex-container" style="${styleMap(flexContainerStyles)}">
 					${this._renderLabel()}
-					${this._renderEditor(this._toolbar.render(this._context, this.sourceEditable), true)}
+					${this._renderEditor(this._toolbar.render(this._context, this.sourceEditable, this.fullPage, this.insertElement), true)}
```
grant-cleary, line 468: We'll need to be careful here. The inline toolbar is expecting `this.fullscreen` to be passed in as its third argument here. To make matters more complicated, the inline toolbar can end up rendering the full toolbar (when it's in fullscreen mode. So I think we can leave this line alone, but we'll want to modify the following two locations to add these new parameters to the inline toolbar render method and keep the fullscreen parameter last:
https://github.com/BrightspaceUI/htmleditor/blob/main/components/toolbar/toolbar-inline.js#L175
https://github.com/BrightspaceUI/htmleditor/blob/main/htmleditor.js#L516

grant-cleary: So for the first case (toolbar-inline.js), we would get something like:
```
export function render(context, sourceEditable, fullPage, insertElement, fullscreen) {
	if (fullscreen) return html`<d2l-htmleditor-toolbar-full .context="${context}" ?source-editable="${sourceEditable}" ?full-page="${fullPage}" ?insert-element="${insertElement}"></d2l-htmleditor-toolbar-full>`;
	else return html`<d2l-htmleditor-toolbar-inline floating .context="${context}"></d2l-htmleditor-toolbar-inline>`;
}
```
And for the second case (htmleditor.js), we would get something like:
```
_renderInlineEditor() {
	...
	return html`
		...
			${this._toolbar ? this._renderEditor(this._toolbar.render(this._context, this.sourceEditable, this.fullPage, this.insertElement, this._fullscreen), false) : ''}
			...
		...
	...
}
```

components/toolbar/toolbar-full.js
grant-cleary: For the record, I did just completely miss that you did in fact add that property. 🤦

components/toolbar/toolbar-full.js
grant-cleary: Cool beans. Added a couple more comments on the new changes (just some stuff to clean up), but we're for sure almost there!

components/plugins/lcs.js
```diff
@@ -0,0 +1,758 @@
...
+			dialog.addEventListener('d2l-element-editor-dialog-close', e => {
+				if (!e.detail || !e.detail.html) return;
+				if (contextNode) {
+					editor.selection.select(contextNode);
+				}
+				editor.insertContent(e.detail.html);
+			}, { once: true });
+		};
+
+		editor.on('SetContent', function(e) {
```
grant-cleary, line 60: Thing #1 you need to change: This should be on `GetContent`, not `SetContent`. `GetContent` applies to HTML being pulled _out_ of the editor, which, paradoxically, is what you're aiming for here. TinyMCE controls what happens in the editor proper and as you've noted, they don't really allow stuff like this to work when in editing mode. Switching to `GetContent` will mean that on save, preview, etc., the stylesheets will be correctly added to the document.

components/plugins/lcs.js
```diff
@@ -0,0 +1,758 @@
...
+			const editorDocument = new DOMParser().parseFromString(e.target.getContent(), 'text/html');
```
grant-cleary, line 61: Thing 2: If you switch this event to `GetContent`, then `e.target.getContent()` here will create an infinite loop. So you'll want to change this to:
```
const editorDocument = new DOMParser().parseFromString(e.content 'text/html');
```

components/plugins/lcs.js
```diff
@@ -0,0 +1,758 @@
...
+						if (!editorDocument.body.querySelector(`script[src="${script.src}"]`)) {
+							requiresContentUpdate = true;
+							const elem = editorDocument.createElement('script');
+							elem.src = script.src;
+							elem.crossOrigin = 'anonymous';
+							editorDocument.body.appendChild(elem);
+						}
+					});
+				}
+				if (requiresContentUpdate) e.target.setContent(editorDocument.documentElement.outerHTML);
```
grant-cleary, line 89: Thing 3: Not so much broken as could be simplified.
```suggestion
				if (requiresContentUpdate) e.content = editorDocument.documentElement.outerHTML;
```

components/plugins/lcs.js
jschlumkoski: @grant-cleary I've applied all three things. It definitely stops the infinite looping from happening, but the CSS still isn't making its way (or staying) in the head of the document. Is it safe to assume we need to tweak something elsewhere also?

jschlumkoski: @grant-cleary So I did some more digging, and it is making it's way into the head, but it's not applying the CSS in the editor, unless I go to the preview.

grant-cleary: One thing you could try is also adding an additional `SetContent` handler and doing something like `e.target.getDoc().head.appendChild` to apply the styles. No idea if that will work, but the iframe head in the editor is _not_ the same as the head of the actual HTML file.

grant-cleary: Though, thinking further, you shouldn't need to use the event target, since this function already has access to an `editor` object that you could just use directly.

jschlumkoski: @grant-cleary Previously, the way I was adding the CSS it would get added to the head and it would apply to the elements in the editor also.

jschlumkoski: Adding this:
`if (!editor.contentDocument.head.querySelector(`link[href="${styleSheet.href}"]`)) {
   editor.contentDocument.head.appendChild(elem);
}`

Into the piece that adds the CSS if it's not found seems to update the editor to use the CSS. Any issues with this approach?

grant-cleary: Right, I'm fairly certain the way you were doing it before worked because of a bug we had on our end that was causing nested documents to show up when using the `fullPage` plugin. I think resolving that bug likely broke your stuff. If what you have there works, then I'm fine with adding that. I suspect you'll still need the additional `GetContent` handler that re-adds it when the content is fetched out of the editor though. TinyMCE likes to keep its documents separate.

grant-cleary (approved):

PR #281, opened 2021-06-25
Properly set editor height in inline/inline-limited modes

grant-cleary:
The editor's height was not being set properly in inline and inline-limited modes. First off, we weren't respecting the minimum height provided by the consumer. Secondly, we weren't handling percentage heights properly. Thirdly, we didn't have a mechanism for the inline editor to change size to accommodate existing text within the editor.


htmleditor-mixin.js
```diff
@@ -364,7 +376,6 @@ export const HtmlEditorMixin = superclass => class extends Localizer(RtlMixin(Pr
 			link_context_toolbar: true,
 			link_default_protocol: 'https',
 			menubar: false,
-			min_height: this._getMinHeight(),
```
grant-cleary, line 367: This being on all the time caused a problem because TinyMCE applies this to its UI elements only. So TinyMCE could initialize with its iframe height at `minHeight` and the editor would snap to force the container to be `minHeight` as soon as you interacted with the resize handler.

grant-cleary: We do still need it for the inline case, however. When the `autoresize` plugin is used, TinyMCE only pays attention to `min_height`, `max_height`, and the text height when determining its initial size.

htmleditor.js
```diff
@@ -394,6 +399,10 @@ class HtmlEditor extends HtmlEditorMixin(SkeletonMixin(LitElement)) {
 			'd2l-editing': this._editing
 		};
 
+		const containerStyles = {};
+		containerStyles.width = this.width;
+		if (!this._fullscreen && this._percentageBasedHeight()) containerStyles.height = this.height;
```
grant-cleary, line 404: We have to be really careful here because if we're relying on TinyMCE's autosize plugin, we can't give the editor container a fixed height. We need it to be able to size itself appropriately either to the underlying `html-block`'s text height (if the editor is closed) or TinyMCE's height (if it's open).

htmleditor.js
```diff
@@ -402,17 +411,19 @@ class HtmlEditor extends HtmlEditorMixin(SkeletonMixin(LitElement)) {
 
 		const htmlBlockStyles = {};
 		htmlBlockStyles.width = this.width;
+		if (this._percentageBasedHeight()) htmlBlockStyles.height = this.height;
+		else htmlBlockStyles.minHeight = this.height;
```
grant-cleary, line 415: To avoid pop-in whenever the editor is opened, we likely don't want the `html-block` to shrink below the consumer-provided height.

dbatiste (approved):

PR #282, opened 2021-06-25
Allow toolbar to expand to full viewport width if floating

grant-cleary:
When the toolbar is floating (e.g. in inline/inline-limited mode), we don't want it constrained by the width of the editor container. Instead, I believe it's ideal for it to be able to expand to at least the full viewport width.


components/toolbar/toolbar-mixin.js
```diff
@@ -348,7 +351,7 @@ export const ToolbarMixin = superclass => class extends Localizer(RtlMixin(super
 	_updateMeasures() {
 		const items = this._getFlexItems();
 		this._measures = {
-			available: this.offsetWidth,
+			available: this.floating ? document.documentElement.clientWidth : this.offsetWidth,
```
grant-cleary, line 354: So this feels like cheating, since I think if we have this floating toolbar, we don't need/want the resize observer to run at all. Things look really, really bad if the floating toolbar chomps. That said, we do seem to be relying on the observer to handle the initial sizing of the toolbar on load and the sizing of the items within it, so I suspect it's non-trivial to just extract the bits we care about.

demo/htmleditor-inline.html
```diff
@@ -22,6 +22,14 @@ <h2>Inline</h2>
 				</template>
 			</d2l-demo-snippet>
 
+			<h2>Inline (narrow)</h2>
+
+			<d2l-demo-snippet>
+				<template>
+					<d2l-htmleditor type="inline" label="Inline (narrow)" height="250px" width="250px" html="<p>Hello World!</p>" source-editable></d2l-htmleditor>
```
grant-cleary, line 29: This is small enough for the toolbar to overhang the editor.

dbatiste (approved):

PR #283, opened 2021-06-28
DE44206 - Render unsupported browser warning earlier in the render chain (older browsers can't handle the dynamic toolbar imports)

grant-cleary:
This operation was happening a bit too late and these older browsers were dying on the dynamic imports used to load the toolbars. Moving this up the render chain should cause this warning to actually display.


htmleditor.js
```diff
@@ -216,6 +216,7 @@ class HtmlEditor extends HtmlEditorMixin(SkeletonMixin(LitElement)) {
 	}
 
 	render() {
+		if (!isShadowDOMSupported) return this._render();
```
dbatiste, line 219: Should this happen before the `disabled`/`d2l-html-block` scenario?

grant-cleary: Hmm, can `d2l-html-block` render if shadowDOM isn't supported? If it can, then I think it makes sense to swap them. If it can't and the `d2l-html-block` will just be broken, I imagine this order is what we want.

grant-cleary: I can check this in FF60 in our demo page.

grant-cleary: Confirmed that `d2l-html-block` does render fine, so I switched the order of these two cases.

htmleditor-mixin.js
```diff
@@ -444,7 +444,7 @@ export const HtmlEditorMixin = superclass => class extends Localizer(RtlMixin(Pr
 
 		return html`
 			<textarea id="${this._editorId}" class="${classMap(textAreaClasses)}" aria-hidden="true" tabindex="-1">${this._html}</textarea>
-			${!isShadowDOMSupported ? html`<d2l-alert>Web Components are not supported in this browser. Upgrade or switch to a newer browser to use the shiny new HtmlEditor.</d2l-alert>` : ''}
+			${!isShadowDOMSupported ? html`<d2l-alert>${this.localize('noWebComponentSupport')}</d2l-alert>` : ''}
```
klhuff, line 447: :shipit:

klhuff (approved):

htmleditor.js
dbatiste: Yeah, the `d2l-html-block` should work fine with webcomponentsjs polyfill.  The real issue we have is that tinyMCE does not work with the polyfill... which means that tinyMCE doesn't work in shadowDOM when browsers do not natively support it.

dbatiste (approved):

htmleditor.js
grant-cleary: Ah, cool, makes sense then.

PR #287, opened 2021-06-28
Render label when editor is disabled

### LINKED ISSUE #271: Setting disabled property to true removes label
When setting `disabled` property to `true` it is removing the label above the editor. I don't believe this is expected behavior. Was able to work around this issue by adding my own `<label>` tag and adding the `label-hidden` property.

Note that the labels are removed when setting `this.readOnly` to `true`.

readonly-removing-label

grant-cleary:
If the editor is disabled, we still want the label to display (and we want to un-hide it for assistive devices so we don't lose context).

closes #271


dbatiste (approved):
Do we already have a visual-diff that captures this?

grant-cleary:
> Do we already have a visual-diff that captures this?

We do, yeah. The disabled visual-diff fixture has a visible label.

PR #288, opened 2021-06-29
Convert relative URLs to absolute URLs when rendering in the inline/disabled html-block when a host is provided

### LINKED ISSUE #274: Images don't display until the HTML editor is clicked
Environment: Quad; site version 20.21.8.31033. Chrome 91.0.4472.114
Description: 
* The Continuous Professional Development tool is an FRA.  
* In that tool, we have a dialog that uses htmleditor so users can answer questions related to personal development activities.
* After embedding an image and saving, when viewing the user's answer, the image does not display until the user clicks in the editor.  Example: 
image-does-not-display-until editor-in-focus

grant-cleary:
We use relative URLs for all of our system-generated URLs (so when a user uses Insert Image, PowerPaste, etc.). This works fine in the editor because the editor prepends the specified base URL to each relative URL implicitly. This does _not_ work automatically in our html-block when rendering the inline editor. In this case, when a user clicks out of the editor in an FRA, the HTML we render in the html-block control has relative URLs that resolve against the CDN.

Proposed solution: While rendering to these HTML block controls, prepend the host to any `src` or `href` attributes we find on `a`, `img`, and `iframe` elements. We don't want to manipulate the actual `_html` property on the control since we don't want to save these absolute paths. FRA consumers will need to handle that case themselves.

closes #274


htmleditor.js
```diff
@@ -281,6 +281,27 @@ class HtmlEditor extends HtmlEditorMixin(SkeletonMixin(LitElement)) {
 		}
 	}
 
+	_getHtmlWithAbsolutePaths() {
+		if (!this._context) return this._html;
+
+		const host = this._context.host;
+		if (!host) return this._html;
+
+		const tempDoc = new DOMParser().parseFromString(this._html, 'text/html');
```
dbatiste, line 290: This implies that `DOMParser` will exist.  What if tinyMCE doesn't load because the browser is old?

What do FRAs do to handle this currently?

grant-cleary: So FRAs don't all handle this consistently as I understand it, but generally they do have to have code that manipulates paths like this. I suppose we could dynamically use a regex if DOMParser doesn't exist I guess.

grant-cleary: Although DOMParser and its functions seem to be super old: https://developer.mozilla.org/en-US/docs/Web/API/DOMParser#browser_compatibility - are we sure this is an issue?

dbatiste: Oh wow... for some reason I thought that was a tinyMCE object.

dbatiste (approved):

htmleditor.js
grant-cleary: Ah, well it's true that TinyMCE does have a `tinymce.DOMParser` API. I'm uncertain if that just exposes the native DOMParser or if they have their own custom implementation, but I think your thought was for sure reasonable. 😄

PR #289, opened 2021-06-29
Resolve a few bugs with the editor in fullscreen mode

grant-cleary:
There are two bugs I'm looking to fix here (I found them while working on another PR; as far as I know clients haven't reported anything):
1. If the viewport is wide enough that the fullscreen editor doesn't take up all of the available space, clicking out of the main editing area triggers the `focusout` event. This is most noticeable in the inline mode because it causes the editor to collapse into readonly mode (which is quite jarring, to say the least).
2. When moving into (or out of) fullscreen, the toolbar rendering logic croaks because the editor component's shadowDOM hasn't been attached yet, so the call to `getEditorComponent` fails and we can't get the LMS context. Given that we already know the context and this seems to just be a consequence of Lit's rendering order (as opposed to any kind of real bug), we likely want to pass down the properties we need to the toolbars directly instead of trying to fetch the component on render.


htmleditor-mixin.js
```diff
@@ -311,7 +311,7 @@ export const HtmlEditorMixin = superclass => class extends Localizer(FormElement
 
 			// check if focus is within toolbars or tinyMCE dialog
 			const focusedWithin = this.shadowRoot.querySelector(':focus-within');
-			if (focusedWithin || this._dropdownOpen) return;
+			if (focusedWithin || this._dropdownOpen || this._fullscreen) return;
```
grant-cleary, line 314: This is the bit that prevents the editor from collapsing in fullscreen mode when clicking into the rest of the viewport. I don't think this is desirable behaviour, though if this is intended behaviour I can back this part out.

components/toolbar/toolbar-full.js
```diff
@@ -460,12 +462,8 @@ class Toolbar extends ToolbarMixin(LitElement) {
 
 customElements.define(tagName, Toolbar);
 
-function allowSourceEditing(component) {
-	return component.sourceEditable || (component._context && component._context.sourceEditable);
-}
```
grant-cleary, line 465: This is now only used in one place now, so better to inline it.

dbatiste (approved):

PR #291, opened 2021-06-29
Ensure editor still exits editing mode properly if CSS transition is interrupted

grant-cleary:
Depending on the container the inline editor is in, some user actions can end up interrupting the CSS transition we rely on to know when the editor should leave editing mode. Thus, we should also listen to the `transitioncancel` event to make sure we still close the editor in this case.


htmleditor.js
```diff
@@ -345,9 +345,9 @@ class HtmlEditor extends HtmlEditorMixin(SkeletonMixin(LitElement)) {
 		if (reduceMotion) {
 			this._editing = false;
 		} else {
-			this.shadowRoot.querySelector('.d2l-htmleditor-toolbar-container').addEventListener('transitionend', () => {
-				this._editing = false;
-			}, { once: true });
+			const toolbarContainer = this.shadowRoot.querySelector('.d2l-htmleditor-toolbar-container');
+			toolbarContainer.addEventListener('transitioncancel', () => this._editing = false, { once: true });
+			toolbarContainer.addEventListener('transitionend', () => this._editing = false, { once: true });
```
grant-cleary, line 350: Hmm, I think this is technically safe, as although one of these handlers will always be left dangling, subsequent re-renders will just attach identical events that will just be ignored. I think it's still cleaner to have each listener clean up the other event though, since we know only one will ever be hit. I'll do that.

grant-cleary: Oh wait, that doesn't work because there's no way to make the containing events identical. 🤦

grant-cleary: I could do something like this:
```
const toolbarContainer = this.shadowRoot.querySelector('.d2l-htmleditor-toolbar-container');

const exitEditingMode = () => {
	this._editing = false;
	toolbarContainer.removeEventListener('transitioncancel', exitEditingMode, { once: true });
	toolbarContainer.removeEventListener('transitionend', exitEditingMode, { once: true });
}

toolbarContainer.addEventListener('transitioncancel', exitEditingMode, { once: true });
toolbarContainer.addEventListener('transitionend', exitEditingMode, { once: true });
```

dbatiste: I think that is technically better, but what you have isn't horrible.  The dangling event just ends up being a no-op if it ends up being a duplicate.  If you do make the change for your latest suggestion, I think `{ once: true }` becomes useless.

dbatiste (approved):

htmleditor.js
grant-cleary: Ah of course. The event handler will only run once because it disconnects itself afterwards. Ok, I'll do it this way instead just to make it a bit cleaner.

PR #293, opened 2021-06-30
US129229 - Remove undesired textpattern plugin patterns

grant-cleary:
Of the standard set of text patterns, we're looking to remove the following patterns to avoid confusion from users:
```
{start: '#', format: 'h1'}
{start: '* ', cmd: 'InsertUnorderedList'}
```


dbatiste (approved):
This is pretty sad.  It's gonna seem sorta strange to not have `#`.  Maybe we should have a user preference to turn on markdown mode.

grant-cleary:
> This is pretty sad. It's gonna seem sorta strange to not have `#`. Maybe we should have a user preference to turn on markdown mode.

Agreed. I think this makes the feature much less useful and I think people are still going to complain about things like `-` triggering this as well.

davewagler-d2l (approved):

jordankay13 (approved):

PR #294, opened 2021-06-30
US129328 - Remember toolbar chomper state across editors

grant-cleary:
We want the user's chomper state to be persisted across editors (and across the platform). So we'll cache that in the browser's localStorage! This for sure gets weird on shared computers (and shared browser accounts), but that's likely a very narrow use case and we want to avoid having to key based on user data.


dbatiste (approved):
Looks great!

Dan-DeAraujo (approved):

components/toolbar/toolbar-mixin.js
```diff
@@ -17,6 +17,34 @@ const keyCodes = Object.freeze({
...
+		} catch (e) {
+			// Local storage is inaccessible, so cache our basic default instead
+			chomperPreference = true;
+		}
+	}
+
+	return chomperPreference;
+};
+
+const setChomperPreference = (chomped) => {
```
klhuff, line 38: image

klhuff (approved):

components/toolbar/toolbar-mixin.js
dbatiste: Turn that into an emoji.  ;). It has a white background though.  :(

PR #295, opened 2021-07-01
Disconnect resize observers on component disconnect to make sure they don't leak elements

grant-cleary:
Based on [this observation](https://twitter.com/jaffathecake/status/1405437361643790337), we should make sure we're disconnecting our resize observers when the components disconnect.


components/toolbar/toolbar-item-mixin.js
```diff
@@ -33,6 +33,10 @@ export const ToolbarItemMixin = superclass => class extends RtlMixin(superclass)
 		this.activeFocusable = false;
 	}
 
+	disconnectedCallback() {
+		resizeObserver.unobserve(this);
```
grant-cleary, line 37: The rendered toolbar items all share an observer so we can't just disconnect it (or we'll unobserve _every_ toolbar item). Instead, we'll just unobserve the component that's actually disconnecting.

components/toolbar/toolbar-mixin.js
```diff
@@ -155,7 +154,18 @@ export const ToolbarMixin = superclass => class extends Localizer(RtlMixin(super
 			});
 
 		});
-		resizeObserver.observe(container);
+	}
+
+	disconnectedCallback() {
```
dbatiste, line 159: I'm not sure this works correctly.  From what I see, we define the observer in `connectedCallback` (but it's not connected or observing), we start observing in `firstUpdated` (which only runs once), and we disconnect it in `disconnectedCallback` (which may run more than once).  Seems like the state of the observer is not maintained.

components/toolbar/toolbar-item-mixin.js
dbatiste: That makes sense, although as with the other comment, it looks like the state of the observer is not maintained.  This will work if we are sure the item will never be reconnected.  We should be careful however if moving the where the observation begins.

Aside, `disconnectedCallback()` should have a `super.disconnectedCallback()`.

components/toolbar/toolbar-mixin.js
grant-cleary: Hmm, I see, yeah. I can start observing in `connectedCallback` to make sure state is maintained across the render cycle. I think that would solve this problem (provided we didn't move the observer to connect at some other point). Is there a different way you would prefer to handle this (e.g. via actual state tracking)?

dbatiste: No I think the trick is to use `connectedCallback` whenever we use `disconnectedCallback` to wire stuff up.

grant-cleary: Yeah okay, that was my thought. Just wanted to make sure there wasn't a more desirable option.

components/toolbar/toolbar-mixin.js
```diff
@@ -161,34 +161,45 @@ export const ToolbarMixin = superclass => class extends Localizer(RtlMixin(super
...
+		super.disconnectedCallback();
+		if (this._resizeObserver) this._resizeObserver.disconnect();
+	}
+
+	firstUpdated() {
+		super.firstUpdated();
+
+		const container = this.shadowRoot.querySelector('.d2l-htmleditor-toolbar-container');
 
 		requestAnimationFrame(() => {
```
grant-cleary, line 204: Hmm, actually, now I'm wondering if this also needs to move earlier. It wasn't an issue in my testing, but I wonder if this can ever run too late now that the observer is attached earlier.

grant-cleary: Yeah... As this change stands now, the responsive behaviour does seem to work just fine, but I'm less sure about edge cases. I'm fine if we want to shift back to wiring it up in `firstUpdated`; I agree that the more stuff we shift out of order, the sketchier things get.

components/toolbar/toolbar-mixin.js
```diff
@@ -166,29 +166,37 @@ export const ToolbarMixin = superclass => class extends Localizer(RtlMixin(super
...
+		super.disconnectedCallback();
+		if (this._resizeObserver) this._resizeObserver.disconnect();
+	}
+
 	firstUpdated() {
 		super.firstUpdated();
 
-		const container = this.shadowRoot.querySelector('.d2l-htmleditor-toolbar-container');
-		const resizeObserver = new ResizeObserver(async entries => {
+		if (!this._resizeObserver) {
```
dbatiste, line 177: I think this will always be `undefined` when this runs.

components/toolbar/toolbar-item-mixin.js
```diff
@@ -33,9 +33,14 @@ export const ToolbarItemMixin = superclass => class extends RtlMixin(superclass)
 		this.activeFocusable = false;
 	}
 
-	firstUpdated() {
-		super.firstUpdated();
+	connectedCallback() {
```
dbatiste, line 36: We're you going to change this one back as well?

grant-cleary: Mm, did we think it was necessary to change this one? My understanding is that the broader toolbar resizing depends on the individual items resizing but not vice versa (in which case being a bit later in this case won't matter). Is that understanding accurate?

dbatiste: I think it's ok.  I was just thinking along the lines of minimizing the changes.

dbatiste (approved):

components/toolbar/toolbar-item-mixin.js
grant-cleary: Ah okay, fair. I can switch it back. Thinking on it, since we're blowing up the toolbar resizing when the element gets disconnected/reconnected, it doesn't buy us anything to have this one work optimally. 🤷

components/toolbar/toolbar-mixin.js
grant-cleary: Ah, you're right. Just copy/pasted it back from when it was in `connectedCallback` where this could be defined already. I'll remove it.

components/toolbar/toolbar-mixin.js
grant-cleary: Eh, so I didn't notice that the order of operations is changed a bit here. Basically we're just fetching the container after creating the resizer instead of before. 🤷

PR #296, opened 2021-07-01
Share one lazily loaded ifrau client across all editor instances

grant-cleary:
Ideally, we would only connect one ifrauclient per document. However, we can get a bit closer to that ideal state by lazily loading/connecting a single client shared by all editor instances. It won't prevent other apps or components from creating clients, but it will at least allow the editor to share one instead of creating one per ifrau action.


grant-cleary:
Note: I still need to test this in the hello-world FRA. I just wanted to get the PR up first in case there are strong feelings about this implementation.

components/plugins/quicklink.js
```diff
@@ -104,87 +105,12 @@ class QuicklinkDialog extends RequesterMixin(LitElement) {
...
-						createResult.AddListener(quicklinks => resolve(quicklinks));
-
-					});
-
-				});
-
-			}
-
-			const result = await resultPromise;
+			const result = await openQuicklinksDialog(
```
grant-cleary, line 108: Whoops, this name isn't accurate.

components/lms-adapter.js
```diff
@@ -271,3 +259,111 @@ export async function getTempFile(src, fileId) {
 
 	}
 }
+
```
dbatiste, line 262: It's sort of unfortunate to have this (mentions and quicklinks) domain specific stuff creeping into this file.

grant-cleary: Yeah, we for sure could. I moved them because I felt centralizing them made more sense at this point, but I don't feel strongly about it and I can certainly move them back and just export/use `getIfrauClient()`.

dbatiste: I don't feel super strongly about it, but the decision of where to place such code begins to get really blurry.  I think isolating the domain specific code has some advantages.

grant-cleary: Yeah, that's fine. I can just move it back.

dbatiste (approved):

htmleditor-mixin.js
```diff
@@ -416,7 +416,7 @@ export const HtmlEditorMixin = superclass => class extends Localizer(FormElement
 			link_default_protocol: 'https',
 			menubar: false,
 			mentions_fetch: (query, success) => {
-				setTimeout(async() => success(await queryMentions(this, query.term)), 0);
+				setTimeout(async() => success(await queryMentions(this._context.orgUnitId, query.term)), 0);
```
grant-cleary, line 419: Ah crap, this is wrong now. I forgot to correct this when swapping back to the plugin file.

PR #297, opened 2021-07-01
Validation tooltip should be positioned relative to editor container

grant-cleary:
We want the validation error tooltip to be positioned relative to the editor container, which wasn't actually a given based on the current rendering logic. The tooltip renders relative to its target's `offsetParent`, but because the target was the editor container and the editor container had no positioned ancestors within the editor component, we were implicitly reliant on the consumer to provide a positioned parent for the editor so the tooltip could render in the right place.


htmleditor.js
```diff
@@ -77,6 +77,7 @@ class HtmlEditor extends HtmlEditorMixin(SkeletonMixin(LitElement)) {
 			.d2l-htmleditor-label-flex-container {
 				display: flex;
 				flex-direction: column;
+				position: relative;
```
grant-cleary, line 80: Making this container positioned forces it to be the `offsetParent` for the validation tooltip, guaranteeing that it will always render below the editor (instead of being reliant on consumers of the component).

grant-cleary: The equivalent inline container is already positioned, so no change is necessary there.

README.md
```diff
@@ -55,7 +55,7 @@ Types of editors (toolbar features):
 | `isDirty` | Boolean | Whether or not the editor is [dirty](https://www.tiny.cloud/docs/api/tinymce/tinymce.editor/#isdirty). If `auto-save` is enabled, set `isDirty` to false when triggering workflows where you do not want the autosave prompt. |
 | `label` | String, required | Label for the editor. |
 | `label-hidden` | Boolean | Hides the label visually. |
-| `max-length` | Number, default: null | Maximum length of HTML contents, including HTML control characters. Enables validation on this criterion when set. |
+| `max-length` | Number, default: null | Maximum length of HTML contents, including HTML control characters. Enables validation on this criterion when set. A `null` value is equivalent to no limit. |
```
grant-cleary, line 58: Sorry @dbatiste, you requested a clarification like this on the initial PR but I completely forgot to implement it. 🤦

htmleditor.js
dbatiste: I might be included to add a comment next to this, since it might not be obvious why this is specified.

```css
position: relative; /* required for validation tooltip positioning */
```

Hopefully a visual-diff would catch an error if someone erroneously removed this though.

I would probably make sure this works ok in all the main contexts we expect - standard page, both the primary & secondary panels in the page template, and dialogs.  You might have already checked those.

dbatiste (approved):
Just a couple notes to consider before merging.

htmleditor.js
grant-cleary: Good thought on the comment; I'll add that. I did indeed test this in all the places; it works fine everywhere except in the dialog. The dialog case is an interesting one... The problem is that the tooltip has no room to expand, since the dialog container bounds it. It tries to reposition itself to the side of the editor instead and gets cut off because while there's more room to render the tooltip there, there still isn't enough room. I'm not super sure there's much we can do there, since all of that logic is internal to the tooltip and none of its APIs seem to help us here.

PR #299, opened 2021-07-06
Tweak editor height and margin properties to make autoresizing more sensible in the inline editor

grant-cleary:
There were some minor discrepancies with how the inline editor behaved on resize:
1. TinyMCE seemed to want to expand more than it should have given the `min_height` parameter we gave it.
2. There was some minor pop-in because the height of the editor proper didn't account for the 2px wide border on the container.


htmleditor-mixin.js
```diff
@@ -356,7 +357,10 @@ export const HtmlEditorMixin = superclass => class extends Localizer(FormElement
 		}
 
 		const heightConfig = {};
-		if (applyEditorHeight && !this._percentageBasedHeight()) heightConfig.min_height = this._getUnitlessHeight(this.height);
+		if (applyEditorHeight && !this._percentageBasedHeight()) {
+			heightConfig.autoresize_bottom_margin = 0;
```
grant-cleary, line 361: So it doesn't seem to be documented anywhere, but this seems to get set to a default of `50`, which is why we get that whopping giant block of bottom padding inside the editor iframe in inline mode.

htmleditor-mixin.js
```diff
@@ -14,6 +14,7 @@ import { RtlMixin } from '@brightspace-ui/core/mixins/rtl-mixin.js';
 import { tinymceLangs } from './generated/langs.js';
 
 const defaultMinHeight = 300;
+const inlineHeightOffset = 4; // Inline editor has a 2px width border
```
grant-cleary, line 17: Not sure if there's a better way to do this, but since TinyMCE is managing the inline editor height, we need to give it a slightly smaller height to account for the border.

dbatiste: I think it's fine, although if it were me I probably wouldn't declare a `const` just for this, since it's only used once.  I'd probably just specify the `4` inline where it is used, and move the comment down to be with it.

htmleditor-mixin.js
dbatiste: Interesting.

dbatiste (approved):

htmleditor-mixin.js
grant-cleary: Yeah, okay, I'll switch that for posterity.

PR #300, opened 2021-07-07
Prevent editor from creating a stacking context and use CSS to force TinyMCE's dialogs to behave instead

grant-cleary:
Applying a `transform` on the editor when in an FRA context has the side effect of creating a new stacking context for each editor. This causes some unfortunate behaviour when editors are placed next to each other in the DOM such that the menus/dialogs of one editor can overlap with the others; in this case, the later editors take precedence over the previous ones for stacking and the previous editors' menus end up hiding underneath them. We can't really fix this internally because anything we do in the editor proper is still going to be within that top-level stacking context on the component.

So here's the proposed fix: Instead of applying the transform, use sketchy-ish CSS to force TinyMCE's dialogs to behave properly when opening within an iframed FRA. The CSS in this PR forces the dialogs to open over the center of the editor and also has the advantage of allowing TinyMCE's dialog shims to still cover the entire app frame.

I did test this in the Hello World app with editors spread throughout the iframe and they all worked as I would expect when opening TinyMCE dialogs, so... 🤞


htmleditor-mixin.js
```diff
@@ -126,15 +126,26 @@ export const HtmlEditorMixin = superclass => class extends Localizer(FormElement
...
-			}
 			:host([fra-context][_fullscreen]) .tox-dialog-wrap {
 				bottom: auto;
 			}
 			:host([fra-context]) .tox-dialog {
 				max-height: 100vh;
 			}
+			/* The below styles are css magic to work around tinymce dialog fixed/flex positioning in large frames */
+			:host([fra-context]:not([_fullscreen])) .tox-dialog-wrap {
+				height: 100vh;
```
grant-cleary, line 137: This allows the wrap (and thus the shim) to fill the entire iframe.

dbatiste (approved):
This seems ok.  The path I was thinking of was to only apply 
```css
:host([fra-context]:not([_fullscreen])) {
	transform: scale(1); /* css magic to work around tinymce dialog fixed/flex positioning in large frames */
}
```
when the editor has the focus.  That way, editors below (which the user is not interacting with) don't get this, and thus would not stack on top of previous editors.

grant-cleary:
> This seems ok. The path I was thinking of was to only apply
> 
> ```css
> :host([fra-context]:not([_fullscreen])) {
> 	transform: scale(1); /* css magic to work around tinymce dialog fixed/flex positioning in large frames */
> }
> ```
> 
> when the editor has the focus. That way, editors below (which the user is not interacting with) don't get this, and thus would not stack on top of previous editors.

Oh I see. I guess the one downside is that we would maybe still want to apply some CSS to the shim so that it actually shims the iframe (otherwise it only shims the editor element). Still, I think your way + a bit of extra CSS on the shim probably does solve this problem with less CSS overriding...

grant-cleary:
> Oh I see. I guess the one downside is that we would maybe still want to apply some CSS to the shim so that it actually shims the iframe (otherwise it only shims the editor element). Still, I think your way + a bit of extra CSS on the shim probably does solve this problem with less CSS overriding...

I gave this a shot and I'm unsure how I feel about it. The problem is that if you click on the page outside of the iframe the editor loses focus and the dialog fires its way back up to the middle of the iframe. I could fix that by reflecting the `_focused` property and selecting `:host([fra-context][_focused]:not([_fullscreen]))` instead, but I'm not sure whether we want to change attributes on the component every time it gains/loses focus.

dbatiste:
> but I'm not sure whether we want to change attributes on the component every time it gains/loses focus

We do this sometimes in other components.  Usually the preference to to not sprout private properties/attributes on the host, but rather contain that within the shadowDOM in the form of attributes or classes on those elements.  I'm good either way.

grant-cleary:
> We do this sometimes in other components. Usually the preference to to not sprout private properties/attributes on the host, but rather contain that within the shadowDOM in the form of attributes or classes on those elements. I'm good either way.

Right, so if we applied the transform to the child container that would let us do that without exposing an attribute on the component itself. That seems fine; I was more worried about the potential cost of doing a re-render every time the editor gains or loses focus.

dbatiste:
> I was more worried about the potential cost of doing a re-render every time the editor gains or loses focus.

It is a valid concern but I would not expect those events to fire that often, and would avoid targeting tinyMCE's elements in CSS.  There is risk in creating this magical stacking context that can potentially burn us, as we've seen - your sol'n is probably safer in that regard but subject to break when tinyMCE changes stuff, which hopefully we'd catch in testing.  They both have trade-offs.

grant-cleary:
> It is a valid concern but I would not expect those events to fire that often, and would avoid targeting tinyMCE's elements in CSS. There is risk in creating this magical stacking context that can potentially burn us, as we've seen - your sol'n is probably safer in that regard but subject to break when tinyMCE changes stuff, which hopefully we'd catch in testing. They both have trade-offs.

That's fair. In this case, I suspect your suggestion is more stable in the long run. I think it makes sense to try your suggestion first and if there are cases where that blows up, we can switch to the more fragile way I've done things here.

htmleditor-mixin.js
```diff
@@ -135,6 +136,13 @@ export const HtmlEditorMixin = superclass => class extends Localizer(FormElement
 			:host([fra-context]) .tox-dialog {
 				max-height: 100vh;
 			}
+			/* The below styles are css magic to work around tinymce dialog fixed/flex positioning in large frames */
+			:host([fra-context]:not([_fullscreen])) .tox-dialog-wrap__backdrop {
+				height: 100vh;
+				left: auto;
+				top: auto;
+				width: 100%;
+			}
```
grant-cleary, line 145: I still think this has value. Without this, TinyMCE's dialog shim only applies over the editor itself and the user can interact with elements outside of the editor.

PR #301, opened 2021-07-07
Now that ISF is included in the inline-editor, we need the replace-strings plugin as well to format images and Kaltura links

grant-cleary:
We need this plugin to properly format some ISF contents (e.g. media iframes). I forgot to add this when ISF was added to the inline-limited editor.


components/toolbar/toolbar-inline-limited.js
```diff
@@ -45,6 +46,7 @@ export function getPlugins(component) {
 		component.fullPage ? addFullPagePlugin() : '',
 		addISFPlugin(),
 		addQuicklinkPlugin(),
+		addReplaceStringsPlugin(),
```
dbatiste, line 49: Maybe `addISFPlugin()` should be doing this... since it sounds like it is in fact a dependency of ISF.  I guess the trade-off is that replace-strings plugin could get added to the list of plugins more than once - I think tinyMCE ignores duplicates though, and if it didn't, we could remove them, albeit at the cost of some array processing.

grant-cleary: Hmm, yeah. I think whether or not that makes sense depends on whether we eventually expand this out to do more replace strings. I'm unsure whether we would always expect ISF to include those. 🤔

dbatiste: What is it about the replacestring plugin that this needs?

grant-cleary: So right now, all of it. The encoding and decoding of `{orgUnitId}` replace strings. However, I know we originally wanted this plugin to do much more than just that. That expanded functionality would be irrelevant to ISF.

dbatiste: So are there parts that should be factored out?  Right now, one way or another, we are treating it as a dependency but not importing in such a way that the dependency is removed if ISF is removed.

grant-cleary: So originally in dev I did include this as a subset of the ISF plugin, but then we made the deliberate decision to move it out into its own plugin. I'll check back to those older PRs to see if I can figure out why we did that and whether our decision makes sense anymore.

grant-cleary: Oh okay, I remember now. Originally when I included this logic, it applied directly to the output ISF spat out. That doesn't work because we need to decode and re-encode these whenever content is set, not just on ISF output. I think we decided to include it as its own plugin because we figured this was generically useful outside of ISF; like we could expand this to apply to replace strings in general.

grant-cleary: So I think what makes sense is maybe that your original idea makes sense. The stuff that's currently there should be split out and used generically both by the replace-strings plugin _and_ by ISF. I think we could split out the encode/decode functions as exports that other plugins could then import and use as needed, though we would probably want to ensure then that we didn't try to double encode or decode content.

grant-cleary:
Ok, for now I just decided to have ISF load the replace-strings plugin directly because the only logic that plugin has is needed by ISF. I added a comment signifying that we should revisit if we decide to build out that plugin at some point.

dbatiste (approved):

PR #304, opened 2021-07-12
Track state when a dialog is currently open to ensure we don't allow focus to leave the editor

grant-cleary:
When in an FRA, we can't rely on checking `document.activeElement` to determine if a dialog is currently open, since LMS dialogs are created as children of the parent page, not the FRA iframe. As such, this PR switches the editor to save dialog open/close states itself directly.


components/plugins/attributes.js
```diff
@@ -37,6 +37,7 @@ export function addAttributesPlugin() {
 			if (!dialog) dialog = component.shadowRoot.appendChild(document.createElement('d2l-htmleditor-attributes-dialog'));
 
 			dialog.opened = true;
+			component._dialogOpen = true;
```
grant-cleary, line 40: Side note: There's a bunch of stuff that's now common to all of these dialogs. I'm considering creating a helper to consolidate all of this common logic (not in this PR though... Too risky this close to release).

htmleditor-mixin.js
```diff
@@ -311,7 +311,7 @@ export const HtmlEditorMixin = superclass => class extends Localizer(FormElement
 
 			// check if focus is within toolbars or tinyMCE dialog
 			const focusedWithin = this.shadowRoot.querySelector(':focus-within');
-			if (focusedWithin || this._dropdownOpen || this._fullscreen) return;
+			if (focusedWithin || this._dialogOpen || this._dropdownOpen || this._fullscreen) return;
```
grant-cleary, line 314: This leaves the check on line 321 obsolete (technically so is line 322 for other reasons), but I also don't see specific harm in leaving those in place. I can remove them if we want though.

components/plugins/attributes.js
grant-cleary: Other side note: I'm totally not consistent in the order in which I apply these. It doesn't matter, but it bothers me, so I'll push a small change tomorrow to correct the formatting.

htmleditor-mixin.js
dbatiste: Are you sure it makes 322 obsolete?

While I am nervous about removing them, I also know that in the future we may be looking at this, afraid to remove them with even less confidence for lack of familiarity as time passes.  😬

dbatiste (approved):

htmleditor-mixin.js
grant-cleary: This change doesn't make 322 obsolete. I think what makes 322 obsolete is the changes made to the TinyMCE dialog handling to force the shim to fill the iframe. Though... I guess someone could still click out of the FRA iframe itself, so maybe not lol.

grant-cleary: I removed 321 because it's actually obsolete and added a comment indicating why 322 is not. :)

PR #306, opened 2021-07-12
Make sure we're fetching toolbar element selector at run-time instead of on init (so closing menus works properly)

grant-cleary:
Because we initialize TinyMCE only once, we were incidentally locking in the toolbar type too early. Instead, we need to make sure we're fetching this when the editor's `click` event fires, not when it's registered.


components/toolbar/toolbar-inline.js
```diff
@@ -2,22 +2,22 @@ import 'tinymce/tinymce.js';
...
+export function getTagName(fullscreen) {
+	return fullscreen ? getFullTagName() : tagName;
 }
 
 export { getExternalPlugins };
 
 export function getPlugins(component) {
 	return [
 		...getFullPlugins(component),
-		!component._fullscreen && !component._percentageBasedHeight() ? 'autoresize' : ''
```
grant-cleary, line 20: This first check wasn't doing anything because of course the editor never loads in fullscreen mode and once we init TinyMCE we're stuck with this plugin for good. We're going to have to deal with this use case being crappy for now. :(

dbatiste (approved):
From what I gather, the issue is that without this change, it will be closing dropdowns for the wrong toolbar if the user is in fullscreen mode.  This changes ensures that we select the fullscreen toolbar for closing dropdowns when in fullscreen mode.

grant-cleary:
> From what I gather, the issue is that without this change, it will be closing dropdowns for the wrong toolbar if the user is in fullscreen mode. This change ensures that we select the fullscreen toolbar for closing dropdowns when in fullscreen mode.

That's correct. Closing menus for the wrong toolbar ends up being a no-op from the user's perspective, though it's possible we do end up re-rendering them out of sight. I didn't verify that bit, since this is (imo) the right solution regardless.

PR #307, opened 2021-07-13
Avoid using transform at all for TinyMCE dialogs in FRAs

grant-cleary:
Unfortunately the previous solution (applying the transform only when focused) still didn't solve the problem. While the editor doesn't get overlapped by other editors, it can be overlapped by other page elements because it still creates its own stacking context in this case.

Solution: Switch to the more fragile CSS scheme.


htmleditor-mixin.js
```diff
@@ -463,6 +469,30 @@ export const HtmlEditorMixin = superclass => class extends Localizer(FormElement
 					this._handleFocusIn();
 				});
 
+				if (this._fraContext) {
```
grant-cleary, line 472: If we're okay with how hacky this is I'll add a comment explaining why this exists.

dbatiste: This is something we'll definitely want to test whenever we do tinyMCE updates.

dbatiste (approved):

htmleditor-mixin.js
grant-cleary: Yeah, this is going to be super fragile.

PR #308, opened 2021-07-14
Ensure we do replace string processing in the inline editor's HTML block

grant-cleary:
The HTML block within the inline editor needs to be have replace strings processed to ensure iframes/images load correctly when the editor is moved into readonly mode. As such, I think it makes sense to shift this replace-strings logic into the editor's core setup. If/when we expand out the replace-strings plugin, that plugin can also then either be kept separate from this logic or merged together, depending on what we decide at the time.

Note that I also refactored the actual replacing code to use the DOMParser to parse the entire block of HTML and pick out the img/iframe tags instead of using regex to do that bit. This simplifies things a fair bit because the regex was relatively complex and the DOMParser can just handle that complexity for us.


components/plugins/replace-strings.js
```diff
@@ -1,130 +1,44 @@
...
-	{ text: 'OrgUnitId', value: '{OrgUnitId}' },
-	{ text: 'OrgUnitName', value: '{OrgUnitName}' },
-	{ text: 'OrgUnitPath', value: '{OrgUnitPath}' },
-	{ text: 'OrgUnitTypeId', value: '{OrgUnitTypeId}' },
-	{ text: 'RoleCode', value: '{RoleCode}' },
-	{ text: 'RoleId', value: '{RoleId}' },
-	{ text: 'RoleName', value: '{RoleName}' },
-	{ text: 'UserId', value: '{UserId}' },
-	{ text: 'UserName', value: '{UserName}' }
-];*/
```
grant-cleary, line 22: I'm deleting all of this for now. Since the file is still going to remain, we'll have GitHub's history to refer back to if we want to see what this prototype looked like. Now that we're not using this as a plugin at all, it doesn't make much sense to keep all of this around.

components/plugins/replace-strings.js
```diff
@@ -1,130 +1,44 @@
...
+	if (mceSrc !== null && mceSrc !== undefined) elm.setAttribute('mce_src', src);
+}
 
-			html = html.replace(/<(img|iframe)[^>]*\sdata-d2l-src[\s]*=[\s]*(\'([^\']*)\'|\"([^\"]*)\"|([^>^\s^\/]*))[^>]*>/gi, decode); // eslint-disable-line no-useless-escape
+export function encodeSrcReplaceStrings(html, orgUnitId) {
+	const tempDoc = new DOMParser().parseFromString(html, 'text/html');
+	if (!tempDoc.body) return html;
 
-			// convert %7Bstuff%7D to {stuff} -> Mozilla based browsers
-			html = html.replace(/(%7B(orgId|orgName|orgUnitId|OrgUnitName|OrgUnitCode|OrgUnitPath|UserId|UserName|OrgDefinedId|FirstName|LastName|ExternalEmail|InternalEmail|RoleId)%7D)/gi, '{$2}');
```
grant-cleary, line 89: I imagine this must have been for some old version, as I have no issue with this encoding in modern Firefox.

htmleditor.js
```diff
@@ -295,8 +294,12 @@ class HtmlEditor extends HtmlEditorMixin(SkeletonMixin(LitElement)) {
 			elm.setAttribute(pathAttrName, `${host}${path}`);
 		};
 
-		tempDoc.body.querySelectorAll('img, iframe').forEach(elm => prependHost(elm, 'src'));
-		tempDoc.body.querySelectorAll('a').forEach(elm => prependHost(elm, 'href'));
+		tempDoc.body.querySelectorAll('img, iframe').forEach(elm => {
+			encodeElementSrcReplaceStrings(elm, this._context.orgUnitId);
```
grant-cleary, line 298: We already need to use the DOMParser here to prepend the host to the image/iframe src (if needed), so exporting this element-by-element function lets us avoid parsing the HTML twice.

dbatiste (approved):
I think it would be useful to have some unit tests for a couple of these exported methods.

grant-cleary:
> I think it would be useful to have some unit tests for a couple of these exported methods.

Good idea. I was planning to look at writing some unit tests for the editor in a general sense soon, so I can look at this too.

PR #310, opened 2021-07-15
Correct typo in component fetching

grant-cleary:
Does what it says on the tin!


ronan-f (approved):
Thanks for that! 🙏

grant-cleary:
> Thanks for that! 🙏

No problem! I broke it so I should probably fix it. 😆

PR #311, opened 2021-07-15
Switch fonts to px instead of pts and identify our Brightspace default

grant-cleary:
We want to switch the editor's font-size menu from `pt` to `px`, as well as add our system default font.


components/toolbar/toolbar-full.js
```diff
@@ -445,6 +440,21 @@ class Toolbar extends ToolbarMixin(LitElement) {
 		`);
 	}
 
+	_getFontSizeItems() {
+		const defaultSize = (0.95 * parseInt(this.rootFontSize.replace('px', ''))) || 19; // Default font-size of 0.95 rem (.d2l-typography)
```
grant-cleary, line 444: I'm unsure about this. On the one hand, our default font size is definitely relative to the root font size, but I worry about the potential of this getting messy if the root font size is something that doesn't resolve nicely.

dbatiste: It would be better to avoid hard-coding the root font-size (`20px`) in this file, as well as the `0.95` multiplier. 
 Although getComputedStyle is kinda expensive, I would be inclined to call it once in this module to get the size.  
```javascript
window.getComputedStyle(document.body).fontSize
```

grant-cleary: Hmm, okay. I had considered doing that but wasn't sure if we wanted to avoid the expensiveness of it. If you're not concerned overly much at the cost of that, I can easily do it that way.

components/toolbar/toolbar-full.js
```diff
@@ -445,6 +440,21 @@ class Toolbar extends ToolbarMixin(LitElement) {
...
+
+		const sizes = [ defaultSize, 10, 12, 14, 16, 20, 24, 32, 48 ].sort();
```
dbatiste, line 446: Is it possible to make this stateless (not depend on `this`) by moving the localization up into the template, and then be able to move this code out of the class so that it's only invoked once per page, as opposed to once per `render()` of the toolbar?

grant-cleary: Yeah, I'm sure I can make something work there.

components/toolbar/toolbar-full.js
dbatiste: Yeah I think it is better than hard-coding these details.

grant-cleary: Oh, whoops. I realize one half of this now. I didn't intend to hard-code the `20px` root font size. I exposed it as an attribute on the toolbar only to forget to actually pass it in. 🤦 

Irrelevant because I incorporated both of your suggestions now; just kind of funny that it slipped my mind and just happened to work because I chose our "usual" root font size as the default for that property.

components/toolbar/toolbar-full.js
```diff
@@ -44,6 +44,14 @@ import { cmds } from '../commands.js';
 import { hasLmsContext } from '../lms-adapter.js';
 import { ToolbarMixin } from './toolbar-mixin.js';
 
+const defaultFontSize = parseInt(
+	window.getComputedStyle(document.body, null).getPropertyValue('font-size').replace('px', '')
+);
+
+const fontSizeOptions = [ defaultFontSize, 10, 12, 14, 16, 20, 24, 32, 48 ].sort().map((size) => {
+	return { default: size === defaultFontSize,	value: `${size}px` };
```
grant-cleary, line 52: Actually, I can simplify this further to remove an iteration of this array.

components/toolbar/toolbar-full.js
```diff
@@ -403,16 +404,10 @@ class Toolbar extends ToolbarMixin(LitElement) {
 				</d2l-dropdown-menu>
 			</d2l-htmleditor-button-menu>
 			<d2l-htmleditor-separator ?responsive-hidden="${!this._showMainFontMenus}"></d2l-htmleditor-separator>
-			<d2l-htmleditor-button-menu cmd="${cmds.fontSize}" ?responsive-hidden="${!this._showMainFontMenus}" text="${this.localize('font:size')}" style="width: 80px;">
+			<d2l-htmleditor-button-menu cmd="${cmds.fontSize}" ?responsive-hidden="${!this._showMainFontMenus}" text="${this.localize('font:size')}" style="width: 85px;">
```
dbatiste, line 407: Does this width change affect the breakpoints at all?

grant-cleary: It appeared not when I tested it. I'll test again for double/triple certainty. It would only be the last breakpoint (collapsing the font menus) that would be impacted.

grant-cleary: Yup, I can confirm that we collapse the font menus into each other a few px before this would create a third toolbar row.

dbatiste (approved):

PR #312, opened 2021-07-16
Ensure we're passing existing attributes to the dialog when it opens

grant-cleary:
Up until now we weren't actually passing the current attribute values into the dialog inputs, so it was easy for users to accidentally overwrite their attributes.


components/plugins/attributes.js
```diff
@@ -61,10 +61,9 @@ export function addAttributesPlugin() {
 						else currentNode.setAttribute(name, val);
 					};
 
-					currentNode.className = dialog.shadowRoot.querySelector('#d2l-attributes-class').value;
-
 					setAttr('title', dialog.shadowRoot.querySelector('#d2l-attributes-title').value);
 					setAttr('id', dialog.shadowRoot.querySelector('#d2l-attributes-id').value);
+					setAttr('class', dialog.shadowRoot.querySelector('#d2l-attributes-class').value);
```
grant-cleary, line 66: This is for consistency. Setting `className=''` leaves an empty `class` attribute on the element, which is unsightly.

components/plugins/attributes.js
```diff
@@ -46,7 +46,7 @@ export function addAttributesPlugin() {
 					title: currentNode.getAttribute('title') || '',
 					id: currentNode.getAttribute('id') || '',
 					className: currentNode.className || '',
-					style: currentNode.style || '',
+					style: currentNode.getAttribute('style') || '',
```
grant-cleary, line 49: `element.style` returns a CSS style declaration object, which is nice, but not what we want (we need a string). We could massage that object into a string representation, or we could take the easy way out and just fetch the attribute directly. I think the latter makes sense given how simple this dialog is.

components/plugins/attributes.js
```diff
@@ -55,16 +55,15 @@ export function addAttributesPlugin() {
 				dialog.opened = false;
 				dialog.attributes = {};
 
-				if (e.detail.action === 'create' && currentNode) {
+				if (e.detail.action === 'create' && currentNode && canAddAttributes(currentNode)) {
```
grant-cleary, line 58: Just a bit of added safety just to make sure we're not blindly dropping attributes on elements we can't add attributes to. If this case does come up, one thing that _does_ happen is that the values in the dialog inputs stay persistent. I don't think it's worth a re-render in this case though; for a user to run into this they would have to find a way to select an invalid node, open the dialog, add a bunch of junk to it, then select another invalid node and try again. Seems unlikely.

grant-cleary: The technically more resilient alternative would be to use TinyMCE's command states to track whether the current selected node is valid and entirely disable the dialog if it is. I'm not opposed to doing that either, though it's a change of behaviour from the old editor, so 🤷 .

dbatiste: Is it even possible for the user to select a non-element node?  i.e. a text node for example?

grant-cleary: Not as far as I know. They would have to use TinyMCE's API.

grant-cleary: Mostly I wanted to add this check because if someone somehow did select a bad node (or somehow we surfaced that bug where selection flips to the body), then at least we would just get a no-op and not a console error.

components/plugins/attributes.js
dbatiste: So I know this code isn't new, but I would suggest that it's not great for the consumer to be piercing the shadow boundary of the dialog's component to get the fields.  Rather, it is better for the dialog component to provide these details as data on a custom close event.

grant-cleary: Yeah, you're entirely right and it should have been designed this way initially. Would you object if I fixed this up in a separate PR?

dbatiste: That's fine yeah.  I know this isn't new and not the focus of this change.

grant-cleary: Sweet. When I was looking over the code I found some other stuff I would do differently if I were writing this today, so I want to take a bit more time and give it more scrutiny. :)

dbatiste: Yeah this is a nice simple plugin/dialog to hone how the general way this should be done.

dbatiste: Another thing is - maybe we can capture visual-diffs for this dialog component.  In theory we don't need the editor at all to capture it, which would also indirectly test its API.

dbatiste (approved):

components/plugins/attributes.js
grant-cleary: Yeah for sure, I think that's a great idea.

PR #313, opened 2021-07-19
Refactor Attributes dialog to better follow web component standards

grant-cleary:
This should bring this dialog more into line with our existing web component standards, as well as add some visual diff tests to keep an eye on its behaviour.


test/dialogs.visual-diff.js
```diff
@@ -0,0 +1,54 @@
...
+		].forEach(info => {
+			it(info.name, async function() {
+				await openDialog(info.selector);
+				await visualDiff.screenshotAndCompare(page, this.test.fullTitle());
+			});
+		});
+
+	});
+
+});
```
grant-cleary, line 54: I thought this should be its own file, mostly because it's a pain getting the viewport right if this is included in the `components` file (since the shim is translucent, the icons show through and are reflected in the screenshot). Not to mention it's likely we'll have several more dialogs eventually so I think it makes sense to give them their own place to live.

test/dialogs.visual-diff.js
```diff
@@ -0,0 +1,54 @@
...
+		await resetDialog('#attributes-dialog-empty');
+		await resetDialog('#attributes-dialog-populated');
+
+		await visualDiff.resetFocus(page);
+	});
+
+	after(async() => await browser.close());
+
+	const openDialog = async(selector) => {
+		await page.$eval(selector, dialog => dialog.opened = true);
```
grant-cleary, line 38: One downside to this method is that we end up focusing the dialog, and since the first field in the dialog is a text input, the caret shows up in the diffs. I think it's valuable to test that focus is getting placed into the editor and disabling the caret isn't so simple because we're several nested web components in at that point, so I propose that we leave it like this for now. I think if these tests prove to be flaky, a reasonable solution may be to deliberately push focus to the document body before we take the screenshot, but I would prefer to only do that if we need to.

test/dialogs.visual-diff.js
```diff
@@ -0,0 +1,54 @@
...
+		await page.bringToFront();
+	});
+
+	const resetDialog = selector => {
+		return page.$eval(selector, async dialog => {
+			return new Promise(resolve => {
+				dialog.opened = false;
+				resolve();
+			});
+		});
```
grant-cleary, line 25: I took this (and the `openDialog`) helper from the core dialog component visual diff tests, just removing the stuff we don't need/want that's specific to those. I figure this is likely generic enough because (imo) it seems reasonable that we would want all of our dialog components to largely open/close in the same manner. It would be a bit odd if we had one dialog that opened via `dialog.open()` and another dialog that opened via `dialog.opened = true;`.

components/plugins/attributes.js
```diff
@@ -168,9 +162,41 @@ class AttributesDialog extends Localizer(RtlMixin(LitElement)) {
...
+		));
+
+		this.attributes = {};
+		this.opened = false;
+	}
+
+	_handleOpen() {
+		// Default values (including <option ?selected="...">) are only populated on first render,
+		// so when the dialog opens, manually adjust the input values.
+		this.shadowRoot.querySelector('#d2l-attributes-title').value = this.attributes.title || '';
```
dbatiste, line 194: This goes against best practices for Lit rendering - i.e. it is modifying the local DOM outside of `render`.  The inputs in `render` should be bound to variables defined in `render`, properties, or the `attributes` property that is already defined.  If the `attributes` object reference is not updated, then `requestUpdate` would be required when its values change to trigger the update.

grant-cleary: This is true, yes, but the problem I found was that even when manually triggering a re-render when the contents of `attributes` changed, this wasn't actually updating these input fields. i.e. Once they have default values set, modifying the value attribute within `render` doesn't actually seem to do anything.

test/dialogs.visual-diff.js
dbatiste: I think the dialog (or rather focus trap) my have some code that prevents doing that.  I think I wrote some code elsewhere for the editor to change the caret color to transparent or something to avoid that issue for the editor.  An alternative may be to force focus to one of the buttons - it would not reflect true dialog focus behavior but would work around the issue for our purpose.

components/plugins/attributes.js
dbatiste: What happens if you use property binding for those values?  (ex. `.value="..."`)

grant-cleary: As in, Lit doesn't seem to want to re-render the inputs even if I wire up code to explicitly request an update when, say, `attributes.title` changes.

grant-cleary: > What happens if you use property binding for those values? (ex. `.value="..."`)

I'll give that a shot. I didn't think that property bindings might have had a different result.

grant-cleary: RE: The `live` directive; that attributes property _should_ be kept up to date, but I'll test and use the directive if necessary.

grant-cleary: Well, turns out that even if `attributes` _is_ kept up to date, the `live` directive is necessary. This seems to work though.

test/dialogs.visual-diff.js
grant-cleary: Yeah, it hasn't flaked so far (maybe it's fast enough that it won't be an issue), but if it does, I think focusing on the button might be our way out. I was a bit torn on that because making sure focus is applied properly seems like a nice thing to test, but if it becomes a problem I would rather have consistent tests. 😄

dbatiste (approved):

PR #315, opened 2021-07-19
Shift visual diff tests into their own job to make re-running them less time-consuming

grant-cleary:
Just interested in trialing this out for now, since as we add more visual diff tests, re-running them in CI becomes more and more painful as long as they're paired with other unit tests and linting.


PaulKrznaric (approved):

.github/workflows/ci.yml
```diff
@@ -25,10 +25,3 @@ jobs:
           SAUCE_USERNAME: Gaudi1
           SAUCE_ACCESS_KEY: ${{ secrets.SAUCE_ACCESS_KEY }}
         run: npm run test:sauce
-      - name: Visual Diff
```
klhuff, line 28: Do we not still want this to run as part of CI?

.github/workflows/visual-diff.yml
```diff
@@ -0,0 +1,23 @@
+name: Visual Diff
+on: pull_request
```
klhuff, line 2: ah, I see this runs when you create a pull request.

grant-cleary: Yup, the names are just names. Once this gets merged I'll set the Visual Diff job as required so it will still need to be green before merge.

klhuff (approved):

PR #316, opened 2021-07-20
 Expose method on editor to toggle between editing and readonly modes

grant-cleary:
This PR exposes a method that allows the inline editor to be toggled into and out of editing mode. It's designed to have no effect when the editor is in fullscreen mode or when using the full editor. To enable this, I had to do some light refactoring, so I'll call that out in comments on the files themselves.


htmleditor.js
```diff
@@ -235,9 +235,9 @@ class HtmlEditor extends HtmlEditorMixin(SkeletonMixin(LitElement)) {
 
 		if (!isShadowDOMSupported) return;
 
-		this.addEventListener('d2l-htmleditor-blur', () => {
+		this.addEventListener('d2l-htmleditor-blur', async() => {
 			if (this.type === editorTypes.INLINE || this.type === editorTypes.INLINE_LIMITED) {
-				this._hideInline();
+				await this._hideInline();
```
grant-cleary, line 240: This doesn't actually need to be `await`ed because we don't actually care about waiting for its result. I still like doing this because it makes it obvious that this is an async operation, even if we don't actually care about its return value.

htmleditor.js
```diff
@@ -364,20 +374,28 @@ class HtmlEditor extends HtmlEditorMixin(SkeletonMixin(LitElement)) {
 
 	_hideInline() {
```
grant-cleary, line 375: This has been refactored to reflect that we now need to care about when this actually "finishes" (hence returning a promise explicitly).

grant-cleary: My gut feeling is that unlike `this._showInline()`, we don't need to wait for initialization to complete before invoking this, on the grounds that the editor always initializes with `this._editing === false` and for it to toggle back to readonly mode, we first implicitly need to wait for initialization to complete in `this._showInline()`. Nonetheless, I've been musing recently on whether it would be better to just make these relevant methods async and wait for initialization ourselves (instead of forcing consumers to do it). i.e. Right now a consumer can call `focus()` on the editor before it initializes and it will have undefined behaviour.

PaulKrznaric (approved):

njostonehouse (approved):

PR #317, opened 2021-07-20
Introduce max-height property on the editor that applies to manual resizing (in full mode) and the HTML block automatic resizing in inline mode

grant-cleary:
Expose a max-height attribute on the editor to allow consumers to limit how tall it can get. This currently only applies in the following situations:
1. Full mode - when manually resizing (not applicable when in fullscreen mode)
2. Inline mode - when rendering the HTML block in inline mode.

We can't meaningfully apply this to the inline mode editor itself because TinyMCE doesn't let us change configurations on the fly. If we applied a max-height to the editor in inline mode, that max-height would persist if the editor was fullscreened, creating some very odd behaviour. This is just meant as an interim solution to try to limit the problem until we can move away from TinyMCE's resizing code.


htmleditor-mixin.js
```diff
@@ -286,6 +288,13 @@ export const HtmlEditorMixin = superclass => class extends Localizer(FormElement
 		return editor;
 	}
 
+	_getMaxHeight() {
+		if (!this.maxHeight) return null;
+
+		const unitlessHeight = this._getUnitlessHeight(this.height);
+		return this.maxHeight > unitlessHeight ? this.maxHeight : unitlessHeight;
```
grant-cleary, line 295: If someone sets a `max-height` that's larger than the existing height, we ignore it and use the existing height as our max-height instead. Note that this also applies in the case where no height is provided. In that scenario, if `max-height < defaultHeight`, we fallback to use the default height instead.

PaulKrznaric (approved):

htmleditor-mixin.js
klhuff: Is it possible that this makes the max height mutable in a way we don't want? eg. I assume the height can change after render.

grant-cleary: Ah, no, `this.height` is a property set on the editor at instantiation and it remains the same through any kind of resizing process. A consumer could certainly change that property, but that would trigger a re-render and there's no reason we would want/care to stop that.

klhuff (approved):

grant-cleary:
Heh, alright, apparently things get a bit weird if you introduce a new required check while PRs are open. I'm going to make the visual diff tests not mandatory for now until existing PRs get cleared off the board.

grant-cleary:
Actually, let me try merging main in first. Maybe that will just solve the problem.

PR #318, opened 2021-07-23
Prevent undefined/null HTML from being set on the editor

grant-cleary:
We don't want the `html` property on the editor to be undefined, null, or not a string, so we should verify that these bad values can't be set.


htmleditor-mixin.js
```diff
@@ -197,6 +197,8 @@ export const HtmlEditorMixin = superclass => class extends Localizer(FormElement
 	}
 
 	set html(val) {
+		if (!val || typeof val !== 'string') val = '';
```
grant-cleary, line 200: If `html` is not a string, stuff can explode down the line. The getter makes sure this is true for most cases, but not if someone comes along and sets `html = undefined` explicitly.

dbatiste (approved):

htmleditor-mixin.js
dbatiste: Seems like something a unit test could cover.  😏

grant-cleary: Hey now, I'm working on it! Had to make them actually work first. 😆

PR #325, opened 2021-07-26
Fix up test framework to actually run tests and add some starters to test HTML setter behaviour

grant-cleary:
Let's get some unit tests working. Woo!


karma.conf.js
```diff
@@ -19,6 +19,11 @@ module.exports = config => {
 				// if you are using 'bare module imports' you will need this option
 				nodeResolve: true,
 			},
+			client: {
+				mocha: {
+					timeout: 5000,
```
grant-cleary, line 24: The default timeout is 2000(ms), which is a little too short for the a11y tests to pass locally. I figure we can bump it to 5s and see if that's sufficient to keep flake away.

test/htmleditor.test.js
```diff
@@ -1,13 +1,39 @@
...
 
-	describe('accessibility', () => {
+	const createEditorComponent = async(attrs) => {
+		let attrString = '';
+		for (const attrName in attrs) {
+			attrString += `${attrName}="${attrs[attrName]}"`;
+		}
+
+		const component = await fixture(html`
+			${unsafeHTML(`<d2l-htmleditor ${attrString}></d2l-htmleditor>`)}
```
grant-cleary, line 15: I tried to find a better documented way to do this with `lit-html` templates, but came up kind of short. We do a lot on first render/update so I sort of wanted to make sure these attributes were initially present in testing. If this is super gross, I can change it so the tests set these properties after first render.

test/htmleditor.test.js
```diff
@@ -1,13 +1,39 @@
...
+
+	describe('accessibility - has title', () => {
 		it('should pass all axe tests', async() => {
-			const el = await fixture(html`<d2l-htmleditor></d2l-htmleditor>`);
-			await expect(el).to.be.accessible();
+			const el = await createEditorComponent({
+				'label': 'I am a title!'
+			});
+			await expect(el).to.be.accessible({
+				ignoredRules: ['landmark-complementary-is-top-level'], // This role nesting is entirely within TinyMCE's DOM
```
grant-cleary, line 35: This rule triggers a failure within TinyMCE's DOM. Its top level `div` has `role="application"` and one of that `div`'s children has `role="complementary"`, which fails this rule. I did take a look and this rule is considered a "best practice" and doesn't run afoul of WCAG standards (as per https://dequeuniversity.com/rules/axe/4.1/landmark-complementary-is-top-level), so I don't think we need to do anything crazy here to work around this in the editor proper.

test/htmleditor.test.js
```diff
@@ -17,4 +43,36 @@ describe('d2l-htmleditor', () => {
...
+	describe('set invalid html', () => {
+		[
+			{ name: 'undefined', html: undefined },
+			{ name: 'null', html: null },
+			{ name: 'not a string', html: new Number(10) }
+		].forEach(info => {
+			it(`when ${info.name}, it should return an empty string`, async() => {
+				const el = await createEditorComponent();
+				el.html = info.html;
+				expect(el.html).to.equal('');
```
grant-cleary, line 55: I have mixed feelings about this, since we're kind of jumbling up testing the setter and the getter here. The alternative here would be to use the `_html` property to set the HTML when we're testing the getter or get the HTML when we're testing the setter respectively. If we think there's value in breaking those out explicitly, I would be fine doing that.

test/htmleditor.test.js
```diff
@@ -1,13 +1,39 @@
...
+	describe('accessibility - has title', () => {
 		it('should pass all axe tests', async() => {
-			const el = await fixture(html`<d2l-htmleditor></d2l-htmleditor>`);
-			await expect(el).to.be.accessible();
+			const el = await createEditorComponent({
+				'label': 'I am a title!'
+			});
+			await expect(el).to.be.accessible({
+				ignoredRules: ['landmark-complementary-is-top-level'], // This role nesting is entirely within TinyMCE's DOM
+			});
```
grant-cleary, line 36: It occurs to me that I could wrap this up into one test suite by doing the not-accessible check first (sans title), then setting a title and checking again. I'm not super familiar with our usual philosophy around how we want to organize these sorts of tests for web components, so I'm fine either way.

test/htmleditor.test.js
grant-cleary: The spacing here is weird. I don't think it will cause a problem, but if I don't end up changing this otherwise, I'll make sure to adjust the spacing before merging.

test/htmleditor.test.js
```diff
@@ -1,13 +1,39 @@
...
+
+		const component = await fixture(html`
+			${unsafeHTML(`<d2l-htmleditor ${attrString}></d2l-htmleditor>`)}
+		`);
+
+		await component.initializationComplete;
+		return component;
+	};
+
+	describe('accessibility - missing label', () => {
```
dbatiste, line 22: Not a big deal... is there a technical reason for splitting these two a11y tests into different `describe` blocks?

grant-cleary: No, I mention that here: https://github.com/BrightspaceUI/htmleditor/pull/325#discussion_r676994962
I'm just not super familiar with how we usually like to organize these things, or how we specifically want to align these things in the editor.

dbatiste (approved):
Just the minor question about organizing the a11y tests.

test/htmleditor.test.js
grant-cleary: Ok, I reorganized the tests to group them by theme (and by API). It seems reasonable to me to organize it like this, but we can always shift things around later if we don't like it.

PR #327, opened 2021-07-27
Refactor Attributes dialog

grant-cleary:
Backport of this PR: https://github.com/BrightspaceUI/htmleditor/pull/313


components/plugins/attributes.js
```diff
@@ -168,9 +163,31 @@ class AttributesDialog extends Localizer(RtlMixin(LitElement)) {
 	}
 
 	_getDefaultDirection(direction) {
-		if (!this.attributes.textDirection) return direction === '';
 		return this.attributes.textDirection === direction;
 	}
+
+	_handleClose(e) {
```
klhuff, line 169: I'm not 100% sure I follow why this is better. Possibly because listening for a d2l-dialog-close event may not be specific enough? And since we're now sending the event ourselves, it's guaranteed to be correct?

klhuff (approved):

components/plugins/attributes.js
grant-cleary: The issue is that prior to this change, the `d2l-dialog-close` event on the editor component was reaching into the dialog's shadowDOM to fetch input values. This violates Lit best practices. Doing it this way ensures that the dialog is solely responsible for its own internals and the information the editor component needs is passed back up to the editor via this new custom event.

grant-cleary: So not a functional issue, just code that could be rewritten to be better. 😄

dmd2l: ⚠️   I'm not very familiar/comfortable with this UI stuff, so my review may be lacking.  That said, nothing is jumping out at me 🤷

dmd2l (approved):

components/plugins/attributes.js
grant-cleary: > ⚠️ I'm not very familiar/comfortable with this UI stuff, so my review may be lacking. That said, nothing is jumping out at me 🤷

No worries; more exposure means more familiarity and more learning! Feel free to ask questions if you run into anything you're unsure of and you want clarification on.

PR #331, opened 2021-07-28
Properly format XSPL iframes within the editor

grant-cleary:
The LMS editor uses a custom plugin to handle XSPL iframes. This didn't get translated into the new editor, so now I'm doing so! I'm once again switching over to use `DOMParser` instead of regex, as it's less fragile overall.

I've also coupled this with some generic refactoring to try to eliminate some of the boilerplate for the various places we do different HTML replacements already.


components/plugins/xspl.js
```diff
@@ -0,0 +1,58 @@
...
+		if (decode) html = tinymce.html.Entities.decode(html);
+		return new DOMParser().parseFromString(html, 'text/html');
+	};
+
+	const convertNestedIframes = (iframe) => {
+		// Decoding has already been completed in the outer iframe processing code
+		const tempDoc = getOriginalHtmlDoc(iframe, false);
+		if (!tempDoc || !tempDoc.body) return;
+
+		tempDoc.body.querySelectorAll('iframe.d2l-xspl-box').forEach(convertNestedIframes);
```
grant-cleary, line 15: This recursion is gross imo, but we do seem to support nesting these. I figure it's either nested `DOMParser`s or nested regex and I would still prefer the former.

Caveat: I wasn't able to test this kind of nesting without falling back to invalid HTML or something that TinyMCE mangled, so I don't know if it can actually happen in practice.

grant-cleary:
Mm, I'm going to need to take another look at these dialog visual diff tests. It looks like occasionally, the animation for the dialog opening isn't finishing before the screenshot is taken. Probably just needs to be wrapped in a rAF.

htmleditor.js
```diff
@@ -285,23 +285,22 @@ class HtmlEditor extends HtmlEditorMixin(SkeletonMixin(LitElement)) {
 		if (!this._context) return this._html;
 		const host = this._context.host;
 
-		const tempDoc = new DOMParser().parseFromString(this._html, 'text/html');
-		if (!tempDoc.body) return this._html;
```
PaulKrznaric, line 289: Should we also be doing a null check on htmlDoc like we used to tempDoc or is it guaranteed safe when we do the body.querySelectorAll?

PaulKrznaric (approved):

htmleditor.js
grant-cleary: So in this case, `this._convertHtml` does such a null check for us. I think it's reasonable to assume in the encoding func that we have a valid document (with a valid body), especially since we're funneling it into a function that verifies that for us.

components/plugins/xspl.js
```diff
@@ -0,0 +1,53 @@
...
+		tempDoc.body.childNodes.forEach(node => iframe.parentNode.appendChild(node));
```
grant-cleary, line 16: Hmm, I need to tweak this a bit because it won't maintain the original DOM ordering. I'll get a new PR up in the morning for that.

PR #335, opened 2021-07-29
Load toolbars on page load, but don't render them until a user clicks into the inline editor

grant-cleary:
When enough editors are loaded on the page at one time, TinyMCE's init and some of the heavier work that comes from rendering the toolbar and its dependencies can slow the browser to a crawl. In inline mode, we can work around that by deferring some of that work, given that the editor first loads in readonly mode.


htmleditor.js
```diff
@@ -241,17 +241,17 @@ class HtmlEditor extends HtmlEditorMixin(SkeletonMixin(LitElement)) {
...
+			await this.updateComplete;
+			this._initializeTinymce(this._toolbar.getPlugins(this), this._toolbar.getExternalPlugins(this), this._toolbar.getTagName, false);
+			return;
 		}
-		await this.requestUpdate();
-		this._initializeTinymce(this._toolbar.getPlugins(this), this._toolbar.getExternalPlugins(this), this._toolbar.getTagName, applyEditorHeight);
+
+		this._toolbar = this.type === editorTypes.INLINE
+			? await import('./components/toolbar/toolbar-inline.js')
+			: await import('./components/toolbar/toolbar-inline-limited.js');
```
grant-cleary, line 254: We need to load the toolbars early, otherwise when the editor switches into editing mode, we get to wait a whopping 400ms for the toolbars and their dependencies to render 😬 . So in this case, we still load the toolbars, but we don't trigger a re-render immediately.

htmleditor.js
```diff
@@ -527,6 +527,12 @@ class HtmlEditor extends HtmlEditorMixin(SkeletonMixin(LitElement)) {
 	}
 
 	async _showInline() {
+		if (!this._initialized) {
```
grant-cleary, line 530: This gets invoked when a user clicks into the editor, so we first check to see if we're already initialized. If we're not, we additionally request an update to pick up the new toolbar state and then initialize TinyMCE.

htmleditor.js
```diff
@@ -527,6 +527,12 @@ class HtmlEditor extends HtmlEditorMixin(SkeletonMixin(LitElement)) {
...
+			this.requestUpdate();
+			await this.updateComplete;
+			this._initializeTinymce(this._toolbar.getPlugins(this), this._toolbar.getExternalPlugins(this), this._toolbar.getTagName, true);
```
grant-cleary, line 533: I wonder if we should guard for the existence of the toolbar here; i.e. if somehow the toolbar isn't loaded by the time we get here, load it first. 🤔

dbatiste: Is it possible for this scenario to happen?  If it's not loaded, it would be loading, correct?  If necessary, we could store a promise for the toolbar loading.

grant-cleary: In theory, for the toolbar not to be loaded by the time we get here, someone needs to have clicked on the editor _before_ `firstUpdated` first runs and the toolbar finishes loading. That window is very short. The only reason I wonder is because if TinyMCE starts its init, but blows up, that's not really a recoverable situation short of reloading the page.

dbatiste (approved):

htmleditor.js
grant-cleary: I added a guard to be safe. I was able to cause JS errors by throttling my connection and clicking as soon as the editor becomes visible.

PR #337, opened 2021-07-29
Clean up toolbar promises

grant-cleary:
Does what it says on the tin. :)


htmleditor.js
```diff
@@ -246,18 +243,16 @@ class HtmlEditor extends HtmlEditorMixin(SkeletonMixin(LitElement)) {
 
 		if (this.type === editorTypes.FULL) {
 			this._toolbar = await import('./components/toolbar/toolbar-full.js');
```
grant-cleary, line 245: I mused about setting the promise here too, but... The promise is only ever used to set a value for `this._toolbar` and it's only ever invoked when the editor is in inline mode (and only on first render/update). So personally I like this better as opposed to adding extra code to keep the promise in sync with this unnecessarily. That's not a strong opinion though.

dbatiste: I don't have a strong opinion, but if it were me I would probably be consistent with the promise assignment.
```javascript
this._toolbarPromise = import('./components/toolbar/toolbar-full.js');
this._toolbar = await this._toolbarPromise;
```

Not initializing the promise in a similar fashion might trip us up later.  But like you said, it's currently not needed for this case.

dbatiste (approved):

htmleditor.js
grant-cleary: Mm, fair. For the sake of consistency I can do this. As we chatted about offline, I want to look at potentially refactoring all of this at some point, but I don't want to predicate anything on that.

PR #345, opened 2021-08-03
Focus on QL dialog close whether or not QLs were inserted

grant-cleary:
Does what it says on the tin! This should make sure focus moves back to the editor properly when this dialog is closed, whether or not QLs were inserted.


ronan-f (approved):
LGTM

PR #347, opened 2021-08-06
DE44690 - Account for fullPage mode when running pre/post-processing on HTML

grant-cleary:
When the `fullPage` plugin is enabled, our pre-processing and post-processing can end up wiping out the document head, so... Don't do that. 😅


dmd2l (approved):

PaulKrznaric (approved):

htmleditor-mixin.js
```diff
@@ -285,12 +285,22 @@ export const HtmlEditorMixin = superclass => class extends Localizer(FormElement
...
 		converters.forEach(convert => convert(tempDoc));
-		return tempDoc.body.innerHTML;
+		if (!this.fullPage || sourceView) return tempDoc.body.innerHTML;
+
+		const docType = tempDoc.doctype;
+		if (!docType) return tempDoc.documentElement.outerHTML;
+
+		let docTypeId;
+		if (docType.publicId) docTypeId = ` PUBLIC "${docType.publicId}"`;
+		else if (docType.systemId) docTypeId = ' SYSTEM';
```
njostonehouse, line 300: Should this append `docType.systemId`?

htmleditor-mixin.js
```diff
@@ -285,12 +285,22 @@ export const HtmlEditorMixin = superclass => class extends Localizer(FormElement
...
+
+		const docTypeString = `<!DOCTYPE ${docType.name}${docTypeId || ''}>`;
```
njostonehouse, line 302: Can `docType.name` be `null`?

htmleditor-mixin.js
```diff
@@ -502,13 +512,13 @@ export const HtmlEditorMixin = superclass => class extends Localizer(FormElement
 						converters.push(htmlDoc => encodeSrcReplaceStrings(htmlDoc, this._context.orgUnitId));
 					}
 
-					e.content = this._convertHtml(e.content, converters);
+					e.content = this._convertHtml(e.content, e.source_view, converters);
 				});
 
-				editor.on('PostProcess', (e) => {
+				editor.on('GetContent', (e) => {
```
grant-cleary, line 518: `PostProcess` was earlier in the flow than it needed to be for this and also didn't contain the `source_view` information we needed.

htmleditor-mixin.js
grant-cleary: Yes, it should. Good catch.

htmleditor-mixin.js
grant-cleary: I don't think it can be if `docType` is defined, but better safe than sorry.

htmleditor-mixin.js
```diff
@@ -285,12 +285,22 @@ export const HtmlEditorMixin = superclass => class extends Localizer(FormElement
...
 		converters.forEach(convert => convert(tempDoc));
-		return tempDoc.body.innerHTML;
+		if (!this.fullPage || sourceView) return tempDoc.body.innerHTML;
+
+		const docType = tempDoc.doctype;
+		if (!docType) return tempDoc.documentElement.outerHTML;
+
+		let docTypeId;
+		if (docType.publicId) docTypeId = ` PUBLIC "${docType.publicId}"`;
+		if (docType.systemId) docTypeId += ` ${docType.publicId ? '' : 'SYSTEM '}"${docType.systemId}"`;
```
grant-cleary, line 300: Ah crap, I need to init `docTypeId` for this to work. One sec.

njostonehouse (approved):

PR #360, opened 2021-08-25
DE41149 - Allow a title to be specified when authoring an HTML document

grant-cleary:
When authoring a full HTML document, we want to allow a `<title>` element to be specified within the document head. Right now this can be done via the source code editor, but it would be ideal if this could be specified by consumers and automatically passed in.


grant-cleary:
We're re-evaluating how this might work before we make any changes here. There are some significant complexities to trying to implement this feature as it is in this PR. So I'm going to close this out for now.

PR #363, opened 2021-08-30
Origin/jschlumkoski/insert element updates

jschlumkoski:
This is a collection of fixes / updates from our internal testing stage.


icons/insert-element.svg
```diff
@@ -1,14 +1,8 @@
...
-          <path id="Shape-2" d="M319.016,512.144a1.685,1.685,0,0,1-1.2-.5l-.085-.085a1.7,1.7,0,0,1,0-2.4l5.332-5.334a1.676,1.676,0,0,1,1.195-.5h.012a1.7,1.7,0,0,1,1.194.5l.086.086a1.706,1.706,0,0,1,0,2.4l-2.755,2.755-2.578,2.578A1.687,1.687,0,0,1,319.016,512.144Zm5.248-7.639h0a.52.52,0,0,0-.369.153l-5.336,5.335a.528.528,0,0,0,0,.743l.085.083a.528.528,0,0,0,.743,0l5.332-5.333a.529.529,0,0,0,0-.743l-.085-.085a.522.522,0,0,0-.371-.153Zm6.466.587a.588.588,0,0,1-.414-.171l-1.173-1.174a.587.587,0,0,1,.826-.83l1.174,1.174a.587.587,0,0,1-.415,1Zm-2.927,0a.586.586,0,0,1-.587-.587v-.587a.587.587,0,0,1,1.174,0v.587a.587.587,0,0,1-.587.586Zm2.928-2.93h-.587a.587.587,0,0,1,0-1.174h.587a.587.587,0,0,1,0,1.174Zm-5.27,0h-.587a.587.587,0,1,1,0-1.174h.587a.587.587,0,0,1,0,1.174Zm.587-1.757a.584.584,0,0,1-.414-.172l-1.173-1.173a.587.587,0,0,1,.828-.827l1.174,1.173a.587.587,0,0,1-.416,1Zm1.76-.587a.586.586,0,0,1-.587-.587v-.583a.587.587,0,1,1,1.173,0v.587a.587.587,0,0,1-.587.587H327.8Z" transform="translate(-317.238 -498.021)" fill="#fff"/>
-        </g>
-      </g>
-    </g>
-  </g>
-</svg>
+<svg width="18" height="18" viewBox="0 0 18 18" xmlns="http://www.w3.org/2000/svg">
+  <style>
```
dbatiste, line 2: This is looking pretty good.  For this particular icon though, we should follow [these guidelines](https://github.com/BrightspaceUI/core/tree/main/components/icons#svg-format). To summarize for this case, we should move the fill color into `fill` attributes, remove the `style` element and class attributes ([example](https://github.com/BrightspaceUI/core/blob/main/components/icons/images/tier1/accelerator.svg?short_path=9f18029)).

jschlumkoski: @dbatiste I've corrected this in my most recent commit.

components/plugins/lcs.js
```diff
@@ -49,11 +49,15 @@ export function addElementEditorDialogPlugin() {
 			dialog.opened = true;
 
 			dialog.addEventListener('d2l-element-editor-dialog-close', e => {
-				if (!e.detail || !e.detail.html) return;
+				if (!e.detail || !e.detail.html) {
+					editor.editorContainer.getRootNode().querySelector('.tox-tinymce-aux').innerHTML = '';
```
grant-cleary, line 53: If someone cancels out of the dialog and immediately clicks on something that can open a context toolbar, does the new context toolbar open properly? Also, where does focus go when the dialog cancels like this? I assume collapsing the selection should set focus/selection in a sensible place.

jschlumkoski: @grant-cleary If someone cancels out of the dialog and immediately clicks on something that can open a toolbar, for example another element, it just opens like normal. Focus stays in the editor, and the caret goes right after the element you were editing when you cancelled.

grant-cleary:
I think linting is still failing, but otherwise this looks good to me!

grant-cleary (approved):

PR #367, opened 2021-09-07
Validate when inline editor is first loaded (deferred) and make sure we don't blow up when blurring the deferred editor

### LINKED ISSUE #366: Validation skipped for type=inline
When using an inline html editor in a form, validation is bypassed if the component hasn't been initialized by focusing the component prior to submitting the form.

Sample:
```html
<d2l-form @submit=${this._submit}>
    <d2l-htmleditor type="inline" id="editor" required></d2l-htmleditor>
</d2l-form>
<d2l-button @click=${this._submitForm}>Submit</d2l-button>

_submitForm() {
    this.shadowRoot.querySelector('d2l-form').submit();
}
```

grant-cleary:
Inline validation broke a little bit when we started deferring TinyMCE initialization, since our validation only worked when TinyMCE initialized or when the HTML setter was invoked. Before we deferred initialization, this covered all of our bases! But now it doesn't.

closes #366


grant-cleary:
FYI @khorwood. Thanks for raising this!

htmleditor.js
```diff
@@ -253,6 +253,7 @@ class HtmlEditor extends HtmlEditorMixin(SkeletonMixin(LitElement)) {
 			return;
 		}
 
+		this._validate(false);
```
grant-cleary, line 256: Doing this on `firstUpdated` (and only when we're in inline mode) guarantees that we won't invoke this more than necessary. The HTML setter re-does validation, as does TinyMCE's initialization, so there shouldn't be any way to get HTML in without triggering validation short of manipulating the DOM.

dbatiste (approved):

htmleditor.js
```diff
@@ -238,7 +238,7 @@ class HtmlEditor extends HtmlEditorMixin(SkeletonMixin(LitElement)) {
 		if (!isShadowDOMSupported) return;
 
 		this.addEventListener('d2l-htmleditor-blur', async() => {
-			if (this.type === editorTypes.INLINE || this.type === editorTypes.INLINE_LIMITED) {
+			if (this.type !== editorTypes.FULL && !this._inlineContainerHidden) {
```
grant-cleary, line 241: This change resolves the explosion mentioned in the second part of the issue. It's entirely possible for the editor to blur while it's still in readonly mode and now that the toolbar loading is deferred, the toolbar container isn't guaranteed to be loaded when this happens. So yeah, just another bit of fallout from the deferred loading business.

PR #368, opened 2021-09-08
Align HTML setter and focus APIs to be eventually consistent on TinyMCE initialization

grant-cleary:
Where it makes sense, we want the editor APIs to at least be eventually consistent as TinyMCE initializes. Effectively, if someone sneaks in pre-init and sets the HTML property (or tries to programmatically focus the editor), we should still at least try to eventually do those things once TinyMCE finishes initializing.


htmleditor-mixin.js
```diff
@@ -280,7 +254,8 @@ export const HtmlEditorMixin = superclass => class extends Localizer(FormElement
 		editor.undoManager.clear();
 	}
 
-	focus() {
+	async focus() {
+		await this.initializationComplete;
```
grant-cleary, line 258: The `htmleditor.js` focus method currently remains unchanged because I'm not sure what the purpose would be of returning a promise here. I'm open to changing that if we want to propagate the promise through though.

htmleditor.js
```diff
@@ -232,6 +232,38 @@ class HtmlEditor extends HtmlEditorMixin(SkeletonMixin(LitElement)) {
...
+				if (window.tinymce) {
+					const editor = tinymce.EditorManager.get(this._editorId);
+					if (editor) {
+						editor.setContent(this._html);
+					}
+				}
+				this.requestUpdate('html', oldVal);
+			};
+
+			if (this.type === editorTypes.FULL) this.initializationComplete.then(setEditorContent);
```
grant-cleary, line 262: It turns out that property getters/setters can't be explicitly async. We could refactor this as a function if we needed to do so.

grant-cleary:
Interesting that the tests are failing. I'll take a look.

dbatiste (approved):

PR #369, opened 2021-09-09
DE45047 - Fullpage HTML conversion shouldn't add the full document to the event unless that was originally provided

grant-cleary:
When the fullpage plugin is active, we don't want to add the full HTML body to TinyMCE's return events unless they were originally provided. This means if the HTML doesn't contain a full document, we don't want to return a full document in the conversion logic.


htmleditor-mixin.js
```diff
@@ -290,7 +290,7 @@ export const HtmlEditorMixin = superclass => class extends Localizer(FormElement
 		if (!tempDoc.body) return html;
 
 		converters.forEach(convert => convert(tempDoc));
-		if (!this.fullPage || sourceView) return tempDoc.body.innerHTML;
+		if (!this.fullPage || sourceView || !/<head[^>]*>.*<\/head>/.test(html)) return tempDoc.body.innerHTML;
```
klhuff, line 293: This is definitely a new syntax for me, but it looks like the gist of this is if the html doesn't include an opened and closed <head> that we return the innerHtml instead of the outer, even if we're in the full page view.

klhuff (approved):

davewagler-d2l (approved):

htmleditor-mixin.js
grant-cleary: Yup, the idea here is that TinyMCE's events can sometimes return the full document, or sometimes just an HTML snippet, depending on context. Since we don't know what that context was, we need some other way to figure out whether we should return the full document ourselves vs. just keeping the snippet as is.

PR #374, opened 2021-09-10
Make sure we clear dialog state before closing them

grant-cleary:
If we don't wipe out the state on our dialogs before closing them, that state can stay "sticky". In reality, I only see this present as an actual problem in the equation editor dialog, but I tackle a couple others to be safe.


grant-cleary:
Note that ISF doesn't appear to need this treatment because the parameters passed into the dialog are fixed when the editor component renders.

components/plugins/image.js
```diff
@@ -269,6 +269,12 @@ class FileSelectorDialog extends RequesterMixin(LitElement) {
 				}
 			);
 
+			this.allowSaveToCourseFiles = false;
+			this.areaFilters = 'Url';
+			this.fileType = fileTypes.All;
+			this.forceSaveToCourseFiles = false;
+			this.responseKeys = 'files';
+			this.uploadHelp = '';
```
grant-cleary, line 277: I'm fairly certain these are technically superfluous because they all appear to be set explicitly on the dialog every time it opens. However, if we ever change the code in a way that makes that false, I think it's safer to explicitly reset them here. The same reasoning applies to the extra code added to the word count dialog.

PaulKrznaric (approved):

PR #376, opened 2021-09-15
Fix fullpage regex to account for newlines within head element

grant-cleary:
Depending on the source of the HTML, newlines can show up within the head block, causing our regex to fail and trigger an early return when it shouldn't.


htmleditor-mixin.js
```diff
@@ -303,7 +303,7 @@ export const HtmlEditorMixin = superclass => class extends Localizer(FormElement
 		if (!tempDoc.body) return html;
 
 		converters.forEach(convert => convert(tempDoc));
-		if (!this.fullPage || sourceView || !/<head[^>]*>.*<\/head>/.test(html)) return tempDoc.body.innerHTML;
+		if (!this.fullPage || sourceView || !/<head[^>]*>.*<\/head>/s.test(html)) return tempDoc.body.innerHTML;
```
grant-cleary, line 306: The `s` global flag causes `.` to match newlines as well.

klhuff (approved):

PR #379, opened 2021-09-15
DE44388 -- Added a fix to enable the HTML Editor toolbar to not be cutoff when inside scrollable regions

mpharoah-d2l:
https://rally1.rallydev.com/#/57732444928d/dashboard?detail=%2Fdefect%2F603964368371

Added a new boolean `escape-scroll-areas` attribute to HTML Editor. When this attribute is off, the HTML editor behaves just like it did before. When the attribute is on, the toolbar will used fixed positioning to "break out" of the scroll area and ensure it doesn't get cutoff.

Keeping the toolbar positioned correctly requires updating it's position every frame, which is why I put it behind the `escape-scroll-areas` attribute instead of always being on. As an optimization, it only does this when the editor is focused, so having 100 editors on the page will still only have one check running instead of 100, and only if the editor is actually focused (showing the toolbar) and the browser tab is open.


htmleditor.js
```diff
@@ -46,10 +46,14 @@ class HtmlEditor extends HtmlEditorMixin(SkeletonMixin(LitElement)) {
...
 			type: { type: String },
+			escapeScrollAreas: { type: Boolean, attribute: 'escape-scroll-areas' },
 			_editing: { type: Boolean },
 			_inlineContainerHidden: { type: Boolean },
 			_inlineEditButtonFocusing: { type: Boolean },
-			_resizing: { type: Boolean }
+			_resizing: { type: Boolean },
+			_trackingToolbar: { type: Boolean },
+			_toolbarX: { type: Number },
+			_toolbarY: { type: Number }
```
grant-cleary, line 56: Nit: Properties should be in alphabetical order (with private properties at the end, also sorted alphabetically.

htmleditor.js
```diff
@@ -220,10 +224,14 @@ class HtmlEditor extends HtmlEditorMixin(SkeletonMixin(LitElement)) {
...
 		this.noBorder = false;
 		this.type = editorTypes.FULL;
+		this.escapeScrollAreas = false;
 		this._editing = false;
 		this._inlineContainerHidden = true;
 		this._inlineEditButtonFocusing = false;
 		this._resizing = false;
+		this._trackingToolbar = false;
+		this._toolbarX = 0;
+		this._toolbarY = 0;
```
grant-cleary, line 234: Same note as above (prefer setting these in alphabetical order).

htmleditor.js
```diff
@@ -437,7 +460,14 @@ class HtmlEditor extends HtmlEditorMixin(SkeletonMixin(LitElement)) {
 		containerStyles.width = this.width;
 
 		const toolbarStyles = {};
-		if (this.toolbarHorizontalOffset) toolbarStyles.right = this.toolbarHorizontalOffset;
+		if (this.escapeScrollAreas && this.type !== editorTypes.FULL) {
+			toolbarStyles.position = 'fixed';
+			toolbarStyles.top = `${this._toolbarY}px`;
+			toolbarStyles.left = `${this._toolbarX}px`;
+			toolbarStyles.zIndex = 1000;
+		} else if (this.toolbarHorizontalOffset) {
```
grant-cleary, line 468: This does mean the offset won't be usable with this feature, but I think that's likely fine here.

htmleditor.js
```diff
@@ -321,6 +336,14 @@ class HtmlEditor extends HtmlEditorMixin(SkeletonMixin(LitElement)) {
 		return this._convertHtml(this._html, false, [encode]);
 	}
 
+	_handleFocusIn() {
+		super._handleFocusIn();
+		if (this._shouldTrackToolbar() && !this._trackingToolbar) {
+			this._trackingToolbar = true;
+			this._updateToolbar();
+		}
+	}
```
grant-cleary, line 345: I'm wondering if all of these props/functions should move into the mixin. The focus handling stuff is all done in the mixin, so this sets up a bit of a weird situation where the mixin calls this function, which then calls back to the mixin. It might be better to just integrate this logic into the existing `_handleFocusIn` function in the mixin and move the props/helper functions over. That way the core editor class can call functions and check props in the mixin to accomplish this, but not vice versa.

grant-cleary: Though, hmm... `_updateToolbar` has knowledge about the editor's DOM structure.

grant-cleary: Yeah okay, and there's more in this that crosses the line between the mixin and this class (e.g. checking the type). I'm not sure there's a great way to entangle that.

mpharoah-d2l:
Getting this error in tests now even though all I did was reorder the variables
>      validation
>      max-length
>        ✖ should throw validation error on blur when over max-length
>        ✔ should not throw validation error on blur when live validation is disabled
>      required
>        ✖ should throw validation error on blur when empty
>        ✔ should not throw validation error on blur when live validation is disabled

htmleditor.js
```diff
@@ -46,10 +46,14 @@ class HtmlEditor extends HtmlEditorMixin(SkeletonMixin(LitElement)) {
 			noBorder: { type: Boolean, attribute: 'no-border' },
 			toolbarHorizontalOffset: { type: String, attribute: 'toolbar-horizontal-offset' },
 			type: { type: String },
+			escapeScrollAreas: { type: Boolean, attribute: 'escape-scroll-areas' },
```
grant-cleary, line 49: (Same note RE: alpha ordering here and in the ctor)

grant-cleary:
> Getting this error in tests now even though all I did was reorder the variables

Hmm, I'll give them a re-run and see if it changes anything.

htmleditor.js
mpharoah-d2l: Fixed

grant-cleary:
> EDIT: Still broken. I'll pull this branch locally and see if I can reproduce. These tests haven't been flaky in the past but reordering some properties obviously shouldn't change anything.

Ok, I think I see what's going on here. I'm not super sure how this passed before you re-ordered things. I think it might be because validation changed in main between that CI run and now. I'll suggest a fix.

htmleditor.js
```diff
@@ -277,6 +285,13 @@ class HtmlEditor extends HtmlEditorMixin(SkeletonMixin(LitElement)) {
 		return this._renderInlineEditor();
 	}
 
+	updated() {
+		if (this._shouldTrackToolbar() && !this._trackingToolbar) {
```
grant-cleary, line 289:
```suggestion
	updated(changedProperties) {
		super.updated(changedProperties);
		if (this._shouldTrackToolbar() && !this._trackingToolbar) {
```

grant-cleary: Validation handling happens on the `updated` lifecycle in the mixin, so basically this inclusion (without the `super` call) was short-circuiting that and preventing validation from running at all when the tests changed the `html` property.

grant-cleary: The tests themselves are new as of a week or so ago, so your previous commit still broke validation; it just didn't surface in any tests.

grant-cleary (approved):

PR #380, opened 2021-09-16
Allow opt-out of live validation (on editor blur)

grant-cleary:
Add an optional property to disable live validation (on blur). Validation will still occur on form submission.


dbatiste (approved):
Any way we can capture this sorta thing in a test?

grant-cleary:
> Any way we can capture this sorta thing in a test?

Hmm, probably. I think we can read the validation state from the mixin, so it might be enough to purposely trigger a blur event on the editor and then check that state. Let me poke around.

PR #381, opened 2021-09-17
Don't try to close the dropdown if it doesn't exist

grant-cleary:
It's possible for this little bit of code to produce JS console errors under some circumstances. This doesn't appear to impact actual functionality, so for now we'll add a quick check to keep the noise down while we do further investigation.


components/toolbar/button-menu.js
```diff
@@ -189,7 +189,7 @@ class ButtonMenu extends DropdownOpenerMixin(ToolbarItemMixin(FocusVisiblePolyfi
 
 		const content = this.shadowRoot.querySelector('#content').assignedNodes()
 			.filter(node => node.hasAttribute && node.hasAttribute('dropdown-content'))[0];
```
grant-cleary, line 191: If the array produced by `filter` is empty, `[][0] === undefined`, which triggers an error when trying to call `close()`.

dbatiste (approved):

PR #384, opened 2021-09-29
Use new ifrau version and new FRA detection helper

grant-cleary:
Now that we can import ifrau and construct a slim client explicitly that way, let's do it! There are a few caveats here, but I'll note those inline.


components/lms-adapter.js
```diff
@@ -1,13 +1,14 @@
-let configuredIfrauClient;
+import { isFramed } from '@brightspace-ui/core/helpers/framed.js';
```
grant-cleary, line 1: This is cached very aggressively and can change at most once after it's resolved (by _any_ component), so I'm not super concerned about this usage. The alternative would be to pass the component around (or at least `_fraContext`) to all of these adapter functions, which I'm not keen on.

components/lms-adapter.js
```diff
@@ -1,13 +1,14 @@
...
+
+let ifrauClient;
 
 export async function getIfrauClient() {
-	if (configuredIfrauClient === undefined && window.ifrauclient) {
-		configuredIfrauClient = await window.ifrauclient({
-			resizeFrame: false
-		}).connect();
+	if (!ifrauClient && await isFramed()) {
```
grant-cleary, line 6: `isFramed` has no dependence on the ifrau client, so we no don't need to worry about this breaking or returning the wrong thing if the FRA is still using an old client version.

components/plugins/equation.js
```diff
@@ -182,7 +182,7 @@ export function addEquationPlugin() {
 				const lang = document.documentElement.getAttribute('lang') || 'en';
 
 				let text;
-				if (window.ifrauclient) {
+				if (component._fraContext) {
```
grant-cleary, line 185: I felt like in this case (and in other cases where we had access to the component), checking this property makes more sense than importing `isFramed` and checking that again.

htmleditor-mixin.js
```diff
@@ -256,6 +257,7 @@ export const HtmlEditorMixin = superclass => class extends Localizer(FormElement
 		if (!isShadowDOMSupported) return;
 
 		this._context = await getContext();
+		this._fraContext = await isFramed();
```
grant-cleary, line 260: So the changes in this file represent the biggest change here, which is that we can't bank anymore on `this._fraContext` being set synchronously. This also does mean that we're inevitably going to generate another render here if we are indeed in an FRA. I'm a bit worried about this, but I don't know what other options we have here. Technically with our reliance on `window.ifrauclient`, we were still banking on FRAs setting that in a timely manner via importing the client script from the CDN, so I'm not sure this is actually adding that much more risk... It's just transferring it from the FRAs to the editor.

grant-cleary:
I have it in my personal backlog to investigate moving off of karma and onto the `web-test-runner` stuff. If we did that, I think we gain the ability to mock out module imports, which would make it possible to test whether `_fraContext` resolves properly. As it stands, we can't really do that without setting up an iframe and attaching the appropriate listeners (and an ifrau host) to it.

PR #423, opened 2021-10-20
Use new reactive controller to check for the addition of the editor's context object

grant-cleary:
It's possible (specifically in an FRA) for the HTML editor to render before the context object it relies on is written to the document's `html` element. We can work around this by using a reactive controller to watch for changes to that object and update the editor accordingly.


htmleditor-mixin.js
```diff
@@ -336,14 +337,26 @@ export const HtmlEditorMixin = superclass => class extends Localizer(LabelledMix
...
 		super.firstUpdated(changedProperties);
 
 		if (!isShadowDOMSupported) return;
 
-		this._context = await getContext();
+		this._context = getContext(this._contextController);
+
 		this._fraContext = await isFramed();
-		this.provideInstance('context', { ...this._context, ...{ component: this } });
+		const host = await getHost();
```
grant-cleary, line 358: I think we want to fetch this _before_ we pull in the context object, since pulling in the context object will trigger an update and we want this to be bundled in.

grant-cleary: Actually, I think what we _really_ want to do here is manually `requestUpdate()` after this point if we're in inline mode and TinyMCE hasn't yet initialized. We don't need the manual update otherwise because we're either covered by the context observer forcing an update or by the toolbar instantiation forcing an update.

components/lms-adapter.js
```diff
@@ -1,3 +1,4 @@
+import { HtmlAttributeObserverController } from '@brightspace-ui/core/helpers/htmlAttributeObserverController.js';
```
grant-cleary, line 1: This should probably be moved now that `core` has a `controllers` directory. I'll do that first.

htmleditor-mixin.js
```diff
@@ -336,14 +337,26 @@ export const HtmlEditorMixin = superclass => class extends Localizer(LabelledMix
...
+	connectedCallback() {
+		super.connectedCallback();
+		this._contextController.hostConnected();
+	}
+
+	disconnectedCallback() {
+		super.disconnectedCallback();
+		this._contextController.hostDisconnected();
+	}
+
```
Trandrew17, line 349: What are these Callbacks for?

grant-cleary: So this hooks into these callbacks in the context controller: https://github.com/BrightspaceUI/core/blob/main/controllers/attributeObserver/htmlAttributeObserverController.js#L21

The intent here is to mimic Lit 2.0 reactive controllers (https://lit.dev/docs/composition/controllers/). If we were on Lit 2.0 and could actually implement that controller interface, we could ditch these because Lit would handle the hook-ups itself. More or less the controller needs to connect/disconnect whenever the host connects/disconnects from the DOM.

Trandrew17 (approved):

PR #436, opened 2021-10-26
Remove global event handlers when setting content and add filtering/processing in readonly mode

grant-cleary:
We ideally don't want these attributes to be possible outside of fullpage contexts.


components/lms-adapter.js
```diff
@@ -258,3 +258,32 @@ export async function getTempFile(src, fileId) {
...
+export async function getViewableHtml(html, orgUnitId, noFilter) {
+	const filterMode = noFilter ? 'None' : 'Strict';
+
+	if (await isFramed()) {
+
+		if (!htmlConversionService) {
+			htmlConversionService = await ifrauClient.getService('convert-to-viewable-html', '0.1');
+		}
+
+		return htmlConversionService.filterHtml(html, filterMode);
```
grant-cleary, line 273: This requires an LMS change.

components/plugins/filter-global-handlers.js
```diff
@@ -0,0 +1,27 @@
...
+		for (let i = 0; i < elm.attributes.length; i++) {
+			const attrName = elm.attributes[i].name;
+			if (!attrName.startsWith(handlerPlaceholderPrefix)) continue;
+
+			const handler = attrName.split(`${handlerPlaceholderPrefix}-`)[1];
+			elm.setAttribute(handler, elm.attributes[i].value);
+			elm.removeAttribute(attrName);
+		}
+	});
+}
```
grant-cleary, line 27: I decided to break these out into a plugin, since it keeps the editor a bit cleaner and we could add more here later if we wanted.

grant-cleary: I also opted to take the safer route of temporarily storing the handler values in their own attributes and then adding them back on fetch.

htmleditor-mixin.js
```diff
@@ -638,7 +643,8 @@ export const HtmlEditorMixin = superclass => class extends Localizer(LabelledMix
 
 					e.content = this._convertHtml(e.content, e.source_view, [
 						convertImagePlaceholderToOriginalContent,
-						decodeSrcReplaceStrings
+						decodeSrcReplaceStrings,
+						restoreGlobalEventHandlers
 					]);
```
grant-cleary, line 648: I no longer restrict this to non-fullpage editors now that we restore this behaviour when fetching the content (no risk of breaking client code).

grant-cleary:
Ok, so something in our `runAsync` directive wasn't playing nicely with the html-block template. Using a combination of Lit's `guard` and `until` directives does the trick in this case, however.

PaulKrznaric (approved):

dmd2l (approved):

PR #440, opened 2021-11-01
Origin/jschlumkoski/insert element revisions

jschlumkoski:
This is changes to the LCS Insert Element tool based on internal feedback.


components/plugins/lcs.js
```diff
@@ -432,48 +533,62 @@ class ElementEditorDialog extends LitElement {
...
+					<d2l-input-textarea id="${id}" data-property-key="${property.key}" data-type="textarea" .value="${live(property.value)}" placeholder="${property.placeholder}" ?required="${property.required === true}" label-hidden label="${label}"></d2l-input-textarea>
+				</div>
+				`;
+
+			case 'htmleditor':
+				return html`
+				<div class="d2l-htmleditor-templates-field">
+					<label class="d2l-input-label ${property.required ? 'd2l-input-label-required' : ''}" aria-hidden="true">${label}</label>
+					${property.help ? html`<d2l-button-icon id="help_${property.key}" class="d2l-help" text="Additional Information" icon="tier1:help" @click="${this._openHelp}"></d2l-button-icon><d2l-dialog title-text="${label} Information"><div class="d2l-help-content">${property.help}</div></d2l-dialog>` : ''}
+					<d2l-htmleditor id="${id}" data-property-key="${property.key}" .html="${live(property.value)}" ?required="${property.required === true}" label-hidden label="${label}"></d2l-htmleditor>
```
grant-cleary, line 568: I'm unsure about this. This is something I haven't run into with Lit before, so it's out of my area of expertise, but @dbatiste might know... Is nesting a custom element like this okay?

dbatiste: Like you mean and editor inside of an editor?

grant-cleary: Yeah. The inner editor doesn't have Insert Element enabled so it's not like we can get a forever editor situation going, but I'm not sure if this poses any kind of problem in and of itself, given that there's a bunch of common stuff that all editors on the page share.

dbatiste: From a custom element perspective, it is valid, but I would be surprised if everything worked as expected.

jschlumkoski: @grant-cleary @dbatiste In my testing, everything seems to function as it should. I haven't tested every facet of it, but things like images, quicklinks, and general formatting all work as expected. The only weirdness I've encountered is some overflowing that happens to the dialog window when contextual toolbars appear in one of the sub-HTML Editors.

grant-cleary: My main concern is more subtle issues. Like you mentioned, there's already some weirdness from having dialogs within dialogs, but I wouldn't be surprised if other stuff was a bit strange here. It's also a small thing, but this works because we know the editor's custom element must be loaded on the page already (because it's required to load the outer editor) and we sort of piggy-back on that here. Otherwise we would end up with a circular import reference between here and the core editor code.

Is there a reason we need full rich text support here?

jschlumkoski: @grant-cleary Since it was demoed at Fusion, we've had people who have seen / tested it routinely ask if it will support more than just text. There's a desire to be able to include things like lists, formatted text, images, etc. in certain elements, which is something that can be done today with the elements if you were building them manually.

@bohlc may be able to offer more insight here.

grant-cleary: Thanks for the context. I'm still a bit leery about this though. My main concerns are that:
1. The editor might be broken here in subtle and unpredictable ways. It's _really_ hard to test for this kind of use case because inputs are arbitrary.
2. Adding this in implies that we need to support this nested editor-in-an-editor workflow in the future, which seems not ideal given the ways this could go wrong.

I'm not really sure if there's a good answer to this problem, but I'm not super comfortable with the idea of nesting an editor within the editor directly like this. The only place this happens right now is in some LMS dialogs, but those are a different scenario because those dialogs are iframed and the editor is being loaded in a completely different document.

jschlumkoski: @grant-cleary Your concerns sound totally valid to me! Out of curiosity, would it be better if we limited the options of the editor that appeared in our tool? Is that even possible?

grant-cleary: It's not really possible, no. The only exception would be to use `type="inline-limited"`, since that editor is much less feature-rich, but the inline editors come with their own caveats that worry me in a similar vein. I'm not 100% sure what the right answer is here.

grant-cleary: Ok, after today's meeting I think the one thing I'm curious about is what the weird context menu case you referred to looks like. Do you happen to have a screenshot I can take a look at? I think if we can be reasonably confident that nothing else is broken and we're keeping functionality as limited as possible, that's likely okay. I'm just curious how TinyMCE's UI pieces behave (context menus, a11ychecker, Insert Emoji dialog, etc.).

grant-cleary: Got your email. I think if that's the main issue we're seeing with this, I'm not super concerned and this can just go forward. If we run into other issues later, we can revisit as needed.

grant-cleary (approved):

PR #447, opened 2021-11-02
Add shortcut key combination to open code editor

grant-cleary:
We want to add a new shortcut key to trigger opening the source code editor (Ctrl+Shift+E/Command+Shift+E for Windows/Mac respectively).


components/plugins/source-code.js
```diff
@@ -0,0 +1,30 @@
...
+const name = 'd2l-sourcecode';
+
+export function addSourceCodePlugin() {
+	tinymce.PluginManager.add(name, function(editor) {
+
+		const context = requestInstance(editor.getElement(), 'context');
+		const component = context.component;
+
+		const handleShortcutKey = e => {
+			if (e.key !== "E") return;
```
grant-cleary, line 13: `KeyboardEvent.key` is supported across the browsers we support (https://caniuse.com/keyboardevent-key) and I think it's more elegant since it knows about casing. Saves having to check `e.shiftKey` in the logic below.

components/toolbar/toolbar-full.js
```diff
@@ -84,6 +85,7 @@ export function getPlugins(component) {
 		component.fullPage && allowInsertElement(component) ? addElementEditorDialogPlugin() : '',
 		hasLmsContext(component) ? addPreviewPlugin() : 'preview',
 		addQuicklinkPlugin(),
+		component.sourceEditable || (component._context && component._context.sourceEditable) ? addSourceCodePlugin() : '',
```
grant-cleary, line 88: I left this logic in place when deciding whether to load the base `advcode` plugin as well. Given that it loads codemirror, it's expensive enough that I think extra guards to avoid loading it unnecessarily are nice.

dmd2l (approved):

PaulKrznaric (approved):

PR #465, opened 2021-11-17
Add linting for lang files/translations

grant-cleary:
Linting for lang files. Hopefully this will keep translation errors down to a minimum.


lang/ja.js
```diff
@@ -78,7 +78,7 @@ export default {
 	"table:column:pasteBefore": "前に列を貼り付け",
 	"table:delete": "表の削除",
 	"table:insert": "表の挿入",
-	"table:insert:columsByRows": "{columns, plural, one { 1 列 } other { { columns } 列 }} × {rows, plural, one { 1 行 } other { { rows } 行 }}",
+	"table:insert:columsByRows": "{columns, plural, other { { columns } 列 }} × {rows, plural, other { { rows } 行 }}",
```
grant-cleary, line 81: These should end up re-translated since the `en-us` terms were changed. I don't really want to make any guesses about how these translations should be formatted in the interim, however.

lang/en.js
```diff
@@ -78,7 +78,7 @@ export default {
 	"table:column:pasteBefore": "Paste Column Before",
 	"table:delete": "Delete Table",
 	"table:insert": "Insert Table",
-	"table:insert:columsByRows": "{columns, plural, one { 1 column } other { { columns } columns }} by {rows, plural, one { 1 row } other { { rows } rows }}",
+	"table:insert:columsByRows": "{columns, plural, one { 1 column by {rows, plural, one { 1 row } other { { rows } rows }} } other { { columns } columns by {rows, plural, one { 1 row } other { { rows } rows }} }}",
```
grant-cleary, line 81: This is more correct. In English, a phrase like `1 column by 2 rows` can be broken up into two independent parts, i.e. `1 column` is independent of `2 rows` and vice versa. That's not a given in all languages, however, so it's better to nest these plural cases.

lang/en.js
```diff
@@ -110,8 +110,8 @@ export default {
 	"wordcount:dialog:title": "Word Count",
 	"wordcount:dialog:words": "Words",
 	"wordcount:footer:characterCount": "{count, plural, one { 1 Character } other { { count } Characters }}",
-	"wordcount:footer:selectionCharacterCount": "{count, plural, one { 1 Character } other { { count } Characters }} in selection",
-	"wordcount:footer:selectionWordCount": "{count, plural, one { 1 Word } other { { count } Words }} in selection",
+	"wordcount:footer:selectionCharacterCount": "{count, plural, one { 1 Character in selection } other { { count } Characters in selection }}",
```
grant-cleary, line 113: Similar to the above; how `in selection` is translated could depend on the pluralization of `Character`, so include it in the pluralization context.

lang/ja.js
```diff
@@ -78,7 +78,7 @@ export default {
 	"table:column:pasteBefore": "前に列を貼り付け",
 	"table:delete": "表の削除",
 	"table:insert": "表の挿入",
-	"table:insert:columsByRows": "{columns, plural, one { 1 列 } other { { columns } 列 }} × {rows, plural, one { 1 行 } other { { rows } 行 }}",
```
grant-cleary, line 81: `ja`, `ko`, `zh-cn`, and `zh-tw` don't use different translations for single vs. plural forms, so they can be omitted here.

package-lock.json
```diff
@@ -2154,6 +2154,12 @@
         }
```
dmd2l, line 2154: Probably just my lacking understanding this area, but why does all this stuff need to be 'locked' in order to change language terms?  😕

dmd2l:
Painfully out of expertise area here, but can at least saying nothing glaring stands out 🤷

dmd2l (approved):

package-lock.json
grant-cleary: `package-lock.json` is an NPM thing meant to ensure that anyone pulling and installing this repo will get deterministic versions for its dependencies (https://docs.npmjs.com/cli/v8/configuring-npm/package-lock-json). Otherwise, NPM would just pull all of the latest versions allowed in `package.json` every time you installed, and you could in theory get different dependencies when building a few minutes apart if a package happens to update in that gap.

So because this change adds new dependencies, `package-lock.json` gets updated with the current latest allowed versions of those dependencies.

grant-cleary: Yep, it's generated via npm whenever you `npm i`.

lang/en.js
bearfriend: Ideally, this is:

```
"{columns, plural, one {{rows, plural, one { 1 column by 1 row } other { 1 column by { rows } rows }} } other {{rows, plural, one {{ columns } columns by 1 row} other {{ columns } columns by { rows } rows }}}}",
```

package-lock.json
dmd2l: Should these one have the caret remove?

image

I was just looking what that meant, and found `^version: will update to all future minor versions, without incrementing the major version. ie ^2.3.4 will use releases from 2.3.4 to <3.0.0.`   I'm think you don't want that to keep the deterministic version?  🤔

lang/ja.js
bearfriend: The funny thing is it's probably just `{columns} 列 × {rows} 行`

package-lock.json
grant-cleary: No, these `requires` blocks are just used to tell NPM what versions various subdependencies are allowed to range over. Ultimately the top-level dependencies are still deterministic.

lang/en.js
grant-cleary: Ah, right, that nesting makes more sense.

package-lock.json
bearfriend: Those are just a list of what that dependency itself requires. Up one level in that object, all those dependencies will be listed separately, with deterministic versions that satisfy these `^` versions.

lang/ja.js
grant-cleary: Haha yeah, that's my guess as well. Oh well, I guess we'll churn out a chunk of money to figure that out for sure. 😅

bearfriend: They won't even do it correctly. They never do. It'll come back with `one` cases, and we'll have to correct it ourselves. After we pay them! 😆

grant-cleary: You're not wrong...

bearfriend (approved):

dmd2l (approved):

PR #470, opened 2021-11-23
Atran/DE45938

Trandrew17:
Adjustment to `setDefaultImageStyle` where the full html after the img tag was being removed. 
This was meant for single images so the fix is to only apply attributes and styles if there is only an img tag.


components/plugins/isf.js
```diff
@@ -167,10 +169,25 @@ export function addISFPlugin() {
 				|| img.hasAttribute('width') || img.hasAttribute('height')
 				|| img.style.width || img.style.height) return;
 
-			img.style.maxWidth = '100%';
-			img.setAttribute('data-d2l-editor-default-img-style', 'true');
+			let imgEndIndex = html.indexOf('>');
+			if (html.substr(imgEndIndex - 1, 1) === '/') {
```
grant-cleary, line 173: Consider refactoring this to use `substring` as well. I believe `substr` is deprecated.

grant-cleary:
As it stands, this change will still apply to any `SetContent` action whose content happens to contain a single `img` tag. This means it can still fire when content is set outside of Insert Image and ISF. Is that what we want here?

components/plugins/isf.js
```diff
@@ -167,10 +169,25 @@ export function addISFPlugin() {
 				|| img.hasAttribute('width') || img.hasAttribute('height')
 				|| img.style.width || img.style.height) return;
 
-			img.style.maxWidth = '100%';
-			img.setAttribute('data-d2l-editor-default-img-style', 'true');
+			let imgEndIndex = html.indexOf('>');
+			if (html.substring(imgEndIndex - 1, 1) === '/') {
```
Trandrew17, line 173: This isn't doing what I think it does.

grant-cleary: Yeah, that second param is an end index. It's not quite as straightforward to use it like this.

Trandrew17:
> As it stands, this change will still apply to any `SetContent` action whose content happens to contain a single `img` tag. This means it can still fire when content is set outside of Insert Image and ISF. Is that what we want here?

Hmm, I think ideally no. We would not want this to fire outside of Insert Image and ISF.

grant-cleary:
> Hmm, I think ideally no. We would not want this to fire outside of Insert Image and ISF.

Ick. If that's the case, I'm not 100% sure TinyMCE's event handlers are the best way to do this. I think when this was implemented there perhaps wasn't a clear understanding that this would apply *any* time content is set in the editor. If it's meant to be restricted to when ISF spits out an image, I think we need to re-evaluate where this happens. =/

Trandrew17:
> > Hmm, I think ideally no. We would not want this to fire outside of Insert Image and ISF.
> 
> Ick. If that's the case, I'm not 100% sure TinyMCE's event handlers are the best way to do this. I think when this was implemented there perhaps wasn't a clear understanding that this would apply _any_ time content is set in the editor. If it's meant to be restricted to when ISF spits out an image, I think we need to re-evaluate where this happens. =/

Yeah, I think this would help if it was urgent but we expect this was like this for the last 4 years. Back to the drawing board I think.

PR #473, opened 2021-11-23
fix: DE45221/DE42293 - HTML editor dropdowns misaligned in fullscreen mode

annabelsegalld2l:
This PR fixes the HTML editor defect DE42293, as well as Quizzes DE45221.

I added a new demo component to show the bounded behavior. The bug and fix are easily togglable by removing the one-line CSS. 

Explanation: The dropdown-content can detect when it is in a bounded container in the LMS. This is useful to prevent the editor dropdowns form overflowing to another portion of the page. For example, FACE assignments uses the bounding of the primary-secondary template, making it look more clean. 
html-mini-correct-bound

The issue arises when fullscreen mode is toggled - the dropdown still identifies the bound as the primary template, even though the bound is just the full screen, causing shifted dropdowns. 
html-fullscreen-incorrect-bound

The fix for this is to prevent the html-editor from allowing overflows in fullscreen mode. This makes logical sense as if the dropdown flowed off the screen, we would not be able to see the contents. Adding overflow: auto tells the dropdown to treat the editor as the bound, rather than the template div. So we can use the whole screen!

I did some preliminary regression testing on Chrome to double-check that this change wouldn't break any editor behaviour. Looks like this change fixes the issue in FACE assignments and the quizzing preview. Otherwise, editors where the issue was not present beforehand are still functioning correctly (ie. Rubrics, MVC test page).


htmleditor.js
```diff
@@ -126,6 +126,11 @@ class HtmlEditor extends HtmlEditorMixin(SkeletonMixin(LitElement)) {
 				height: 100%;
 				position: relative;
 			}
+			/* DE42293 - use HTML editor as bounding rectangle
+			 to prevent mis-aligned toolbar dropdowns */
+			.d2l-htmleditor-fullscreen.d2l-htmleditor-container {
+				overflow: auto;
+			}
```
grant-cleary, line 133: This seems reasonable. I can't imagine the overflow having significant negative impacts on our (or TinyMCE's dialogs), given that we can't overflow in fullscreen mode anyways.

tylergee: The changes looks fine to me. Thanks for finding a solution! Do we think there might be any scenario whereby the use of overflow auto can result in unwanted scrollbars popping up?

grant-cleary: I believe we'll still see scrollbars if the contents of the editor ever extend beyond the bounding container, but I don't think it's likely for anything else to crop up.

grant-cleary:
I feel pretty comfortable with this change as is and wouldn't be opposed to it opening up for review. :)

htmleditor.js
```diff
@@ -126,6 +126,11 @@ class HtmlEditor extends HtmlEditorMixin(SkeletonMixin(LitElement)) {
 				height: 100%;
 				position: relative;
 			}
+			/* DE42293 - use HTML editor as bounding rectangle
+			 to prevent mis-aligned toolbar dropdowns */
+			.d2l-htmleditor-fullscreen.d2l-htmleditor-container {
```
dbatiste, line 131: The reason why this fixes this issue might not be immediately obvious.  I _think_ it relates to this code in [getBoundingAncestor](https://github.com/BrightspaceUI/core/blob/main/helpers/dom.js#L29) - yes? Just commenting so we have some reference in the PR to it in case future-us is wondering why this is the solution.  We might also want to mention the helper in the code comment so that if someone changes that helper and does a code search, they'll find this.

tylergee:
Just wondering if we have a estimated time for merging this change.

annabelsegalld2l:
> Just wondering if we have a estimated time for merging this change.

Grant is back from vacation on Monday. Looks like he didn't have any further comments so we should be able to merge it then.

dbatiste:
Everything looks fine from my perspective.  There was just my question above about the reason why the fix works.  Grant added a 👍 to my question - I'm not sure whether that is a confirmation that my assumption is correct, or if he also was curious.

annabelsegalld2l:
> Everything looks fine from my perspective. There was just my question above about the reason why the fix works. Grant added a 👍 to my question - I'm not sure whether that is a confirmation that my assumption is correct, or if he also was curious.

Your assumption is correct, Dave. The boundingAncestor function is looking for any overflow that is not "visible". I will add a comment there with a link to the function.

dbatiste (approved):

grant-cleary (approved):

PR #490, opened 2021-12-02
removing web component polyfill

dlockhart:
Was in here doing something else and noticed this. No need for this polyfill anymore (especially as a non devDependency).


dbatiste (approved):
Thanks Dave.

@grant-cleary : we also still have the shadyDOM check and the alert in place.  Eventually we'll want to remove that as well, but for the moment it's probably not a bad idea to keep it around in case a user with an old unsupported browser lands on the page with the editor.

grant-cleary:
> Thanks Dave.
> 
> @grant-cleary : we also still have the shadyDOM check and the alert in place. Eventually we'll want to remove that as well, but for the moment it's probably not a bad idea to keep it around in case a user with an old unsupported browser lands on the page with the editor.

Yeah, seems reasonable to me (and thanks as well from me, Dave!). I feel like how long that will be seems to be a bit nebulous... It's really hard to get unsupported versions of Chrome/Firefox installed, but I guess there are likely still old Edge users hanging around?

grant-cleary (approved):

PR #503, opened 2021-12-13
Consolidate dependency updates under dependabot

grant-cleary:
The action worked reasonably well, but conflicted with Dependabot a bit too much for my liking (and freaked out when a visual diff test failed). I think it makes more sense to consolidate updates under Dependabot instead.


.github/dependabot.yml
```diff
@@ -3,4 +3,4 @@ updates:
   - package-ecosystem: "npm"
     directory: "/"
     schedule:
-      interval: "weekly"
+      interval: "daily"
```
grant-cleary, line 6: Weekly is a little too long, given that we would ideally like to keep abreast of `core` changes on a quicker cadence. If `daily` gets too overwhelming, we can swap it back, but I think Dependabot is clever enough to keep the noise down and this will give us better control over which dependency updates we take.

Trandrew17 (approved):

grant-cleary:
I've disabled auto-merge in this repo given this change.

PR #511, opened 2021-12-15
Rollback to previous advcode plugin version to fix screenreader forms/focus mode bug

grant-cleary:
Does what it says on the tin. This appears to function correctly, but I can't yet confirm whether it fixes the issue completely. The goal here is to get this merged and into a test site so we can take a confirm that it solves the issue (and/or revert this and look for another solution if it doesn't).


grant-cleary:
I just checked out the contents of this directory from the commit immediately prior to our 5.9 -> 5.10 update, so it's possible/probable that some of the changes are irrelevant. I'm not particularly concerned about that.

Trandrew17 (approved):

PR #512, opened 2021-12-16
Origin/jschlumkoski/insert element updates nov2021

jschlumkoski:


components/plugins/lcs.js
```diff
@@ -94,6 +96,7 @@ export function addElementEditorDialogPlugin() {
 		};
 
 		editor.on('GetContent', function(e) {
+			console.log('Getting');
```
grant-cleary, line 99: I don't think we want this when the code goes live.

components/plugins/lcs.js
```diff
@@ -376,6 +406,10 @@ class ElementEditorDialog extends LitElement {
 	}
 
 	render() {
+		scripts = new LCSTemplates(LitElement).scripts();
```
grant-cleary, line 409: What is using `LitElement` as an argument accomplishing here? I suspect we probably want this to be refactored a bit like so:
```
const lcsTemplates = new LCSTemplates();
const scripts = lcsTemplates.scripts();
...
```
More broadly though, I'm not sure why `LCSTemplates` needs to be a class at this point. I think you could keep your old functions, but have `templates` take the localize function as a parameter as you do here.

components/plugins/lcs.js
```diff
@@ -481,30 +515,36 @@ class ElementEditorDialog extends LitElement {
...
+				${template ? html`<d2l-alert-toast type="critical" data-target="delete" .open="${live(this._deletedItem)}" @d2l-alert-toast-close="${this._closeAlert}">${this.localize('insertElement:deletedItem', { type: template.item ? template.item : '', index: this._deletedItemIndex + 1 })}.</d2l-alert-toast>` : ''}
+				${template ? html`<d2l-alert-toast type="default" data-target="move" .open="${live(this._movedItem)}" @d2l-alert-toast-close="${this._closeAlert}">${this._movedItemUp ? this.localize('insertElement:movedItemUp', { type: template.item ? template.item : '', index: this._movedItemIndex + 1 }) : this.localize('insertElement:movedItemDown', { type: template.item ? template.item : '', index: this._movedItemIndex + 1 })}.</d2l-alert-toast>` : ''}
 			</d2l-dialog>
 		`;
 	}
 
+	updated() {
+		if (this.shadowRoot.querySelector('#editor').shadowRoot.querySelector('dialog') && this.shadowRoot.querySelector('#editor').shadowRoot.querySelector('dialog').style.overflow !== 'hidden') {
+			this.shadowRoot.querySelector('#editor').shadowRoot.querySelector('dialog').style.overflow = 'hidden';
```
grant-cleary, line 544: What's the purpose of this? It's typically not fantastic to peek inside another element's shadowDOM, so I wonder if we can leverage the dialog to do what you need to do differently.

components/plugins/lcs.js
```diff
@@ -874,6 +916,7 @@ class ElementEditorDialog extends LitElement {
 
 			if (propVal === 'null') {
 				if (this._properties[propidx].type === 'htmleditor') {
+					console.log(this._properties[propidx].value);
```
grant-cleary, line 919: Same comment RE: console logging.

components/plugins/lcs.js
jschlumkoski: We definitely don't, forgot to remove this!

components/plugins/lcs.js
jschlumkoski: @grant-cleary This was something I had intended to ask about. Previously we discussed how when the htmleditor is in the dialog window, and you select something so a contextual toolbar, the dialog tag gets horizontal and vertical scrollbars for seemingly no reason. This was the only way I could resolve it in the tool, not sure the process for getting it fixed in the dialog component itself?

components/plugins/lcs.js
jschlumkoski: My apologies, I thought that it needed to be a class 😬. I've rejigged it so that's no longer the case, and more closely reflects the previous structure.

components/plugins/lcs.js
jschlumkoski: Should be removed now!

components/plugins/lcs.js
grant-cleary: Hmm. I might want to dig into what's causing that overflow when the context menu opens, since I would want to explore other options before we start rooting around in the dialog's shadowDOM.

grant-cleary: Turns out this issue crops up whenever the editor loads in a dialog, so my thinking is that we'll want to remove this workaround from this PR and I'll address the underlying issue directly.

grant-cleary (approved):

PR #516, opened 2021-12-16
Add support for compact styles

grant-cleary:
This adds support for specifying a compact style set in the HTML editor and in the inline editor's HTML block. These styles are centrally managed in core, but we want to align the editor experience with that.


htmleditor-mixin.js
```diff
@@ -519,6 +524,12 @@ export const HtmlEditorMixin = superclass => class extends Localizer(LabelledMix
 			heightConfig.min_height = this._getUnitlessHeight(this.height) - 4; // Inline editor has a 2px width border, so subtract that x2 for the overall height
 		}
 
+		const compactStylesConfig = {};
+		if (this.compact) {
+			compactStylesConfig.body_class = 'd2l-html-block-compact';
```
grant-cleary, line 529: This is possibly a bit odd if this is set to `fullpage`, since at that point the class just isn't going to do anything. I don't know that it's worth explicitly omitting it though.

htmleditor-mixin.js
```diff
@@ -545,6 +556,7 @@ export const HtmlEditorMixin = superclass => class extends Localizer(LabelledMix
 			content_css: `${baseImportPath}/tinymce/skins/content/default/content.css`,
 			content_style: this.fullPage ? `${fullPageStyles} ${isfStyles}` : `${contentFragmentStyles} ${isfStyles} ${htmlBlockContentStyles}`,
 			contextmenu: 'image imagetools',
+			deprecation_warnings: false,
```
grant-cleary, line 559: This is unrelated, but I noticed that TinyMCE started spitting deprecation warnings into the console as of 5.9. We don't want/need end users to see these.

Trandrew17 (approved):

PR #528, opened 2021-12-21
Prevent dialog from overflowing when TinyMCE context menu opens

grant-cleary:
When the editor is loaded in a dialog, opening a TinyMCE context menu inside the editor causes the dialog's outer div to overflow, creating scrollbars in both the x and y-directions and resizing the internal content div. This is due to some CSS we override in the editor, which may or may not still be relevant.


htmleditor-mixin.js
```diff
@@ -226,10 +226,6 @@ export const HtmlEditorMixin = superclass => class extends Localizer(LabelledMix
 			:host([_dropdown-open]) .tox-tinymce-aux {
 				z-index: 990; /* needs to be less than dropdowns; can't override always due to tinyMCE dialogs in fullscreen */
 			}
-			.tox.tox-silver-sink.tox-tinymce-aux {
-				position: fixed !important; /* Safari fix */
-				width: 100%;
-			}
```
grant-cleary, line 232: This may still be necessary, but I can't test on Safari right now. Until I can verify this, I'm just opening this PR to keep it top-of-mind over the holidays.

grant-cleary:
I just gave this a run in a SauceLabs VM and all of the TinyMCE dialogs seem to open fine here. I think this is likely safe to get rid of.

dbatiste (approved):

htmleditor-mixin.js
mostlyfabulous: https://github.com/BrightspaceUI/htmleditor/pull/99

PR #548, opened 2022-01-06
Add page attributes to the attributes dialog when fullPage is specified

grant-cleary:
This allows users to specify a document lang/title for fullPage uses of the editor, piggy-backing on the existing Insert Attributes dialog (designs provided by Noah - see US134808 for details). If fullPage is not specified, this extra tab in the dialog is not present.


components/plugins/attributes.js
```diff
@@ -39,32 +41,74 @@ export function addAttributesPlugin() {
...
-					textDirection: currentNode.getAttribute('dir') || ''
-				};
+				dialog.attributes.title = currentNode.getAttribute('title') || '';
+				dialog.attributes.id = currentNode.getAttribute('id') || '';
+				dialog.attributes.className = currentNode.className || '';
+				dialog.attributes.style = currentNode.getAttribute('style') || '';
+				dialog.attributes.textDirection = currentNode.getAttribute('dir') || '';
+			}
+
+			if (component.fullPage) {
```
grant-cleary, line 58: Note: When we roll our own full-page plugin, I want to split this out and handle it more cleanly, if possible. Fetching the document like this from TinyMCE feels... Unfortunate.

components/plugins/attributes.js
```diff
@@ -39,32 +41,74 @@ export function addAttributesPlugin() {
...
+
+						if (e.detail.attributes.pageLanguage && (!langAttr || (langAttr !== e.detail.attributes.pageLanguage))) {
+							editorDocument.documentElement.setAttribute('lang', e.detail.attributes.pageLanguage);
+							hasChanges = true;
+						} else if (langAttr) {
+							editorDocument.documentElement.removeAttribute('lang');
+							hasChanges = true;
+						}
+
+						if (hasChanges) editor.setContent(editorDocument.documentElement.outerHTML);
```
grant-cleary, line 110: Setting content this way is expensive because it triggers all of our setContent handlers (as well as all of TinyMCE's internal handlers), so I don't want to fire this unless absolutely necessary. This is mostly why the above logic is as complicated as it is.

components/plugins/attributes.js
```diff
@@ -176,6 +190,17 @@ class AttributesDialog extends Localizer(RtlMixin(LitElement)) {
...
 			attributes.style = this.shadowRoot.querySelector('#d2l-attributes-style').value;
 			attributes.textDirection = this.shadowRoot.querySelector('#d2l-attributes-direction-select').value;
+
+			if (this.includePageAttributes) {
+				attributes.pageLanguage = this.shadowRoot.querySelector('#d2l-attributes-page-language').value;
+				attributes.pageTitle = this.shadowRoot.querySelector('#d2l-attributes-page-title').value;
+			}
+		}
+
+		// Reset tab selection when dialog is closed
```
grant-cleary, line 200: This is necessary because we re-use this dialog. Otherwise, if a user saves while on the Page Attributes tab and then re-opens the dialog, the dialog will size to the (much smaller) Page Attributes tab and things look pretty gross when switching tabs.

components/plugins/attributes.js
```diff
@@ -39,32 +41,74 @@ export function addAttributesPlugin() {
...
+				dialog.attributes.className = currentNode.className || '';
+				dialog.attributes.style = currentNode.getAttribute('style') || '';
+				dialog.attributes.textDirection = currentNode.getAttribute('dir') || '';
+			}
+
+			if (component.fullPage) {
+				const editorDocument = new DOMParser().parseFromString(editor.getContent(), 'text/html');
+				const titleElem = editorDocument.head.querySelector('title');
+
+				dialog.attributes.pageTitle = (titleElem && titleElem.textContent) || '';
```
dbatiste, line 62: I'm sorta confused by the difference between the attributes title and the page title.

dbatiste: Is the attributes title meant to be applied to any element, whereas the page title is for the `<title>` element?

grant-cleary: That's right, yeah.

components/plugins/attributes.js
```diff
@@ -39,32 +41,74 @@ export function addAttributesPlugin() {
...
+								editorDocument.head.appendChild(newTitle);
+								hasChanges = true;
+							}
+						} else if (titleElem) {
+							titleElem.remove();
+							hasChanges = true;
+						}
+
+						if (e.detail.attributes.pageLanguage && (!langAttr || (langAttr !== e.detail.attributes.pageLanguage))) {
+							editorDocument.documentElement.setAttribute('lang', e.detail.attributes.pageLanguage);
```
dbatiste, line 103: Hmmm, this is interesting.  As we begin to allow replacements in the user-authored HTML, such as with @mentions, the document language will become important for translations, and our widgets aren't necessarily going to support _any_ language the user might enter in the dialog.

grant-cleary: Yeah, so we're sort of caught between a rock and a hard place there. Because we can't bank on user-authored HTML using one of our supported languages, that also means we can't rely on that for stuff like @mentions... Or at least we have to be careful how we handle what happens if a user enters an unsupported language in there (or complete junk for that matter).

dbatiste: Do we really need to include this, or are we just filling the shopping cart?

grant-cleary: The problem is that consumers can choose whether the source code editor is available, and absent that, there's no other way to specify a `lang` attribute at the document level in full-page mode. I don't know if that's technically a WCAG violation, but it definitely runs afoul of a11y best practices at least.

dbatiste: To close the loop on this - authors require the ability to control the document language for fullpage documents - for example, a Russian course even though we do not support Russian lang.  Worth mentioning, document language should be specified for: "WCAG Success Criterion 3.1.1 requires that a page language is specified in a way which may be 'programmatically determined' (i.e. via the lang attribute)."  

Replacements that render components, as can be expected for @mentions, perhaps the media player, and code samples will require a valid lang that we support for successful translation, and that lang should be consistent with the rest of the UI, not necessarily the lang of the authored document.

Currently, component translation is done based on the document lang. When the time comes to support these component replacements in full-page user-authored HTML where translations are required, we will need to find a way to resolve the lang for localization, perhaps by enabling the localization mixins to honour `lang` on the web component as an override. Such a  solution would require the replacement to render this attribute, and the localization library to look for it and resolve translations accordingly. The parent UI context would also need to provide this lang either as an attribute on the frame, or via some postMessage API. Open to other ideas, but this is outside the scope of this change.

dbatiste (approved):
I know this might not be the easiest thing to test, but ideally we can launch the dialog and test that the values are properly scraped and updated. Ideally we have that sort of test for all the dialogs, even if it's just the most basic permutation.

grant-cleary:
> I know this might not be the easiest thing to test, but ideally we can launch the dialog and test that the values are properly scraped and updated. Ideally we have that sort of test for all the dialogs, even if it's just the most basic permutation.

Good thought. I think it should be doable. The test can fire the command that opens the dialog and scrape the resulting HTML. We already set IDs on those fields so selecting them after the fact should be reasonable.

Trandrew17 (approved):

test/htmleditor.test.js
```diff
@@ -147,4 +148,82 @@ describe('d2l-htmleditor', () => {
...
+					fullPage: true
+				});
+				const attributesDialog = await openAttributesDialog(el);
+
+				setTextInputVal(attributesDialog, 'd2l-attributes-page-title', testTitle);
+				setTextInputVal(attributesDialog, 'd2l-attributes-page-language', testLangAttribute);
+
+				await closeAttributesDialog(attributesDialog, false);
+
+				requestAnimationFrame(() => {
```
grant-cleary, line 205: I actually don't think this is necessary. I'll remove it and the wait below as well. These are relics of before I made better use of `oneEvent`.

PR #555, opened 2022-01-11
Consume TinyMCE from the Brightspace CDN instead of NPM

grant-cleary:
To make transitioning to Lit 2 easier (and for a few other reasons), we want to stop consuming TinyMCE via NPM and statically host it on the Brightspace CDN. This PR makes the editor adjustments necessary to support this.


LICENSE
```diff
@@ -0,0 +1,201 @@
+Apache License
```
grant-cleary, line 1: The htmleditor now hosts only our own code.

README.md
```diff
@@ -203,7 +203,11 @@ npm install @brightspace-ui/htmleditor
 
 ## Developing, Testing and Contributing
 
-After cloning the repo, run `npm install` to install dependencies, `npm run build` to extract the supported tinyMCE languages.
+After cloning the repo, run `npm install` to install dependencies, `npm run build` to extract the supported tinyMCE languages and generate imports required to consume tinyMCE from the CDN.
```
grant-cleary, line 206: Ick, I need to adjust this. Forgot that we're not generating languages anymore.

cli/tinymce-import-generator.js
```diff
@@ -0,0 +1,49 @@
...
+	// eslint-disable-next-line prefer-template
+	const output = '// auto-generated\n'
+		+ `import { tinymceLangs as importedLangs } from '${baseLibPath}/langs/langs.js';\n`
+		+ '\n'
+		+ 'export const tinymceLangs = importedLangs;\n'
+		+ 'export const tinymcePaths = {\n'
+		+ 	`\tbaseImport: '${baseLibPath}',\n`
+		+ 	`\tcoreEditor: '${baseLibPath}/tinymce/tinymce.min.js',\n`
+		+ 	`\tskin: '${baseLibPath}/skins/d2l'\n`
+		+ '};\n';
```
grant-cleary, line 27: This is basically done to make sure we only ever need to update the TinyMCE version in one place within this repo. It also allows us to import (and then re-export) the lang mappings from the CDN statically, which is nice.

components/toolbar/toolbar-full.js
```diff
@@ -1,30 +1,9 @@
...
-import 'tinymce/plugins/textpattern/plugin.js';
-import 'tinymce/themes/silver/theme.js';
-
 import '@brightspace-ui/core/components/dropdown/dropdown-menu.js';
 import '@brightspace-ui/core/components/menu/menu.js';
 import './button.js';
 import './button-menu.js';
 import './button-toggle.js';
 import './menu-item.js';
-import '../plugins/emoji.js';
```
grant-cleary, line 27: The emoticons plugin was updated to have an open/close command, so we don't need our forked version anymore.

components/toolbar/toolbar-full.js
```diff
@@ -59,18 +38,6 @@ export function getTagName() {
 	return tagName;
 }
 
-export function getExternalPlugins(component) {
```
grant-cleary, line 62: All standard plugins are now loaded from the CDN as well. We no longer need external plugins.

htmleditor-mixin.js
```diff
@@ -41,6 +41,14 @@ if (!tinymceLangs.includes(documentLang)) {
 	}
 }
 
+const tinymceLoaded = new Promise(resolve => {
+	const script = document.createElement('script');
+	script.async = 'async';
+	script.onload = resolve;
+	script.src = tinymcePaths.coreEditor;
+	document.head.appendChild(script);
+});
```
grant-cleary, line 50: I'm not super sure about this? I think we do want it to be a promise so we don't try to init TinyMCE before this is done (stuff _really_ explodes if we do). I'm less sure about having it be top-level though. My thinking is that we want to load it ASAP, but I struggled to think of the best way to accomplish that.

htmleditor-mixin.js
```diff
@@ -559,8 +559,8 @@ export const HtmlEditorMixin = superclass => class extends Localizer(LabelledMix
 			deprecation_warnings: false,
 			directionality: this.dir ? this.dir : 'ltr',
 			elementpath: false,
+			emoticons_database_url: `${pathFromUrl(import.meta.url)}/components/plugins/emojis.js`,
```
grant-cleary, line 562: We do still need our custom emoticons database.

package.json
```diff
@@ -10,20 +10,17 @@
   "files": [
     "custom-elements.json",
     "htmleditor.js",
-    "htmleditor.serge.json",
```
grant-cleary, line 13: I have no idea why I put this in here. There's no reason to publish the serge config.

components/plugins/lcs.js
```diff
@@ -10,7 +10,6 @@ import '@brightspace-ui/core/components/list/list.js';
 import '@brightspace-ui/core/components/list/list-item-button.js';
 import '@brightspace-ui/core/components/validation/validation-custom.js';
 import '../../htmleditor.js';
-import 'tinymce/tinymce.js';
```
grant-cleary, line 13: I'm not 100% sure why this was referenced here, but I'm reasonably certain we don't want plugins referencing TinyMCE's init code directly.

tinymce-lib-version.js
```diff
@@ -0,0 +1 @@
+export const version = '0.2.0';
```
grant-cleary, line 1: Surprisingly, this isn't starting at `0.2.0` because I screwed up `0.1.0`! Just accidentally published `0.2.0` as the first release. 😄

grant-cleary:
A thing I'm not sure of: Will this explode BSI when this releases? I assume rollup ignores static files that just don't exist, so I _think_ the answer is no. But if anyone knows differently, let me know. If so, it will just be a case of updating BSI to remove the editor's static file references at roughly the same time this merges.

htmleditor-mixin.js
dlockhart: The downside here is that the act of including the `htmleditor-mixin.js` file will fetch the editor, vs. actually rendering a `<d2l-htmleditor>` on the page somewhere. Loading this dynamically at render time though would mean having your `render()` be async -- which is doable, but might slow things down. You also want to be able to dedupe that loading so only the first editor does the loading.

package.json
dlockhart: I don't think this is set up for Oslo in BSI, but if it were that would be a reason to include this file.

package.json
```diff
@@ -66,7 +63,6 @@
     "@brightspace-ui/core": "^1.220",
     "@brightspace-ui/intl": "^3",
     "lit-element": "^2",
-    "resize-observer-polyfill": "^1",
-    "tinymce": "5.10.2"
```
dlockhart, line 70: 🙌

dlockhart (approved):
As an outsider looking in, this seems like what I'd expect!

package.json
grant-cleary: | I don't think this is set up for Oslo in BSI, but if it were that would be a reason to include this file.

It totally is and I set that up myself. That would be why. 🤦

package.json
grant-cleary: > I don't think this is set up for Oslo in BSI, but if it were that would be a reason to include this file.

It totally is and I set that up myself. That would be why. 🤦

htmleditor-mixin.js
grant-cleary: That's true actually. I was thinking it was a no-op because the toolbar used to import TinyMCE (https://github.com/BrightspaceUI/htmleditor/blob/main/components/toolbar/toolbar-full.js#L1) but we also import the toolbar dynamically so I don't think we're loading TinyMCE immediately regardless.

package.json
dlockhart: Haha I even scanned the `.serge-mappings.json` file and didn't see it. But it's there.

grant-cleary: Yeah I guess I forgot that exporting the serge config was one of the required steps for setting up Oslo. I'll add it back. :P

htmleditor-mixin.js
grant-cleary: So prior, we pulled in TinyMCE on `firstUpdated` for the full editor, or on-demand for the inline editors when someone clicked into the editor. My feeling is that we probably want to keep that behaviour, but I'm curious what @dbatiste thinks as well.

Certainly for the inline cases there were some serious perf concerns with having everything load on page load... But that cost was mostly from rendering the toolbar itself (figuring out sizing, what buttons to render, etc.) and from the actual editor init process. I think TinyMCE's global scripts run relatively quickly in comparison.

grant-cleary: My first go at this did have the script being added to the document in the mixin's `_initializeTinyMCE` method. There are minor changes needed to support that; plugin loading needs to be deferred because otherwise the toolbar code tries to attach plugins to the TinyMCE instance before it's created, and we need to temporarily hide the `textarea` TinyMCE attaches to, otherwise we get visual pop-in (I'm sort of thinking maybe it would be ideal to skeletize the editor during load regardless, though that's perhaps a bigger question). It's doable for sure.

grant-cleary: Mulling over this a bit more, I shifted TinyMCE's script loading back to init, with some caching to make sure we don't try to duplicate things.

htmleditor-mixin.js
```diff
@@ -787,8 +802,12 @@ export const HtmlEditorMixin = superclass => class extends Localizer(LabelledMix
 			'd2l-htmleditor-no-tinymce': !isShadowDOMSupported
 		};
 
+		const textAreaStyles = {
+			visibility: 'hidden'
```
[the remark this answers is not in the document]
grant-cleary: Not to mention if this goes wrong it's blindingly obvious, as the editor iframe is just invisible and no content shows.

htmleditor-mixin.js
```diff
@@ -612,7 +613,7 @@ export const HtmlEditorMixin = superclass => class extends Localizer(LabelledMix
 			},
 			mentions_selector: 'span[data-mentions-id]',
 			object_resizing: true,
-			plugins: plugins.join(' '),
+			plugins: getPlugins(this).join(' '),
```
grant-cleary, line 616: Plugin loading needs to be deferred until after the TinyMCE scripts are loaded. This was formerly a given because TinyMCE's scripts were loaded on toolbar import, but shifting the script loading into the initialization delays that a bit.

htmleditor-mixin.js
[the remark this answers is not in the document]
grant-cleary: As for why we want to hide this... Loading TinyMCE's scripts slightly later means the textarea (and any HTML contents) are visible for slightly longer. This was technically visible for a bit before as well, but it was such a brief time that most folks wouldn't notice. It's now noticeable even on my comparatively fast dev machine with the slightly delayed loading.

components/plugins/lcs.js
dbatiste: I might be because this plugin technically depends on `tinymce` - i.e. `tinymce.PluginManager.add(...)`.  Having said that, this shouldn't actually run until tinymce is loaded, correct?

grant-cleary: That's right. All of our plugins rely on the implicit assumption that the global TinyMCE scripts have run. Importing `tinymce.js` basically does the same thing as adding it as a script to the page... It creates a bunch of globals and such (basically `window.tinymce` and everything it exposes).

htmleditor-mixin.js
```diff
@@ -152,6 +155,11 @@ export const HtmlEditorMixin = superclass => class extends Localizer(LabelledMix
 			 * @type {string}
 			 */
 			html: { type: String },
+			/**
+			 * Whether to enable the insert element LCS plugin
+			 * @type {boolean}
+			 */
+			insertElement: { type: Boolean, attribute: 'insert-element' },
```
dbatiste, line 162: Should this have been introduced with Justin's changes?

dbatiste: Oh I see, it was moved from `html-editor.js` into the mixin.

htmleditor.js
```diff
@@ -324,7 +318,7 @@ class HtmlEditor extends HtmlEditorMixin(SkeletonMixin(LitElement)) {
 			this.requestUpdate();
 			await this.updateComplete;
 
-			this._initializeTinymce(this._toolbar.getPlugins(this), this._toolbar.getExternalPlugins(this), this._toolbar.getTagName, false);
```
dbatiste, line 327: Does the disabled case still work ok?  Just wondering if moving the loading of tinyMCE affects it at all, given that it would no longer be loaded.

grant-cleary: It does work, yeah, both mixed with other editors on the page and as the only editor in the page. In the disabled case, we don't load the toolbar or try to initialize TinyMCE at all until after the editor is enabled, so in that case the workflow is more or less the same as inline mode.

dbatiste (approved):

PR #570, opened 2022-01-19
Add license checker to PR workflow

grant-cleary:
Does what it says on the tin; now that TinyMCE isn't referenced by the repo, we can add this workflow.


grant-cleary:
The new job passed, so I'm just going to merge this up. Note that this will be mandatory for `main`, but it can't be mandatory for release branches until we're only building 1.40+. Runs against 1.39.x or earlier will fail.

PR #574, opened 2022-01-21
Prepare backwards compatible changes for Lit 2 migration

grant-cleary:
This PR makes backwards-compatible changes to prepare this repo for a Lit 2 upgrade. Since the changes are backwards compatible, it can be merged right away.

Non-breaking change checklist:
- [ ] ~Add missing `.js` extensions for lit-html or lit-element imports~ Already completed
- [ ] ~`await requestUpdate` -> `await elem.updateComplete`~ Already completed
- [x] add non-null checks to shadowRoot accesses (not including test files)
- [x] ensure `undefined` is not passed into unsafeHTML calls
- [ ] ~remove dynamic code from inside  `<template>` tags (Lit files only)~ Already completed
- [ ] ~remove any dependencies on old Lit versions~ Only Lit-dependent deps are core and @open-wc stuff
- [x] audit `ifDefined` usages for null parameters

For more information on the Lit 2 upgrade, visit https://lit.dev/docs/releases/upgrade/ or contact team Gaudi.


components/toolbar/button-menu.js
```diff
@@ -152,11 +153,13 @@ class ButtonMenu extends DropdownOpenerMixin(ToolbarItemMixin(FocusVisiblePolyfi
 	}
 
 	focus() {
+		if (!this.shadowRoot) return;
```
grant-cleary, line 156: In general I opted to add this check to any public APIs, just in case some consumer calls them post-construction but pre-render (somehow). I think that's unlikely to occur, but :shrug:.

grant-cleary:
I'll also prep the breaking changes PR, but the package-lock is going to be broken until core updates.

package.json
```diff
@@ -64,6 +64,7 @@
     "@brightspace-ui/core": "^1.224",
     "@brightspace-ui/intl": "^3",
     "lit-element": "^2",
+    "lit-html": "^1",
```
annabelsegalld2l, line 67: This happens in a lot of repos, actually. Lit-element requires lit-html so _technically_ you don't need the explicit import.

grant-cleary: Haha fair enough.

annabelsegalld2l (approved):
LGTM

PR #592, opened 2022-02-02
Make sure APIs don't try to reference tinymce if global isn't set

grant-cleary:
With TinyMCE now served via the CDN, there are some timing differences in terms of when the editor loads. Specifically, depending on timing, a consumer can call the editor's public APIs either before TinyMCE has loaded or after the editor has been disconnected from the DOM, both of which can cause problems. Hence, for the APIs that expect TinyMCE to be present, we should just no-op them or proceed with any non-TinyMCE actions and ignore the rest.


htmleditor-mixin.js
```diff
@@ -386,16 +386,16 @@ export const HtmlEditorMixin = superclass => class extends Localizer(LabelledMix
 	}
 
 	clear() {
-		const editor = tinymce.EditorManager.get(this._editorId);
+		this.html = '';
```
grant-cleary, line 389: The `html` setter already checks for `window.tinymce` before deciding how to handle input, so we can actually call this safely here without verifying the editor's existence first (in that case, we just set `this._html` accordingly). Of course, we do need to verify the editor's existence before trying to clear its undoManager.

htmleditor-mixin.js
```diff
@@ -386,16 +386,16 @@ export const HtmlEditorMixin = superclass => class extends Localizer(LabelledMix
...
 		if (!editor) return;
 
-		this.html = '';
 		editor.undoManager.clear();
 	}
 
 	async focus() {
-		await this.initializationComplete;
-		const editor = tinymce.EditorManager.get(this._editorId);
+		const editor = await this._getEditor();
```
grant-cleary, line 398: I'm pretty sure this is what's breaking the tests right now (all of those validation tests focus the editor)... But I'm slightly confused as to why, because `this._getEditor()` literally just does what the removed code did. 😕

htmleditor.js
```diff
@@ -361,14 +361,14 @@ class HtmlEditor extends HtmlEditorMixin(SkeletonMixin(LitElement)) {
...
-	focus() {
+	async focus() {
 		if (!this.shadowRoot) return;
 
 		if ((this.type === editorTypes.INLINE || this.type === editorTypes.INLINE_LIMITED) && !this._editing) {
 			const inlineButton = this.shadowRoot.querySelector('.d2l-htmleditor-inline-button');
 			inlineButton.focus();
 		} else {
-			super.focus();
+			await super.focus();
```
grant-cleary, line 371: This is why the tests were failing. The mixin's focus API is async, but this one wasn't. I have no idea why changing the mixin's focus API changed this behaviour though. =/

dbatiste: Do consumers of this focus method need to wait?  Otherwise I don't see why adding this does anything different.

grant-cleary: Depends. If a consumer calls this while TinyMCE is currently initializing and then try to do something assuming the editor is focused and ready for input, yes. This is specifically a problem in the tests because they try to focus the editor and then blur it. I'm a bit on the fence about doing this explicitly for tests, since I'm dubious that any consumers would need to programmatically focus and then do some other thing.

dbatiste: Do the tests already wait... I don't see any updates for them.

Yeah I think this is the reason the public focus method was not already `async`.

grant-cleary: The tests already `await`ed, yeah... It just wasn't actually awaiting anything. I think the timing just happened to work out such that TinyMCE would init and focus would apply before the blur happened.

grant-cleary: I'm going to test an alternative option though. Technically the editor fires a focus event when it's focused (surprise surprise)... It might happen too early for it to be useful for the tests, but it might delay things long enough for the tests to pass. I'm nervous about having the focus API be async in case it throws off timings in other apps.

dbatiste (approved):

PR #593, opened 2022-02-02
Reference emoji DB from the CDN

grant-cleary:
Now that we serve up the emoji DB from the CDN, reference it as such from the editor to avoid potential issues with Rollup builds.


htmleditor-mixin.js
```diff
@@ -561,7 +560,7 @@ export const HtmlEditorMixin = superclass => class extends Localizer(LabelledMix
 			deprecation_warnings: false,
 			directionality: this.dir ? this.dir : 'ltr',
 			elementpath: false,
-			emoticons_database_url: `${pathFromUrl(import.meta.url)}/components/plugins/emojis.js`,
+			emoticons_database_url: `${baseImportPath}/emoji/emojis.js`,
```
grant-cleary, line 563: Whoops. I explicitly set a path for this. 🤦

Trandrew17 (approved):

PR #598, opened 2022-02-07
Implement custom advanced source code editor

grant-cleary:
This is a shot at writing our own source code editor. It uses CodeMirror 6 under the hood and adds a bit of custom styling to look more like one of our typical input components.


components/plugins/source-code.js
```diff
@@ -1,5 +1,14 @@
+import '@brightspace-ui/core/components/colors/colors.js';
+import '@brightspace-ui/core/components/dialog/dialog.js';
+import { css, html, LitElement } from 'lit-element/lit-element.js';
+import { EditorState, StateEffect } from '@codemirror/state';
```
grant-cleary, line 4: I should check how large the state and view packages are. These could be dynamically imported as well; it would just be a bit harder because the imports would need to be tracked throughout the dialog.

components/plugins/source-code.js
```diff
@@ -27,5 +55,255 @@ export function addSourceCodePlugin() {
...
+	_handleOpen() {
+		if (this._codeEditor) {
+			this._codeEditor.focus();
+			this._updateCodeEditorTooltipPositioning();
+			return;
+		}
+
+		const editorContainer = this.shadowRoot.querySelector('.d2l-htmleditor-sourcecode-container');
+		editorContainer.dispatchEvent(new AsyncStateEvent(
+			this._renderCodeEditor(editorContainer)
```
grant-cleary, line 157: Because we're asynchronously loading packages into CodeMirror, we need the dialog to load asynchronously. We're in a bit of a weird situation here because we're modifying the DOM directly and not handling this async behaviour via Lit, so we can't rely on the `runAsync` directive to do this for us. Instead, we have to hook into the dialog open event to make sure we're firing the async state event at the right time. Otherwise, we run into a race condition in which we can fire the event before the dialog contents are slotted into the dialog's shadowDOM.

components/plugins/source-code.js
```diff
@@ -27,5 +55,255 @@ export function addSourceCodePlugin() {
...
+		const fold = await import('@codemirror/fold');
+		const gutter = await import('@codemirror/gutter');
+		const search = await import('@codemirror/search');
+		const history = await import('@codemirror/history');
+		const comment = await import('@codemirror/comment');
+		const highlight = await import('@codemirror/highlight');
+		const commands = await import('@codemirror/commands');
+		const language = await import('@codemirror/language');
+		const rectangularSelection = await import('@codemirror/rectangular-selection');
+		const htmlLang = await import('@codemirror/lang-html');
```
grant-cleary, line 181: CodeMirror's strategy is to break everything up into smaller packages that can be imported as needed in order to keep the base editor `state` and `view` packages as small as possible. I'm not sure I agree with keeping `state` and `view` separate since they're both effectively required to create a minimal working editor, but it is what it is. Either way, because of this strategy, we likely want to dynamically load these to avoid unnecessary loading glut when the editor first initializes.

components/plugins/source-code.js
```diff
@@ -27,5 +55,255 @@ export function addSourceCodePlugin() {
...
+			'next': `${this.localize('source:editor:nextMatch')}`,
+			'previous': `${this.localize('source:editor:previousMatch')}`,
+			'all': `${this.localize('source:editor:findAll')}`,
+			'match case': `${this.localize('source:editor:matchCase')}`,
+			'replace': `${this.localize('source:editor:replace')}`,
+			'replace all': `${this.localize('source:editor:replaceAll')}`,
+			'close': `${this.localize('source:editor:closeSearch')}`,
+			'current match': `${this.localize('source:editor:currentMatch')}`,
+			'on line': `${this.localize('source:editor:currentMatchLineNumber')}`
+		};
```
grant-cleary, line 208: This is not ideal, but it's how CodeMirror's localization strategy currently works. All we can do is map static snippets to lang terms and this isn't super well documented so we may experience drift over time. 😢

components/plugins/source-code.js
```diff
@@ -27,5 +55,255 @@ export function addSourceCodePlugin() {
...
+		const editorTheme = EditorView.theme({
+			'.cm-tooltip-autocomplete': {
+				'margin-top': `${boundingContainer.offsetTop - boundingRect.top}px`,
+				'margin-left': `${leftOffset}px`
+			}
+		});
+
+		this._codeEditor.dispatch({
+			effects: StateEffect.appendConfig.of(editorTheme)
+		});
```
grant-cleary, line 305: This handles the initial positioning of the tooltip absent any resizing events. Once the tooltip has rendered for the first time, we don't need/want to do this again.

components/plugins/source-code.js
```diff
@@ -27,5 +55,255 @@ export function addSourceCodePlugin() {
...
+		if (!boundingContainer || boundingContainer === document.documentElement) return;
+
+		const boundingRect = boundingContainer.getBoundingClientRect();
+		const leftOffset = boundingContainer.offsetLeft - boundingRect.left;
+
+		const tooltip = this.shadowRoot.querySelector('.cm-tooltip-autocomplete');
+		if (tooltip) {
+			tooltip.style.marginLeft = `${leftOffset}px`;
+			return;
+		}
```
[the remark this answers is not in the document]
grant-cleary: There's supposed to be a way to tie this into CodeMirror's state dispatch framework such that we could just dynamically update the CSS attached to the theme, but I wasn't able to find a way to do it that didn't require effectively unloading and reloading all of the extensions, which is more expensive than I would like.

components/plugins/source-code.js
```diff
@@ -27,5 +55,255 @@ export function addSourceCodePlugin() {
...
+			this._renderCodeEditor(editorContainer)
+		));
+	}
+
+	_handleResize() {
+		clearTimeout(this._resizeTimerId);
+
+		this._resizeTimerId = setTimeout(() => {
+			this._updateCodeEditorTooltipPositioning();
+		}, 100);
```
grant-cleary, line 166: Basic debouncing. 100ms was an arbitrary choice that looks more or less okay and doesn't set fire to Chrome while resizing, but I'm open to other options.

components/plugins/source-code.js
```diff
@@ -16,8 +25,27 @@ export function addSourceCodePlugin() {
...
+		const openSourceCodeEditor = () => {
+			let dialog = component.shadowRoot.querySelector('d2l-htmleditor-sourcecode-dialog');
+			if (!dialog) dialog = component.shadowRoot.appendChild(document.createElement('d2l-htmleditor-sourcecode-dialog'));
+
+			dialog.opened = true;
+			component._dialogOpen = true;
+			dialog.html = editor.getContent();
+
+			dialog.addEventListener('d2l-htmleditor-sourcecode-dialog-close', (e) => {
+				if (e.detail !== undefined && e.detail !== null) editor.setContent(e.detail);
```
grant-cleary, line 37: This can't just be an `if (e.detail)` check because `''` is an entirely valid output from the source code editor.

components/plugins/source-code.js
```diff
@@ -27,5 +55,255 @@ export function addSourceCodePlugin() {
...
+			}
+		));
+
+		this.html = '';
+		this.opened = false;
+	}
+
+	_handleOpen() {
+		if (this._codeEditor) {
+			this._codeEditor.focus();
```
grant-cleary, line 150: I might be able to ditch the explicit focus now that the dialog contents are loading asynchronously, but I'll need to verify that. The reason this exists is because the dialog's cancel button was being focused when it opened, which is not what we want.

components/plugins/source-code.js
grant-cleary: Ok, they're both quite small (~180 bytes each) and only have two dependencies between them, each of which are of a similar size. I'll do a quick pass to see if it's trivial to defer loading them, but if it's not, I think we don't need to worry too much about the cost here.

grant-cleary: Eh, I think it over-complicates things and probably isn't worth it for what amounts to less than half a kB of additional overhead. I'm open to other opinions of course.

components/plugins/source-code.js
grant-cleary: These focus checks are still required - otherwise the dialog still assigns focus to the save button and skips over the form entirely. It's a bit weird given that it loads asynchronously though, since focus does then jump from the save button to the editor once the dialog contents finish loading. Unsure if there's an alternative way to handle that, though CodeMirror does set up an aria-live region that I think does help mitigate any confusion there.

components/plugins/source-code.js
dbatiste: What is the purpose of the "state" and "view" packages?

grant-cleary: They contain the fundamental APIs required to interact with the CodeMirror editor. `EditorState` and `StateEffect` are static classes used to create and dispatch changes to the editor, while `EditorView` is used to create/get/modify the editor object itself.

components/plugins/source-code.js
```diff
@@ -27,5 +55,256 @@ export function addSourceCodePlugin() {
...
+			}
+		`;
+	}
+
+	constructor() {
+		super();
+		this.html = '';
+		this.opened = false;
+		this._resizeTimerId = 0;
+		this._handleResize = this._resize.bind(this);
```
dbatiste, line 91: This doesn't look right - based on the current method name, it should be...
```javascript
this._handleResize = this._handleResize.bind(this);
```
Currently, there is a JS error trying to launch it, due to this.

grant-cleary: Ah crap, you're right. Last minute copy/paste error.

htmleditor-mixin.js
```diff
@@ -399,12 +399,12 @@ export const HtmlEditorMixin = superclass => class extends Localizer(LabelledMix
...
 	}
 
-	_convertHtml(html, sourceView, converters) {
+	_convertHtml(html, converters) {
 		const tempDoc = new DOMParser().parseFromString(html, 'text/html');
 		if (!tempDoc.body) return html;
 
 		converters.forEach(convert => convert(tempDoc));
-		if (!this.fullPage || sourceView || !/<head[^>]*>.*<\/head>/s.test(html)) return tempDoc.body.innerHTML;
+		if (!this.fullPage || !/<head[^>]*>.*<\/head>/s.test(html)) return tempDoc.body.innerHTML;
```
grant-cleary, line 407: This exclusion was originally necessary because TinyMCE's fullpage plugin does some magic to inject the proper document markup into the editor when you execute the `mceCodeEditor` command. Thus we needed to make sure we weren't doing that twice. If we're not using TinyMCE's code editor, we don't need this exclusion, and this parameter would never return true anyways.

Trandrew17 (approved):

PR #611, opened 2022-02-15
Propagate host correctly when set within an FRA

grant-cleary:
When switching over to use the context controller, I mistakenly made the host unavailable within the core editor and its mixin. This leads to URLs being referenced against the CDN when loaded in FRAs.


htmleditor-mixin.js
```diff
@@ -353,8 +354,8 @@ export const HtmlEditorMixin = superclass => class extends Localizer(LabelledMix
 		this._context = getContext(this._contextController);
 
 		this._fraContext = await isFramed();
-		const host = await getHost();
-		this.provideInstance('context', { ...this._context, ...{ host: host, component: this } });
+		this._host = await getHost();
+		this.provideInstance('context', { ...this._context, ...{ host: this._host, component: this } });
```
grant-cleary, line 358: Before the context controller came into play, `host` was a property on the `_context` object... But because the context object is now being populated from the controller and not directly from the editor, I separated `host` out to be its own thing. Unfortunately, I forgot to actually set it in a way that these files could access. 😬

htmleditor-mixin.js
```diff
@@ -525,8 +526,8 @@ export const HtmlEditorMixin = superclass => class extends Localizer(LabelledMix
 		};
 
 		const documentConfig = {};
-		if (this._context && this._context.host) {
```
grant-cleary, line 528: `this._context.host` was always `undefined`.

ianlaidlaw (approved):

PR #620, opened 2022-02-18
Add positioning logic to inline editor to avoid it moving out of frame before the editor does

grant-cleary:
This is a shot at a more elegant solution to make sure the inline editor toolbar doesn't fall out of the viewport or get obscured by a bounding ancestor while the editor itself is still visible. If we run out of room on the left side to render the toolbar, we simply anchor the toolbar to the left corner of the editor and extend it to the right instead.

If the editor is partially obscured by another element or out of the viewport, the toolbar still will be as well, but at that point we don't really have any sensible solutions anyways.


components/toolbar/toolbar-mixin.js
```diff
@@ -205,6 +221,7 @@ export const ToolbarMixin = superclass => class extends Localizer(RtlMixin(super
 			container.addEventListener('d2l-htmleditor-toolbar-item-resize', () => {
 				this._updateMeasures();
 				this._updateItemsVisibility(false);
+				this._position();
```
grant-cleary, line 224: Makes sure that we set the initial position properly if the toolbar first renders while it would normally be offscreen or obscured.

components/toolbar/toolbar-mixin.js
```diff
@@ -303,10 +320,38 @@ export const ToolbarMixin = superclass => class extends Localizer(RtlMixin(super
 
 	async _handleResize() { }
 
+	async _position() {
+		if (!this.floating) return;
```
grant-cleary, line 324: If we're not an inline toolbar, get the heck out of here.

components/toolbar/toolbar-mixin.js
```diff
@@ -303,10 +320,38 @@ export const ToolbarMixin = superclass => class extends Localizer(RtlMixin(super
...
+
+		requestAnimationFrame(() => {
+			const offsetParent = getOffsetParent(this);
```
grant-cleary, line 327: I feel like this isn't necessary. The `offsetParent` of the toolbar is always the toolbar container, so I think we can probably just grab it directly. I'll give it a shot.

components/toolbar/toolbar-mixin.js
```diff
@@ -303,10 +320,38 @@ export const ToolbarMixin = superclass => class extends Localizer(RtlMixin(super
...
+
+			let spaceAvailable;
+			if (offsetParent && offsetParent.tagName !== 'BODY') {
+				const boundingContainer = getBoundingAncestor(this);
+
+				let offset = 0;
+				if (boundingContainer !== document.documentElement) {
+					const boundingContainerRect = boundingContainer.getBoundingClientRect();
+					offset = boundingContainerRect.left;
```
grant-cleary, line 336: I think in the general case this might need to be added to `boundingContainer.offsetLeft` in case the container has margins/padding. 🤔

grant-cleary: Er, subtracted from? Whatever the right order of operations is here. I barely remember how these properties work together at the best of times.

components/toolbar/toolbar-mixin.js
```diff
@@ -303,10 +320,38 @@ export const ToolbarMixin = superclass => class extends Localizer(RtlMixin(super
...
+				}
+
+				const parentRect = offsetParent.getBoundingClientRect();
+				spaceAvailable = parentRect.width + parentRect.left + offsetParent.clientLeft - offset;
```
grant-cleary, line 340: The idea here is that we always have at least the full width of the editor as available space, then the width of the editor's border/scrollbar (if applicable). If anything I think I dislike the order of these here, because really `parentRect.left` and `offset` kind of go together as a group. Effectively on top of the space provided by the editor, we also get additional space equivalent to the difference between the editor's position relative to the left edge of the viewport and the left edge of any bounding ancestor that may exist between them.

components/toolbar/toolbar-mixin.js
```diff
@@ -303,10 +320,38 @@ export const ToolbarMixin = superclass => class extends Localizer(RtlMixin(super
...
+			} else {
+				spaceAvailable = -document.documentElement.scrollLeft;
```
grant-cleary, line 342: I think I accidentally removed a term and this is wrong, though I'm not sure why I wasn't seeing this break in the editor's demo. I'm going to take a look.

htmleditor.js
```diff
@@ -53,11 +53,6 @@ class HtmlEditor extends HtmlEditorMixin(SkeletonMixin(LitElement)) {
 			 * @type {boolean}
 			 */
 			disabled: { type: Boolean },
-			/**
-			 * Whether to give the toolbar fixed positioning to ensure it doesn't get cutoff. Applies to editors with type `inline` and `inline-limited`.
-			 * @type {boolean}
-			 */
-			escapeScrollAreas: { type: Boolean, attribute: 'escape-scroll-areas' },
```
grant-cleary, line 60: Rubrics is the only consumer of this property, so once we push this change through, we can safely ditch it.

components/toolbar/toolbar-mixin.js
grant-cleary: Oh, never mind... Silly mistake. We can never hit this `else` block because we know the toolbar container will always exist.

components/toolbar/toolbar-mixin.js
grant-cleary: Eh, it would require passing down a class name, so I'm not super keen on this. I think using `getOffsetParent`.

components/toolbar/toolbar-mixin.js
```diff
@@ -303,10 +320,32 @@ export const ToolbarMixin = superclass => class extends Localizer(RtlMixin(super
 
 	async _handleResize() { }
 
+	async _position() {
```
grant-cleary, line 323: This doesn't need to be async. Originally I had some async stuff happening in here, but that's no longer the case.

htmleditor.js
grant-cleary: Caveat: I haven't tested this with the Polymer version of Rubrics. No idea if that's likely to come back, but if it is, I can't speak to how well this will all work.

components/toolbar/toolbar-mixin.js
```diff
@@ -154,21 +159,32 @@ export const ToolbarMixin = superclass => class extends Localizer(RtlMixin(super
...
-		this._chomping = true;
 		this._state = 'chomping';
 	}
 
 	connectedCallback() {
 		super.connectedCallback();
 		this._chomping = getChomperPreference();
+
+		if (this.floating) {
+			window.addEventListener('resize', this._position.bind(this));
```
dbatiste, line 175: The `bind` method produces a new function, and we need `addEventListener` and `removeEventListener` to reference the same function instances, so the result of `bind()` should be assigned to the class member that is reference elsewhere.  Usually we do this in the constructor, but I think it could be done in this conditional block instead.  Either way, the add/remove calls need to reference the same thing.

```javascript
this._position = this._position.bind(this);
//
window.addEventListener('resize', this._position);
//
window.removeEventListener('resize', this._position);
```

components/toolbar/toolbar-mixin.js
```diff
@@ -303,10 +320,32 @@ export const ToolbarMixin = superclass => class extends Localizer(RtlMixin(super
...
+			const boundingContainer = getBoundingAncestor(this);
+
+			let offset = 0;
+			if (boundingContainer !== document.documentElement) {
+				const boundingContainerRect = boundingContainer.getBoundingClientRect();
+				offset = boundingContainerRect.left + boundingContainer.clientLeft;
+			}
+
+			const parentRect = offsetParent.getBoundingClientRect();
+			const spaceAvailable = parentRect.width + offsetParent.clientLeft + (parentRect.left - offset);
```
dbatiste, line 337: Is the `offsetParent` using `borderBox`?  I am guessing not... since this seems to be calculating for border and whitespace.

components/toolbar/toolbar-mixin.js
dbatiste: I think I would be inclined to add a comment here...
```javascript
// should be toolbar container
const offsetParent = getOffsetParent(this);
```
Since it immensely helps understand the calc.  Or rename the variable.

components/toolbar/toolbar-mixin.js
```diff
@@ -50,6 +50,7 @@ export const ToolbarMixin = superclass => class extends Localizer(RtlMixin(super
 	static get properties() {
 		return {
 			floating: { type: Boolean },
+			_anchorLeft: { type: Boolean, reflect: true },
```
dbatiste, line 53: Looks like this doesn't need to be reflected.

htmleditor.js
MitchelPaulin: The Polymer version will still be around behind as a fail safe so we can pull the plug if anything goes wrong with the LIt version. The hope the Polymer version will be gone for good though

components/toolbar/toolbar-mixin.js
grant-cleary: Ah, you're right. This was a leftover of another approach I tried earlier.

components/toolbar/toolbar-mixin.js
grant-cleary: Ah yeah, I meant to do this in the constructor. But yeah, would be nice if we didn't have defunct event handlers hanging around.

components/toolbar/toolbar-mixin.js
grant-cleary: I just renamed the variable. I think it's reasonable to infer from the variable name that we're expecting the toolbar container to always be the offset parent. I think if we change any of that structure, we're going to break other stuff too, so I think it's a reasonably safe assumption.

MitchelPaulin (approved):

dbatiste (approved):

PR #622, opened 2022-02-18
Disable outer document scrolling when fullscreen is enabled

grant-cleary:
When fullscreen is enabled, we should disable scrolling on the outer document body.


components/plugins/fullscreen.js
```diff
@@ -10,9 +10,19 @@ export function addFullscreenPlugin() {
 		const component = context.component;
 
 		let state = false;
+		let scrollOverflow;
```
grant-cleary, line 13: This could live outside of the function, but two editors can't really fullscreen at once so I don't think it particularly matters. I don't feel strongly about it though.

components/plugins/fullscreen.js
```diff
@@ -10,9 +10,19 @@ export function addFullscreenPlugin() {
...
 
 		const toggleFullscreen = () => {
 			state = !state;
+
+			if (state) {
+				scrollOverflow = document.body.style.overflow;
+				document.body.style.overflow = 'hidden';
```
grant-cleary, line 20: This gets marginally weirder for FRAs, but I figure it's down to those apps to figure out scrolling outside of the iframe. I could do some super complicated stuff like our dialog logic does, but I don't feel it's likely to be worth it.

dbatiste: The [dialog-mixin](https://github.com/BrightspaceUI/core/blob/main/components/dialog/dialog-mixin.js#L328) uses [allowBodyScroll](https://github.com/BrightspaceUI/core/blob/main/components/backdrop/backdrop.js#L137) and [preventBodyScroll](https://github.com/BrightspaceUI/core/blob/main/components/backdrop/backdrop.js#L190) helpers.  If the editor is rendered inside of a dialog, it looks like your method will still preserve the dialog's override, so it looks like either one should work.  What happens when dialogs are launched from the editor while it's in fullscreen mode - I think its logic may override your override here...?  It might be safer to have everything use the helpers... thoughts?

grant-cleary: Hmm, that's a good point. This does appear to be the case, though the scrollbar is hidden behind the backdrop. Still, it looks ugly and I'm happy to use the helpers to keep things consistent regardless.

dbatiste (approved):

PR #645, opened 2022-02-28
Daylight: Label some properties as ADVANCED

margaree:
These properties will be hidden from the main "Properties" table in the Daylight site, but will be visible after clicking a "Show Advanced Properties" button.

This was my best case. Please correct if I missed any or if any of these should be shown by default.


grant-cleary (approved):
LGTM!

PR #697, opened 2022-03-25
fix: add padding to stabilize no-border size

devpow112:
Adds some extra padding to stop size giggle when `no-border` is set and you hover over the editor.

**Before**
beefore

**After**
after


devpow112:
For some reason the cache for this build is corrupt.
```console
/*stdin*\ : Decoding error (36) : Restored data doesn't match checksum 
/usr/bin/tar: Child returned status 1
/usr/bin/tar: Error is not recoverable: exiting now
```

grant-cleary (approved):

grant-cleary:
> For some reason the cache for this build is corrupt.
> 
> ```
> /*stdin*\ : Decoding error (36) : Restored data doesn't match checksum 
> /usr/bin/tar: Child returned status 1
> /usr/bin/tar: Error is not recoverable: exiting now
> ```

Weird. I'm reasonably confident adding one line of CSS isn't going to fail any unit tests, so I'll just force this through.

PR #698, opened 2022-03-28
Revert "fix: add padding to stabilize no-border size"

devpow112:
Reverts BrightspaceUI/htmleditor#697. This causes other style issues with rubrics. Will revert and try and find another fix.


grant-cleary:
Out of curiosity, what issue does this cause? I think this would probably have been my first thought for fixing this (to make sure the overall editor size is maintained in all states).

grant-cleary (approved):

devpow112:
> Out of curiosity, what issue does this cause? I think this would probably have been my first thought for fixing this (to make sure the overall editor size is maintained in all states).

@grant-cleary It seems to remove all the padding in the HTML editor and replaces it with just 2px of padding. I'm trying to see if this is a rubric specific issue so maybe we can hold off on this for now?

devpow112:
Ok so if we just set padding the way I it just overrides this line in `d2l-input`.
https://github.com/BrightspaceUI/core/blob/main/components/inputs/input-styles.js#L53
Which looses all padding. I can make it work by faking a border that counteracts the removal of the `1px` in the padding cals.
```css
:host([no-border]) .d2l-htmleditor-container,
:host([no-border]) .d2l-input {
  border: none;
  box-shadow: none;
}
:host([no-border]) .d2l-input:hover {
  border: 1px solid transparent;
}
```

devpow112:
I'll chuck that up instead of a revert. Abandoning this.

PR #721, opened 2022-04-04
test: add helper tool for mathjax context

devpow112:
We are currently using this is rubrics vdiff, tests and demo pages to auto setup the mathjax context. I'd like to include it with the editor similar to https://github.com/BrightspaceUI/core/blob/main/tools/resize-observer-test-error-handler.js so that we can use it from this component across multiple repo in testing similar to https://github.com/BrightspaceUI/core/blob/main/web-test-runner.config.js#L31-L37.


grant-cleary:
This might be better placed in `core`, since that's where the MathJax stuff lives: https://github.com/BrightspaceUI/core/blob/main/helpers/mathjax.js

I do like the idea though; I was thinking about doing something similar for the editor context. :)

PR #726, opened 2022-04-05
Insert Practice Tool

jschlumkoski:
Initial development of the new Insert Practice Tool built by the Learning Solutions Team.


components/plugins/practice.js
```diff
@@ -0,0 +1,866 @@
...
+import '@brightspace-ui/core/components/menu/menu.js';
+import '@brightspace-ui/core/components/menu/menu-item-checkbox.js';
+import '@brightspace-ui/core/components/validation/validation-custom.js';
+import '@brightspace-ui/core/templates/primary-secondary/primary-secondary.js';
+import '../../components/practice/sequencing.js';
+import '../../components/practice/sorting.js';
+import '../../htmleditor.js';
+
+import { allowBodyScroll, preventBodyScroll } from '@brightspace-ui/core/components/backdrop/backdrop.js';
+import { css, html, LitElement } from 'lit-element/lit-element.js';
```
grant-cleary, line 25: The editor recently finished upgrading to Lit 2, which means we don't import `lit-element` and `lit-html` explicitly anymore. You'll want to switch up these references to reference the `lit` package, like I did in this PR: https://github.com/BrightspaceUI/htmleditor/pull/724

components/plugins/practice.js
```diff
@@ -0,0 +1,866 @@
...
+		const launchPracticeEditor = () => {
+			const context = requestInstance(editor.getElement(), 'context');
+			const component = context.component;
+
+			let dialog = component.shadowRoot.querySelector('d2l-practice-dialog');
+			if (!dialog) dialog = component.shadowRoot.appendChild(document.createElement('d2l-practice-dialog'));
+
+			const contextNode = getSelectedElement();
+			if (contextNode) {
+				dialog._activityDataURL = `${JSON.parse(window.top.document.querySelector('html').dataset.heContext).orgUnitPath}practice/${contextNode.querySelector('iframe').getAttribute('data-file')}`;
```
grant-cleary, line 61: You shouldn't need to query the `heContext` object this way. Everything in there is propagated to the `context` object, so `context.orgUnitPath` should do the trick here.

components/plugins/practice.js
```diff
@@ -0,0 +1,866 @@
...
+			}
+
+			dialog.opened = true;
+			dialog._bodyScrollKey = preventBodyScroll();
+
+			dialog.addEventListener('d2l-practice-dialog-open', () => {
+				editor.editorContainer.getRootNode().querySelector('.tox-tinymce-aux').innerHTML = '';
```
grant-cleary, line 68: I think we chatted about this in the Insert Element PR, but I forget now (lol). Why do we blank out TinyMCE's dialog/toolbar container here?

components/plugins/practice.js
```diff
@@ -0,0 +1,866 @@
...
+				if (!e.detail || !e.detail.data) {
+					editor.editorContainer.getRootNode().querySelector('.tox-tinymce-aux').innerHTML = '';
+					editor.selection.collapse();
+
+					dialog._handleClose();
+
+					return;
+				}
+
+				const heContext = JSON.parse(window.top.document.querySelector('html').dataset.heContext);
```
grant-cleary, line 82: Same as above. `heContext` is the same as the `context` object, so you shouldn't need to fetch it dynamically.

components/plugins/practice.js
jschlumkoski: @grant-cleary This is a sort of workaround to hide TinyMCE's context toolbar when the editor is opened. Without doing this, if you had the context toolbar up for an element and click the edit button, it would appear over the dialog window!

components/plugins/practice.js
jschlumkoski: I'll look to swap this out everywhere it's used!

components/plugins/practice.js
```diff
@@ -0,0 +1,866 @@
...
+
+					if (contextNode) {
+						editor.selection.select(contextNode);
+						editor.insertContent(`${html}`);
+					}
+					else {
+						editor.insertContent(`<p>&nbsp;</p>${html}<p>&nbsp;</p>`);
+					}
+				};
+				vc._uploadCourseContent(heContext.orgUnitId, JSON.stringify(e.detail.data), false, `config-${e.detail.data.id}.practice.json`);
```
grant-cleary, line 108: Same comment w.r.t. `heContext`.

components/plugins/practice.js
jschlumkoski: I did see an example of that, I'll change any lit imports to match the new convention.

components/plugins/practice.js
```diff
@@ -0,0 +1,866 @@
...
+
+			}, { once: true });
+		};
+
+		editor.on('GetContent', function(e) {
+			const editorDocument = new DOMParser().parseFromString(e.content, 'text/html');
+
+			if (editorDocument.querySelectorAll('.d2l-practice').length > 0) {
+				const css = 'https://practices.lcs.brightspace.com/practice/assets/20.22.5/practice.min.css';
+				const js = 'https://practices.lcs.brightspace.com/practice/helpers/20.22.5/events.min.js';
```
grant-cleary, line 118: RE: `20.22.5` - bear in mind that the editor isn't 100% in lockstep with LMS versions, so be aware that if this is the sort of thing that needs to update with every LMS release, it's something you may want to pass into the editor context instead of hardcoding here.

components/plugins/practice.js
```diff
@@ -0,0 +1,866 @@
...
+		super.connectedCallback();
+		if (mediaQueryList.addEventListener) mediaQueryList.addEventListener('change', this._handleResize);
+
+		const dialog = this;
+
+		postMessageListener = function(e) {
+			dialog._handlePostMessages(e, dialog);
+		};
+
+		window.addEventListener('message', postMessageListener, false);
```
grant-cleary, line 528: I think this could just be:
```
window.addEventListener('message', this._handlePostMessages, false);
```
.. and in the constructor:
```
this._handlePostMessages = this._handlePostMessages.bind(this)
```
As it stands, you're going to conflict with anything else that might set a global called `postMessageListener`.

components/plugins/practice.js
```diff
@@ -0,0 +1,866 @@
...
+				</div>
+			`;
+		}
+
+		return html`
+			<dialog class="d2l-practice-editor" ?open="${this.opened}">
+				<d2l-focus-trap ?trap=${this.opened}>
+					<div class="d2l-dialog-header">
+						<div>
+							<h2 id="${getUniqueId()}">${this.localize('practice')}</h2>
```
grant-cleary, line 624: Does this need an `id`? It's not clear to me what it's used for, since you don't seem to persist it anywhere to be fetched later.

components/plugins/practice.js
```diff
@@ -0,0 +1,866 @@
...
+					}
+					else if (tag === 'script') {
+						if (editorDocument.body.querySelector(`script[src="${path}"]`)) {
+							return true;
+						}
+						else {
+							for (let x = 0; x < editorDocument.body.querySelectorAll('script').length; x++) {
+								const scriptTag = editorDocument.body.querySelectorAll('script')[x];
+
+								if (scriptTag.src.indexOf(filename) !== -1 || (/jquery.*\.js/gi.test(scriptTag.src) === true && /jquery.*\.js/gi.test(filename) === true)) {
```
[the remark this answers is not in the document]
jschlumkoski: Nope, byproduct of copying over from Insert Element. I'll remove it.

components/plugins/practice.js
```diff
@@ -0,0 +1,866 @@
...
+				</div>
+				<div class="d2l-preview ${this._mobileView && !this._preview ? 'hidden' : 'fullscreen'}" slot="secondary">
+					${this._buildPreview()}
+					${this._mobileView && this._preview ? html`<d2l-button @click="${this._togglePreview}">${this.localize('practice:editor:done')}</d2l-button>` : ''}
+				</div>
+			`;
+		}
+
+		return html`
+			<dialog class="d2l-practice-editor" ?open="${this.opened}">
```
grant-cleary, line 620: Any reason not to use `d2l-dialog` here? You get some nice features like auto-disabling/re-enabling scrolling, the backdrop, and the focus trap.

components/plugins/practice.js
```diff
@@ -0,0 +1,866 @@
...
+				<div class="preview-icon-container">
+					<d2l-icon class="preview-icon" icon="d2l-tier3:preview"></d2l-icon>
+					<div class="preview-icon-text">${this.localize('practice:editor:preview')}</div>
+				</div>
+			`;
+		}
+	}
+
+	_change(e) {
+		this.dispatchEvent(new CustomEvent('change', { detail: e.target.value }));
```
grant-cleary, line 721: I would rename this event something more specific, e.g. `d2l-htmleditor-practice-change`, to avoid conflicting with native `change` events.

components/plugins/practice.js
```diff
@@ -0,0 +1,866 @@
...
+
+			dialog.dispatchEvent(new CustomEvent(
+				'd2l-practice-dialog-close', {
+					bubbles: true,
+					detail: result
+				})
+			);
+		}
+
+		setTimeout(save, 3000);
```
grant-cleary, line 810: Not hugely keen on usages of `setTimeout` like this. Out of curiosity, is there a reason we need to wait 3 seconds before saving?

components/plugins/practice.js
```diff
@@ -0,0 +1,866 @@
...
+import '@brightspace-ui/core/components/list/list.js';
+import '@brightspace-ui/core/components/list/list-item-button.js';
+import '@brightspace-ui/core/components/loading-spinner/loading-spinner.js';
+import '@brightspace-ui/core/components/menu/menu.js';
+import '@brightspace-ui/core/components/menu/menu-item-checkbox.js';
+import '@brightspace-ui/core/components/validation/validation-custom.js';
+import '@brightspace-ui/core/templates/primary-secondary/primary-secondary.js';
+import '../../components/practice/sequencing.js';
+import '../../components/practice/sorting.js';
+import '../../htmleditor.js';
```
grant-cleary, line 22: I don't think you're using the editor component here, so I think this can be removed.

components/plugins/practice.js
grant-cleary: Ah right, now I remember. You may be interested to know that I'm building out a framework for our own context toolbars so we can ditch TinyMCE's. Eventually when I get to converting the Insert Element / Insert Practice toolbars, I'll make a note to get rid of these. :)

components/plugins/practice.js
jschlumkoski: @grant-cleary The reason I experimented with it is that the saving is generally so instantaneous that the dialog closes and it's not immediately clear that anything happened. Open to suggestions here.

components/plugins/practice.js
jschlumkoski: I can remove this.

components/plugins/practice.js
grant-cleary: Hmm, I see. One thing to bear in mind is that in FRAs like Lessons, one disadvantage to using native dialogs is that we lose the ifrau magic done by our `d2l-dialog` component to allow the dialog to escape the inner iframe. Basically, in Lessons, I think your editor is likely going to be bounded by the FRA's iframe. @dbatiste, I wonder if you have any thoughts on this.

components/plugins/practice.js
grant-cleary: Ah, I see. I think our native dialogs get around this via a transition that makes it visually obvious that state has changed, but if you're going for a truly fullscreen experience that may not apply. I'll think about this a bit more.

components/plugins/practice.js
jschlumkoski: I'll remove it!

components/plugins/practice.js
jschlumkoski: I'll change it to your recommendation

components/plugins/practice.js
jschlumkoski: Yeah, this is a way better way of doing it. I'll swap it 😬

components/plugins/practice.js
jschlumkoski: Should be fixed in the most recent commit.

components/plugins/practice.js
jschlumkoski: Should also be fixed in the most recent commit.

components/plugins/practice.js
jschlumkoski: Should also be fixed in the most recent commit.

components/plugins/practice.js
jschlumkoski: Should also be fixed in the most recent commit.

components/plugins/practice.js
```diff
@@ -0,0 +1,866 @@
...
+		this.activityData = {};
+		this.opened = false;
+		this._activityDataURL = '';
+		this._loadedActivityData = false;
+		this._stepIndex = 0;
+		this._templateIndex = -1;
+		this._mobileView = false;
+		this._modifiers = [];
+		this._preview = false;
+		this._startedEditing = false;
```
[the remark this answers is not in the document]
jschlumkoski: Fixed in the most recent commit.

components/plugins/practice.js
jschlumkoski: Fixed in the most recent commit.

components/plugins/practice.js
jschlumkoski: Removed in most recent commit.

components/plugins/practice.js
jschlumkoski: Naming updated in most recent commit.

components/plugins/practice.js
jschlumkoski: Removed in most recent commit.

components/plugins/practice.js
```diff
@@ -0,0 +1,866 @@
...
+			messageData = JSON.parse(event.data);
+		}
+		catch (e) {
+			messageData = {};
+		}
+
+		const ifrm = dialog.shadowRoot.querySelector('iframe');
+
+		if (messageData.subject === 'getActivityContext') {
+			ifrm.contentWindow.postMessage(JSON.stringify({ subject: 'doneGetActivityContext', message: window.top.document.querySelector('html').dataset.heContext }), event.origin);
```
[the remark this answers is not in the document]
jschlumkoski: I checked and it's only requirement is orgUnitPath right now, so I've modified it to pass an object string with only that parameter.

components/plugins/practice.js
```diff
@@ -0,0 +1,866 @@
...
+					editor.editorContainer.getRootNode().querySelector('.tox-tinymce-aux').innerHTML = '';
+					editor.selection.collapse();
+
+					dialog._handleClose();
+
+					return;
+				}
+
+				const heContext = JSON.parse(window.top.document.querySelector('html').dataset.heContext);
+				const vc = new ValenceController(dialog);
```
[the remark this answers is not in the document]
jschlumkoski: @grant-cleary It's my assumption that it wouldn't currently, as the sandboxed iframe wouldn't have access to the top level frame. This is something I'll need to flag to my team, but the window references have been changed.

components/plugins/practice.js
jschlumkoski: @grant-cleary I'll flag this to my team and determine a way to proceed with this.

components/plugins/practice.js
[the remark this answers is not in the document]
grant-cleary: Right. My guess is that on the LMS side we'll want to factor that in so that `context.insertPractice` returns `false` if sandboxing is enabled. Of course if a user flips sandboxing on after the fact, any Content files that are now sandboxed will break, but that's probably unavoidable. We should probably verify that's true.

components/plugins/practice.js
grant-cleary: The way we've done this previously in the editor so far is to add the path to the context object, so we could replace these static references with a call to `context.insertPracticesAssetsPath` or something similar.

components/practice/sequencing.js
```diff
@@ -0,0 +1,656 @@
...
+					editor._updateValue(editor._imageReference, vc.storage.filename);
+					editor._updateValue(editor._imageAltReference, vc.storage.altText);
+				};
+
+				vc._uploadCourseContent(editor._context.orgUnitId, e.target.result.toString(), true, file.name, altText);
+			});
+
+			setTimeout(function() {
+				fr.readAsDataURL(file);
+			}, 2000);
```
grant-cleary, line 506: What's this timeout for?

components/practice/sorting.js
```diff
@@ -0,0 +1,917 @@
...
+					editor._updateValue(editor._imageReference, vc.storage.filename);
+					editor._updateValue(editor._imageAltReference, vc.storage.altText);
+				};
+
+				vc._uploadCourseContent(editor._context.orgUnitId, e.target.result.toString(), true, file.name, altText);
+			});
+
+			setTimeout(function() {
+				fr.readAsDataURL(file);
+			}, 2000);
```
grant-cleary, line 754: Same question RE: the timeout. Also, this looks like very similar code. I can't tell easily if some of these functions are literally identical, but if they are, maybe a mixin would be a better place for them?

components/practice/sequencing.js
jschlumkoski: @grant-cleary Similar to the one that happens on save, it's because the upload usually happens so fast that the dialog closes and you get very little indication that the upload actually happened.

components/practice/sorting.js
jschlumkoski: @grant-cleary Similar to the one that happens on save, it's because the upload usually happens so fast that the dialog closes and you get very little indication that the upload actually happened.

Right now they're fairly identical, but it's possible that they may need to handle things differently in the future.

components/practice/sorting.js
```diff
@@ -0,0 +1,917 @@
...
+		super.disconnectedCallback();
+	}
+
+	firstUpdated() {
+		super.firstUpdated();
+
+		const e = this.closest('.d2l-editor');
+
+		if (e !== null) {
+			this._resizeObserver.observe(e);
```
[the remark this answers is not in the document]
jschlumkoski: @grant-cleary I've moved this to the `connectedCallback` instead. I've also added a condition to check for categories before doing anything to prevent errors the first time this runs when the categories list won't exist yet.

components/practice/sorting.js
grant-cleary: So usually our design paradigm for something like this is to rely on the user clicking on some other button to close out the dialog after we give them some visual indication that the upload has completed. However, the use of an actual fullscreen dialog maybe makes that awkward?

components/plugins/practice.js
grant-cleary: Yeah, that certainly makes things easier from the editor's point of view, but ultimately I'm fine with both; just if we do want to keep it mapped to an LMS version, we probably want it to live outside of the editor.

jschlumkoski: @grant-cleary We've opted to remove the version numbers.

components/practice/sequencing.js
jschlumkoski: @grant-cleary I've removed the timeouts for the saving of images in the activities. It now updates with a completion message and instructs the user to close the dialog.

components/practice/sorting.js
jschlumkoski: @grant-cleary The file upload timeouts have been removed in favor of a completion message and a click action to close it.

components/plugins/practice.js
jschlumkoski: @grant-cleary My thinking is it may be best to leave this while we work out a better way of handling this. Also in the event we shifted to the d2l dialog, this could be removed.

components/plugins/practice.js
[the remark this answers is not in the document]
jschlumkoski: @grant-cleary Right now, we're going to move forward with the messaging that sandboxing will need to be off to use this tool. We can look at adding that piece to the LMS though.

components/plugins/practice.js
jschlumkoski: @grant-cleary My teams thoughts is that this may be a necessary evil right now. The fullscreen dialog just doesn't provide us enough space for the editing experience we're trying to offer, and while limited, being bound to the FRA's iframe is still preferable right now.

dbatiste: I don't think we should be doing this.  As Grant mentioned, the dialog component is providing a lot of behaviour that we want to be consistent throughout the app.

How much extra space is this providing?  Our "fullscreen" dialog variant nearly fills the entire frame.

grant-cleary: Right, but the backdrop wouldn't/shouldn't be, since that's handled through ifrau. So I believe we lose that as well.

dbatiste: Btw, the fullscreen dialog would be `d2l-dialog-fullscreen`.  Just wanted to mention that in case you weren't aware of it.  It doesn't fill the entire frame, but pretty close.

htmleditor.js
```diff
@@ -654,7 +654,7 @@ class HtmlEditor extends HtmlEditorMixin(SkeletonMixin(LitElement)) {
 						${this._renderProcessedHtml()}
 					</d2l-html-block>
 				</div>
-				${this._toolbar ? this._renderEditor(this._toolbar.render(this._context, this.sourceEditable, this.fullPage, this.insertElement, this._fullscreen), false) : ''}
+				${this._toolbar ? this._renderEditor(this._toolbar.render(this._context, this.sourceEditable, this.fullPage, this.insertElement, this.insertPractice, this._fullscreen), false) : ''}
```
Trandrew17, line 657: Does this require adding to the `insertPractice` parameter here? https://github.com/BrightspaceUI/htmleditor/blob/main/components/toolbar/toolbar-inline.js#L165

grant-cleary: That's probably true, yeah. Otherwise this won't work properly in a fullscreen inline editor. Currently that's not a thing for any editors that use this property, but we should be robust nonetheless.

jschlumkoski: @grant-cleary Is this something we should add to this for Insert Element and Insert Practice? Or does it warrant it's own PR after this stuff gets merged?

grant-cleary:
> @grant-cleary I've just pushed some changes to better align with the approach outlined by Chris this morning. The custom dialog code has been moved into a new component called "d2l-immersive-dialog" and is using the DialogMixin as well as a lot of the existing dialog CSS (I did have to make some changes there to align to our current design though). I also had to make a couple of modifications to the event handling also, primarily to get rid of the timeout and add a specific click action to close the editor dialog. Also, to prevent inner dialog windows from closing out the editor also.

Excellent, thanks much! I'll take a look at the new changes.

htmleditor.js
grant-cleary: Hmm, I think it's already done for `insertElement` so you would just need to do the same for `insertPractice`. Probably pre-merge, given it's a very small change.

Trandrew17:
Heads up I expect to be merging https://github.com/BrightspaceUI/htmleditor/pull/739 soon and I expect some conflicts to come up here from that.

jschlumkoski:
> Heads up I expect to be merging #739 soon and I expect some conflicts to come up here from that.

@Trandrew17 Okay cool, I'll keep an eye out.

components/practice/sequencing.js
```diff
@@ -555,6 +555,8 @@ class SequencingPractice extends RequesterMixin(Localizer(LitElement)) {
 	}
 
 	_stopUpload() {
+		e.stopPropagation();
```
grant-cleary, line 558: I don't think `e` is defined here; it has to be explicitly passed into `_stopUpload` as a param.

jschlumkoski: Ah drat, I missed that when adding this line to this file 🤦‍♂️. Adding it now!

grant-cleary (dismissed):

jschlumkoski:
@grant-cleary I added a new property to our one post message, and some last minute lang term updates (1 typo, 1 reword).

grant-cleary:
> @grant-cleary I added a new property to our one post message, and some last minute lang term updates (1 typo, 1 reword).

LGTM. Feel free to merge.

components/practice/d2l-immersive-dialog.js
```diff
@@ -0,0 +1,256 @@
...
+import { getUniqueId } from '@brightspace-ui/core/helpers/uniqueId';
+import { ifDefined } from 'lit/directives/if-defined.js';
+import { LocalizeCoreElement } from '@brightspace-ui/core/helpers/localize-core-element.js';
+
+/**
+ * A generic dialog that provides a slot for arbitrary content and a "footer" slot for workflow buttons. Apply the "data-dialog-action" attribute to workflow buttons to automatically close the dialog with the action value.
+ * @slot - Default slot for content inside dialog
+ * @slot footer - Slot for footer content such as workflow buttons
+ */
+class ImmersiveDialog extends LocalizeCoreElement(DialogMixin(LitElement)) {
```
grant-cleary, line 15: Oh wait, hold on, I just noticed this. We don't want to use `LocalizeCoreElement` here. Instead we should be using the editor's `Localizer` mixin.

grant-cleary: `LocalizeCoreElement` makes assumptions about where the lang files live and how to hook them up to the LMS that don't apply to the editor.

jschlumkoski: @grant-cleary Ah, gotcha! Should be using the right one now.

grant-cleary (approved):

PR #727, opened 2022-04-05
Implement context toolbars for formatting/links

grant-cleary:
Draft for now to try to iron out some conceptual questions I want to get sorted.


components/context-menu/context-menu.js
```diff
@@ -0,0 +1,456 @@
...
+		const positionStyle = {};
+		if (this._contextMenuShift !== undefined) {
+			const isRtl = this.getAttribute('dir') === 'rtl';
+
+			positionStyle.left = (!isRtl && !this._openLeft) || (isRtl && this._openLeft) ? `${this._contextMenuShift}px` : null;
+			positionStyle.right = (!isRtl && !this._openLeft) || (isRtl && this._openLeft) ? null : `${this._contextMenuShift}px`;
+		}
+
+		return html`
+			<div class="d2l-htmleditor-context-menu-container" role="toolbar" tabindex="${this.showing ? 0 : -1}" @focus="${this._handleFocus}" @keydown="${this._handleKeyDown}">
```
grant-cleary, line 308: Concern #1: Focus Management

Design ideally wants users to be able to tab from the editor into the context menu if it exists and is open. The easiest way to accomplish this is to just put the context menu into the tab order normally, since it falls right after the editor container within the component. However, this also means a bunch of manual focus management to make sure tabbing moves focus where it should. I'm unsure about this focus management strategy.

components/context-menu/context-menu.js
```diff
@@ -0,0 +1,456 @@
...
+				<span class="d2l-offscreen" aria-live="polite">${this.localize('contextMenu:offscreen')}</span>
```
grant-cleary, line 309: Concern 2: Accessibility

Because the context menu is being overlayed over TinyMCE's iframe and thus is in a different document from the user's selection, I'm at a bit of a loss over how we should alert screenreader users to the existence of this new menu. I threw this together super quickly as a start, but it hardly seems ideal.

htmleditor.js
```diff
@@ -513,6 +514,10 @@ class HtmlEditor extends HtmlEditorMixin(SkeletonMixin(LitElement)) {
 		this.focus();
 	}
 
+	_registerContextMenuPlugin(plugin) {
```
grant-cleary, line 517: Concern 3: Registration

For links and formatting, we don't have a need for any kind of component-specific registration, since they're always available. However, if we want to extend this to other context menus (Image Tools, Insert Element, etc.), then they're going to have to be loaded conditionally based on various editor properties. This means we need some means of registering context menu plugins against the component. I went back and forth on the approach a few times before tentatively doing this, but I'm also a bit unhappy because now plugins are accessing private functions/properties on the component.

components/toolbar/toolbar-item-mixin.js
```diff
@@ -25,23 +25,28 @@ export const ToolbarItemMixin = superclass => class extends RtlMixin(superclass)
...
 	disconnectedCallback() {
 		super.disconnectedCallback();
-		resizeObserver.unobserve(this);
+		if (!this.noResize) resizeObserver.unobserve(this);
 	}
 
 	firstUpdated() {
 		super.firstUpdated();
-		resizeObserver.observe(this);
+		if (!this.noResize) resizeObserver.observe(this);
```
grant-cleary, line 49: Er, I forgot to implement this properly. I didn't think it was desirable to exercise resize handlers when we know these context toolbars aren't going to do any resizing, but I did still want to be able to reuse core components between the standard toolbar and any context menu toolbars.

components/context-menu/link/link-actions.js
```diff
@@ -0,0 +1,70 @@
...
+
+	constructor() {
+		super();
+		this.location = '';
+	}
+
+	connectedCallback() {
+		super.connectedCallback();
+		const context = this.requestInstance('context');
+		this._orgUnitId = context && context.orgUnitId;
```
grant-cleary, line 30: Not 100% sure how I feel about having a context reference here, but this is necessary to resolve quicklinks properly and is the primary impetus for why we're making this change in the first place.

components/toolbar/link-button.js
```diff
@@ -0,0 +1,117 @@
...
+	render() {
+		return html`
+			<a aria-label="${this.text}"
+				href="${this.href}"
+				tabindex="${this.activeFocusable ? 0 : -1}"
+				target="_blank"
+				title="${this.text}">
+				<div class="d2l-htmleditor-button-background"></div>
+				${this.icon ? unsafeHTML(icons[this.icon] || '') : html`<slot></slot>`}
+			</a>`;
```
grant-cleary, line 106: My understanding is that it's far better to style a native `<a>` tag as a button than to create an actual `<button>` that opens a link.

components/context-menu/link/link-display.js
```diff
@@ -0,0 +1,107 @@
...
+		let location = '';
+		let text = '';
+
+		if (this.linkLocation) {
+			location = this.linkLocation.length <= maxDisplayLength ? this.linkLocation : `${this.linkLocation.substring(0, maxDisplayLength)}...`;
+		}
+
+		if (this.linkText) {
+			text = this.linkText.length <= maxDisplayLength ? this.linkText : `${this.linkText.substring(0, maxDisplayLength)}...`;
+		}
```
grant-cleary, line 91: As discussed offline, this is something I threw together quickly to make the "super long link" problem go away for testing/demoing. I took a quick look at `core` and didn't see obvious solutions outside of `more-less`, but I wasn't 100% sure that was the best fit here, since expanding out the toolbar also seems strange. Then again, this is already quite the edge case, so maybe "weird" doesn't so much matter.

components/context-menu/formatting/formatting.js
```diff
@@ -0,0 +1,92 @@
...
+		const isNonEditable = (node) => {
+			if (node.nodeType === Node.ELEMENT_NODE) {
+				if (node.hasAttribute('mceNonEditable')) return true;
+
+				const tinyMceContentEditable = node.getAttribute('data-mce-contenteditable');
+				if (tinyMceContentEditable === 'false') return true;
+				if (node.contentEditable === 'false') return true;
+			} else {
+				return false;
+			}
```
grant-cleary, line 42: This block can be simplified by just checking `if (node.nodeType !== Node.ELEMENT_NODE)` first.

htmleditor.js
grant-cleary: An additional note. This doesn't actually need to be tracked as a `Map`. If we wanted to keep the same structure, we could ditch the `name` property on the plugins and simply always pass them around as classes with entirely static functions/properties. They don't store state and in this case I'm just using them as convenient "holders" for all of their functions.

components/context-menu/context-menu.js
```diff
@@ -0,0 +1,456 @@
...
+	}
+
+	_handleKeyDown(e) {
+		if (!this.showing) return;
+		if (e.keyCode !== 27) return;
+
+		// Escape key exits menu and returns focus to the editor
+		this.dispatchEvent(new CustomEvent(
+			'd2l-htmleditor-context-menu-close', { bubbles: true }
+		));
```
grant-cleary, line 361: There's a downside here... There's no means to use a shortcut to close the editor from within TinyMCE. Maybe that's okay, but my worry is that binding `Esc` to close open context menus within the editor iframe could disrupt the editor's existing keybinds for `Esc`. I should test that though (and figure out whether closing the context toolbar that way is even desirable).

htmleditor-mixin.js
```diff
@@ -472,6 +472,13 @@ export const HtmlEditorMixin = superclass => class extends Localizer(LabelledMix
 		// need to wait a frame for tinyMCE to lose focus, or for tinyMCE dialog to place focus
 		requestAnimationFrame(async() => {
 
+			const contextMenu = this.shadowRoot.querySelector('d2l-htmleditor-context-menu');
+
+			if (this._dialogOpen || this._dropdownOpen) {
+				// If we have an open context menu and we're opening a dialog or dropdown, hide it
+				if (contextMenu && contextMenu.showing) contextMenu.showing = false;
```
grant-cleary, line 479: Known gap that I haven't addressed yet: This also needs to happen on TinyMCE's `OpenWindow` event. I didn't think about it first because opening TinyMCE's dialogs typically requires opening one of the editor's dropdowns, which will trigger this logic. However, some enterprising users have discovered that some TinyMCE shortcuts can be used to open their dialogs directly without going through our dropdowns first. In that case, the context menu will get stuck open.

components/context-menu/link/link-display.js
dbatiste: Yeah, custom line-clamping is treacherous.  I think we should explore using [-webkit-line-clamp](https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-line-clamp).  See [caniuse](https://caniuse.com/?search=line-clamp)... all the browsers are at least temporarily supporting this prefix.  This is the result from my quick experiment (in Chrome).

```css
.d2l-heading-4 {
   display: -webkit-box;
   overflow: hidden;
   -webkit-box-orient: vertical;
   -webkit-line-clamp: 2;
}
```

image

components/toolbar/link-button.js
dbatiste: Yeah, the tag should match the semantics regardless of the visuals.  Take for instance, right-clicking on an `<a>` element... does it make sense for the browser's context menu to say "Open in New Tab"?

htmleditor-mixin.js
dbatiste: Maybe we need to somehow listen for those key-strokes... or maybe detect when a tinyMCE dialog has been opened, so that we can manage this state completely?

htmleditor.js
dbatiste: I think this is ok... maybe it only bothers you because it is private, and it should be public?  I'm not a fan of advertising some "public" methods to consumers, as they are only public within the scope of the editor, but I also don't like seeing private members accessed from outside their scope.

components/context-menu/context-menu.js
dbatiste: Normally with dropdown's there would be some Aria that communicates the presence of the popup and the state?  Are we unable to use that here because the focus is in the content-editable?  And commonly, we would move focus into the content when the dropdown opens.

We do have the [announce](https://github.com/BrightspaceUI/core/blob/main/helpers/announce.js) helper, which helps to manage some of the problems that can arise with live regions.  I am not saying to absolutely not use a live region here, but it's another option.

I might play around with it with screen-reader to get a better understanding of the experience.  Perhaps we should discuss also.

components/context-menu/context-menu.js
dbatiste: I think that sounds reasonable but I think it would be good to hash out the edge cases and what the focus management needs to consider.

components/context-menu/context-menu.js
dbatiste: By "close the editor", do you mean close the context menu?

Yeah, that is a concern/interesting.  In core we use the [dismissible](https://github.com/BrightspaceUI/core/blob/main/helpers/dismissible.js) helper to manage this, but it assumes only one dismissible element can be on the top of the stack.  

Is it possible to have both a context menu open, and some other dismissible element open, where we can't figure out what to close?  If a dialog is open, the user's focus must be inside, and the dialog would close if the [Esc] is pressed.  If the user's focus is in the content and they hit [Esc], what dismissible things could be open? 🤔

htmleditor.js
grant-cleary: Yeah, that sums up what bothers me. Making it public (to me) implies that we're potentially advertising this as being available to consumers, whereas that's definitely not the intent. On the other hand, accessing private things feels bad, like you said. I probably just need to get used to accepting that things can be "public" without advertising their use.

components/context-menu/link/link-display.js
grant-cleary: Cool, seems good to me. I'll make the change. I'm certainly happy to use something out-of-the-box.

components/context-menu/context-menu.js
grant-cleary: Yeah, sorry, I meant "close the context menu". Wires just crossed a bit there. I _think_ it's reasonable to say we should only have one dismissible thing open at a time (for context menus... Setting aside the case of potentially nested dialogs)... In general, I assume the logic should be that opening any other dismissible thing should also dismiss the context menu.

components/context-menu/context-menu-toolbar-mixin.js
```diff
@@ -0,0 +1,75 @@
...
+
+	_render(items) {
+		return html`
+			<div
+				class="d2l-htmleditor-context-menu-toolbar-container"
+				@keydown="${this._handleKeyDown}"
+				role="toolbar"
+				tabindex="-1">
+				<div class="d2l-htmleditor-context-menu-toolbar-actions">
+					<div>${items}</div>
```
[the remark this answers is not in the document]
grant-cleary: Yeah, I was musing about that. My desire to reuse the buttons for the toolbar led me into some hang-ups here for what I wanted to accomplish. It's maybe worth revisiting how that is structured as well to make that simpler, but I didn't want to touch that quite yet, haha.

And yeah, I can remove the div; I shamelessly stole this from the main toolbar logic, but in that case I think this div was used for styling that's not relevant in context menus.

components/context-menu/context-menu.js
grant-cleary: Yeah, so the problem I found (and I'm not 100% sure if I was just doing something wrong) was that the appropriate Aria labels on the popup weren't firing and alerting. I assumed this had something to do with focus being in the content-editable (basically because it's an entirely different HTML document from where the menu lives), but I wasn't completely certain.

It's definitely true that if focus moves into the context menu, this isn't a problem, but that's obviously not desirable behaviour for a context menu so I needed something that would announce without shifting focus.

I mostly mucked around with NVDA to test this stuff out, but these interactions are things I'm not as familiar with so I don't want to rule out the possibility that I just missed something obvious. If I didn't and we need to do something like this, the `announce` helper seems better. I found the live region to be a bit hit-or-miss... If it's assertive, it's ridiculously noisy every time someone changes the text selection. If it's polite, it sometimes just doesn't read if the selection covers a big chunk of text.

Also happy to discuss as well. This is something I for sure want to get right before anything gets merged.

components/context-menu/context-menu.js
grant-cleary: Yeah, right now interactions are relatively minimal so it largely seems to work how I would expect. Tabbing into the context menu moves to the first focusable and then tabbing moves through any remaining focusables in the menu until they're done, and then out of the editor proper. That last bit is maybe uncertain. I'm not sure if it makes sense for focus to leave the editor after tabbing out of the menu.

I didn't think a focus trap was appropriate, since shift-tabbing back into the editor seems sensible, but I'm not sure if we want to loop focus back to the editor when tabbing out of the menu. I could see it both ways I suppose.

components/context-menu/context-menu-toolbar-mixin.js
```diff
@@ -0,0 +1,75 @@
...
+				</div>
+			</div>
+		`;
+	}
+
+	_resetActiveFocusable() {
```
dbatiste, line 70: Is this needed here?  Does it ever get called?

grant-cleary: No sir. More stuff that was used that I forgot to prune when I removed it.

grant-cleary: Well, probably. Now that I think about it, it might be appropriate to remove the active focusable on focusout.

components/context-menu/context-menu.js
```diff
@@ -0,0 +1,456 @@
...
+
+			contextMenu.addEventListener('d2l-htmleditor-context-menu-close', () => {
+				// Return focus to the editor when context menu is manually closed
+				component.focus();
+			});
+
+			return contextMenu;
+		};
+
+		editor.on('SelectionChange', async() => {
```
dbatiste, line 112: It looks like this event gets dispatched for pretty much every key-stroke... just moving the caret around without "selecting" anything.  I think this is necessary since some context menus (such as link context menu) doesn't actually require a selection... it just requires being in a certain type of node. My assumption is that tinyMCE runs these handlers synchronously... otherwise I think there would be trouble.  Anyway, more of a comment to be watch for jank moving the caret around, and to look for ways to make this as efficient as possible.

components/context-menu/context-menu.js
```diff
@@ -0,0 +1,456 @@
...
+			contextMenu.showing = false;
+		});
+
+		editor.on('ResizeContent ScrollContent', () => {
+			const contextMenu = component.shadowRoot.querySelector('d2l-htmleditor-context-menu');
+			if (!contextMenu || !contextMenu.showing) return;
+
+			const sel = getSelectionInfo();
+
+			for (const plugin of component.getAllContextMenuPlugins()) {
```
dbatiste, line 132: If we find scrolling or resizing is janky, we have the option to close the context menu.  Seems ok from my testing though.

components/context-menu/context-menu.js
```diff
@@ -0,0 +1,456 @@
...
+				text-align: right;
+			}
+			:host([showing]) {
+				display: inline-block;
+			}
+			.d2l-htmleditor-context-menu-pointer {
+				border: 1px solid transparent; /* fixes a webket clipping defect */
+				box-sizing: border-box;
+				display: inline-block;
+				height: ${pointerLength}px;
```
dbatiste, line 182: Take a peek at the Lit documentation regarding [using expressions in static styles](https://lit.dev/docs/components/styles/#expressions).  I suspect Lit is not complaining because the variable is declared as a number, not a string.  I almost like to get one of the Lit folks to say Numbers are "ok".

components/context-menu/context-menu.js
```diff
@@ -0,0 +1,456 @@
...
+			:host([dir="rtl"][_open-left]) .d2l-htmleditor-context-menu-content {
+				left: 0;
+				right: auto;
+			}
+			.d2l-htmleditor-context-menu-container {
+				height: 100%;
+				width: 100%;
+			}
+			:host .d2l-htmleditor-context-menu-container {
+				-webkit-animation: d2l-htmleditor-context-menu-bottom-animation 200ms ease;
```
dbatiste, line 263: I am pretty sure that `-webkit-animation` is no longer needed - I think Safari has supported `animation` for quite a while.  [caniuse](https://caniuse.com/?search=animation)  I know we have bunches of code that still specify it... just hasn't been cleaned up.

components/context-menu/context-menu.js
```diff
@@ -0,0 +1,456 @@
...
+					</div>
+				</div>
+				<div class="d2l-htmleditor-context-menu-pointer">
+					<div></div>
+				</div>
+			</div>
+		`;
+	}
+
+	updated() {
```
dbatiste, line 322: It is a good practice to always accept the changed properties, and pass them along the the `super` method.

```javascript
updated(changedProperties) {
    super.firstUpdated(changedProperties);
    ...
}
```

components/context-menu/context-menu.js
```diff
@@ -0,0 +1,456 @@
...
+			top: top,
+			width: documentRelativeTargetRect.width,
+			height: 0,
+		};
+
+		// Compute how much the menu is shifted relative to its pointer
+		// (left align context menu with the left edge of the target)
+		this._contextMenuShift = Math.min((documentRelativeTargetRect.width / 2) - (contentPadding + pointerRotatedLength / 2), 0);
+
+		this.style.left = `${positionRect.left}px`;
```
dbatiste, line 448: This appears to be modifying the DOM outside of render.  It looks like tooltip does this also, but dropdown avoids that.  I am surprised that tooltip appears to, because the idea of not updating the DOM outside of render existed long before tooltip was written.  If the context-menu re-renders itself for some reason, these styles will be blown away.

components/context-menu/context-menu.js
```diff
@@ -0,0 +1,456 @@
...
+
+export function addContextMenuPlugin() {
+	tinymce.PluginManager.add(name, function(editor) {
+		const context = requestInstance(editor.getElement(), 'context');
+		const component = context.component;
+
+		const getSelectionInfo = () => {
+			return {
+				range: editor.selection && editor.selection.getRng(),
+				node: editor.selection && editor.selection.getNode()
```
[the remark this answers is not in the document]
grant-cleary: I believe it's just the `commonAncestorNode` of the range, but I'm not super sure if TinyMCE is doing something different under the hood.

components/context-menu/context-menu.js
```diff
@@ -0,0 +1,456 @@
...
+			};
+		};
+
+		const getPositioningInfo = (plugin, selectionInfo) => {
+			const targetRect = plugin.getTargetRect(selectionInfo.range, selectionInfo.node);
+			const boundingRect = editor.getContentAreaContainer().getBoundingClientRect();
+
+			const getRectObject = rect => {
```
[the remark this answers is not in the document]
grant-cleary: Ah. Yeah, I wasn't sure whether it was good practice to bind a Lit property to a `DOMRect` object directly. Creating a new object with only the properties I cared about was my solution to that quandary.

components/context-menu/context-menu.js
grant-cleary: Yeah, Originally I tried doing this on `NodeChange` instead (still fires basically all the time), but I found that made the jank much worse because the `NodeChange` event fires as the selection is starting, not as it ends. But yes, TinyMCE for sure runs these handlers synchronously. We've run into problems with that before (trying to do async things in sync handlers).

components/context-menu/context-menu.js
grant-cleary: Yeah, I found it to be relatively smooth so I wanted to give it a shot at keeping it open. It seems to react pretty well to rapid resizing. But yeah, a fallback is certainly to just close it.

components/context-menu/context-menu.js
grant-cleary: Yeah, that's how I interpret this at least (emphasis mine):
```
To prevent Lit components from evaluating potentially malicious code, the css tag only allows nested expressions that are themselves css tagged strings **or numbers**.
```
But I'm not 100% sure. I don't think it's the end of the world to just hard-code it either. It seems unlikely we're going to ever change this and a comment could easily get the point across that they should change in tandem.

components/context-menu/context-menu.js
grant-cleary: Cool. I can drop it. I just cribbed the styles from the dropdown stuff and didn't look too carefully at what might have been obsolete.

components/context-menu/context-menu.js
grant-cleary: Yeah, this was just missed in error. I'll add it.

components/context-menu/context-menu.js
grant-cleary: It is, yeah, and I did model it after the tooltip logic - the dropdown positioning stuff wasn't quite what we needed here. I'm not 100% sure why tooltip took that direction, but I did find that the timing of some of this positioning logic was extremely sensitive, so it might just be that adding an additional re-render may have made the timing even more hellish.

I can experiment with treating these as properties and seeing if I can make the timing work well. What we're doing here is actually a lot less complicated than what tooltip needs to do, so it's quite possible that it will just work.

components/context-menu/context-menu.js
grant-cleary: Of course, just wrapping it with `unsafeCSS` is the easier option and I could easily do that as well if we want to be extra sure.

components/context-menu/context-menu.js
grant-cleary: Discussed this offline, and this case is a bit weird because (like the tooltip), these styles are being added to the host component (otherwise it gets anchored below the editor, not the opener). Unlike tooltip, the position of the context menu component is always consistent, so I think we could probably get around this by computing the offset of the target compared to where the context menu component is positioned.

That said, for now I moved the style setting into the `render` method for now. If the context menu's parent component re-renders, it will probably nuke those styles... However, the context menu's parent is the editor itself, so at that point we're nuking the context menu too. I'll take another look at this later, but I think for now what we have will work okay.

Trandrew17 (approved):

PR #739, opened 2022-04-11
Add Content Styler Plugin

Trandrew17:


components/plugins/contentstyler.js
```diff
@@ -0,0 +1,42 @@
...
+				const elem = editorDocument.createElement('link');
+				elem.rel = 'stylesheet';
+				elem.href = styleUrl;
+				editorDocument.head.appendChild(elem);
+				e.content = editorDocument.documentElement.outerHTML;
+			}
+
+			const context = requestInstance(editor.getElement(), 'context');
+			const overrideUrl = context.contentStylerCssOverride;
+			const existingOverride = editorDocument.head.querySelector(`link[data-override='override']`);
```
grant-cleary, line 21: `data-override="override"` was correct. The linter is complaining because you're using backticks for this string even though it has no replacements. The line should look like this:
```
const existingOverride = editorDocument.head.querySelector('link[data-override="override"]');
```

components/plugins/contentstyler.js
```diff
@@ -0,0 +1,42 @@
+import { requestInstance } from '@brightspace-ui/core/mixins/provider-mixin.js';
+const name = 'd2l-contentstyler';
```
grant-cleary, line 2: Nit: Would prefer a newline between the import and this line.

components/toolbar/toolbar-full.js
```diff
@@ -42,6 +43,7 @@ export function getPlugins(component) {
 	return [
 		addA11yPlugin(),
 		addAttributesPlugin(),
+		component.fullPage && loadContentStyler(component) ? addContentStylerPlugin() : '',
```
grant-cleary, line 46: Nit: This should probably come right after `addColorPlugin()`.

components/toolbar/toolbar-full.js
```diff
@@ -104,6 +107,7 @@ class Toolbar extends ToolbarMixin(LitElement) {
 	constructor() {
 		super();
 		this.context = {};
+		this.contentStyler = false;
```
grant-cleary, line 110: Nit: Prefer alphabetical order (this should come before `context`).

components/toolbar/toolbar-full.js
```diff
@@ -449,10 +453,14 @@ function allowInsertElement(component) {
 	return component.insertElement && component._context && component._context.insertElement;
 }
 
+function loadContentStyler(component) {
```
grant-cleary, line 456: For consistency's sake, I would probably rename this `allowContentStyler`.

grant-cleary: Though if you did that, you would need to reorder it to come before `allowInsertElement`.

components/toolbar/toolbar-full.js
Trandrew17: Yep, apparently I don't know my abc's.

grant-cleary (approved):

grant-cleary:
Just an FYI, you don't need to create an empty commit like that. If the merge commit (the commit name, not the optional description) contains the keyword, it will still trigger a release.

PR #761, opened 2022-04-20
fix: add missing `.js` from imports

devpow112:
This causes some downstream issue with folio at least.
https://paste.d2l.dev/raw/evipecasiz


devpow112:
> Thanks for this! Probably can be opened for review; I don't think there's anything controversial about this.

I always do draft just in case CI explodes for some reason 😛

PR #767, opened 2022-04-20
Don't directly import ValenceController

grant-cleary:
I don't really like this solution because it feels gross... However, webpack doesn't allow directly importing from external URLs (either statically or dynamically) and this is breaking stuff right now, so we need a solution in place that will do the job. Longer term, I think there are better solutions, but this will probably do for now.


devpow112 (approved):
Agreed it's not great but appreciate the willingness to get us at least unblocked. Damn you webpack!!!!

components/practice/valence-controller.js
```diff
@@ -0,0 +1,20 @@
+export async function loadValenceController(node) {
+	if (window.ValenceController) return new window.ValenceController(node);
+
+	await new Promise(resolve => {
+		window.setValenceControllerLoaded = resolve;
```
grant-cleary, line 5: Inline scripts don't fire `load` events because there's nothing to download. Instead, we provide a dummy function on the window that resolves the promise. If we're worried about polluting the global, the script can certainly set this to undefined once it's done what it needs to do.

Trandrew17 (approved):

PR #772, opened 2022-04-21
Reduce undesirable whitespace from inline editor's internal styling

grant-cleary:
This follows on from some margin removals in the html-block component. We want to mimic that in the editor's inline mode.

I have some uncertainty about whether we want the margin piece to be controlled via a property... I feel like the answer is "no" because we want our inline experience to be consistent, but I'm also open to other suggestions.


htmleditor-mixin.js
```diff
@@ -67,17 +67,26 @@ const contentFragmentStyles = css`
 		font-family: 'Lato', sans-serif;
 		font-size: 0.95rem;
 		font-weight: normal;
-		letter-spacing: 0.01rem;
-		line-height: 1.4rem;
+		letter-spacing: 0.02rem;
+		line-height: 1.2rem;
 		margin: 0;
-		padding: 0.4rem 0.75rem !important;
+		padding: calc(0.4rem - 1px) calc(0.75rem - 1px) !important;
```
grant-cleary, line 73: This accounts for the editor border. It does impact full mode, which doesn't have borders in the same way, but it's nearly imperceptible so I'm not particularly concerned.

htmleditor-mixin.js
```diff
@@ -67,17 +67,26 @@ const contentFragmentStyles = css`
...
 	}
 `.cssText;
 
+const inlineStyles = css`
+	body > :first-child {
+		margin-top: 0;
+	}
+	body > :last-child {
+		margin-bottom: 0;
+	}
```
grant-cleary, line 87: This styling is specifically applied on the `d2l-html-block`'s rendered div, which doesn't apply to TinyMCE's iframe body. So we have to add it ourselves.

htmleditor-mixin.js
```diff
@@ -99,15 +108,15 @@ export const HtmlEditorMixin = superclass => class extends Localizer(LabelledMix
...
 			 */
-			contentStyler: { type: Boolean, attribute: 'content-styler' },
+			compact: { type: Boolean },
 			/**
-			 * Whether or not to use compact styles
+			 * ADVANCED: Whether to enable the content styler plugin
 			 * @type {boolean}
 			 */
-			compact: { type: Boolean },
+			contentStyler: { type: Boolean, attribute: 'content-styler' },
```
grant-cleary, line 119: Re-ordering non-alphabetical params.

htmleditor-mixin.js
```diff
@@ -263,6 +272,7 @@ export const HtmlEditorMixin = superclass => class extends Localizer(LabelledMix
 		super();
 		this.attachedImagesOnly = false;
 		this.autoSave = false;
+		this.compact = false;
```
grant-cleary, line 275: This would have defaulted to false anyways, but it's nice to explicitly include it.

htmleditor-mixin.js
```diff
@@ -514,7 +524,7 @@ export const HtmlEditorMixin = superclass => class extends Localizer(LabelledMix
 		const body = editor.getBody();
 		const editorContainer = editor.getContainer();
 
-		const minHeight = this._getUnitlessHeight(this.height);
+		const minHeight = this._getUnitlessHeight(this.height) - 4; // 2px border
```
grant-cleary, line 527: Prevents autoresize pop-in as it tries to resize the editor container to a height that's actually marginally too large because of the inline border.

grant-cleary:
This is what this looks like post-change:

InlineEditorNoMargins_EditMode
InlineEditorNoMargins_ReadMode

I find it looks a bit weird when the editor is large like this, hence why I wonder if we want to apply some of our own margins to counter-act the loss of implicit margins (and allow consumers to opt out of that).

htmleditor-mixin.js
```diff
@@ -67,17 +67,29 @@ const contentFragmentStyles = css`
 		font-family: 'Lato', sans-serif;
 		font-size: 0.95rem;
 		font-weight: normal;
-		letter-spacing: 0.01rem;
-		line-height: 1.4rem;
+		letter-spacing: 0.02rem;
+		line-height: 1.47;
 		margin: 0;
-		padding: 0.4rem 0.75rem !important;
+		padding: calc(0.4rem - 1px) calc(0.75rem - 1px); /* 1px border */
```
grant-cleary, line 73: This doesn't appear to need to be `!important`. I think it was relevant before when we were still importing some of TinyMCE's body styles.

grant-cleary:
> This is what this looks like post-change:
> ...
> I find it looks a bit weird when the editor is large like this, hence why I wonder if we want to apply some of our own margins to counter-act the loss of implicit margins (and allow consumers to opt out of that).

Ran this by Noah and he agreed that it looks a bit off, but that we definitely want it in the single-line sort of case. He suggested adding 6px of top/bottom padding otherwise, so that's what I implemented. An attribute on the editor allows consumers to explicitly opt out of that additional padding.

Trandrew17 (approved):

PR #773, opened 2022-04-21
Reconfigure update-package-lock action now that dependency patch updates are becoming more frequent

grant-cleary:
Does what it says on the tin. CodeMirror updates very frequently and we're often getting 5-6 Dependabot PRs in a given day. I think it's time to switch back to the action now.


CODEOWNERS
```diff
@@ -1,2 +1,3 @@
 *       @dbatiste @BrightspaceUI/html-editor-maintainers
 README.md @margaree @dbatiste @BrightspaceUI/html-editor-maintainers
+package-lock.json
```
grant-cleary, line 3: Need to exclude this from CODEOWNERS in order for auto-merge to work properly. I'm not too concerned about this, given that any PRs will still require reviews.

Trandrew17 (approved):

PR #780, opened 2022-04-22
Fix up code editor dependencies and remove now-unnecessary resizing logic

grant-cleary:
CodeMirror had a set of breaking version changes, so let's fix it all up!


components/plugins/source-code.js
```diff
@@ -237,32 +219,30 @@ class SourceCodeDialog extends Localizer(RtlMixin(LitElement)) {
...
 				...search.searchKeymap,
-			])
+			]),
+			language.bracketMatching(),
+			language.foldGutter(),
+			language.indentOnInput(),
+			language.syntaxHighlighting(language.defaultHighlightStyle),
+			lineNumbers(),
+			rectangularSelection(),
+			search.highlightSelectionMatches()
```
grant-cleary, line 245: CodeMirror shifted a bunch of stuff from some of their secondary packages into others, so this is mostly a reflection of that.

components/plugins/source-code.js
```diff
@@ -276,35 +256,6 @@ class SourceCodeDialog extends Localizer(RtlMixin(LitElement)) {
 
 		requestAnimationFrame(() => {
 			this._codeEditor.focus();
-			this._updateCodeEditorTooltipPositioning();
-		});
-	}
-
-	_updateCodeEditorTooltipPositioning() {
```
grant-cleary, line 283: This is exciting, but it appears that CodeMirror has adjusted their tooltip positioning logic to account for transformations on ancestors.

components/plugins/source-code.js
```diff
@@ -1,11 +1,8 @@
 import '@brightspace-ui/core/components/colors/colors.js';
 import '@brightspace-ui/core/components/dialog/dialog.js';
 import { css, html, LitElement } from 'lit';
-import { EditorState, StateEffect } from '@codemirror/state';
```
grant-cleary, line 4: Even more exciting, but because we no longer need to access CodeMirror's state and view objects outside of rendering, we can also switch those two packages to use dynamic imports entirely! :D

package.json
```diff
@@ -61,15 +61,8 @@
...
     "@codemirror/commands": "^0.20",
-    "@codemirror/comment": "^0.19",
-    "@codemirror/fold": "^0.19",
-    "@codemirror/gutter": "^0.19",
-    "@codemirror/highlight": "^0.19",
-    "@codemirror/history": "^0.19",
     "@codemirror/lang-html": "^0.20",
     "@codemirror/language": "^0.20",
-    "@codemirror/matchbrackets": "^0.19",
-    "@codemirror/rectangular-selection": "^0.19",
```
grant-cleary, line 72: As much as it's neat that CodeMirror is so modular, it's also super nice to see some of this more basic functionality compressed into fewer imports.

components/plugins/source-code.js
grant-cleary: Whoops, never mind, this was actually down to a change in our dialogs.

PR #793, opened 2022-04-25
fix: Updating Translations 2205 25/04/2022

grant-cleary:
Translations backport


dmd2l:
I'm assume this has provided from someone who actually speak these? 😬   But looks fine to me assuming translating is correct ...

dmd2l:
> > I'm assume this has provided from someone who actually speak these? 😬 But looks fine to me assuming translating is correct ...
> 
> Yes, the translators are native speakers of the languages they're translating. 😄

I'm assume from this, that translators are who have supplied you these then?  It maybe better having one of those folk on here to review - if something has gotten offset or mixed up they'd at least have hope of spotting that 🤷

grant-cleary:
> I'm assume from this, that translators are who have supplied you these then? It maybe better having one of those folk on here to review - if something has gotten offset or mixed up they'd at least have hope of spotting that 🤷

The process of getting the translations is automated and once we get them the translators don't have any more involvement in the process.

dmd2l:
> > I'm assume from this, that translators are who have supplied you these then? It maybe better having one of those folk on here to review - if something has gotten offset or mixed up they'd at least have hope of spotting that 🤷
> 
> The process of getting the translations is automated and once we get them the translators don't have any more involvement in the process.

I feel if we are require a code review to complete the process, then we must be not entirely  trusting of that automation.  If we don't trust it, then using reviewer who can actually understand if something has gone wrong, seems a more appropriate choice of reviewer

PR #803, opened 2022-04-28
Expose rich context object to insulate editor component functionality from plugins

grant-cleary:
Just a first go at what something like this might look like, hiding the most egregious cases where plugins dig into the component's shadowRoot.


components/context.js
```diff
@@ -0,0 +1,132 @@
...
+
+	get host() {
+		return this._host;
+	}
+
+	set host(host) {
+		this._host = host;
+	}
+
+	get activeOrgEnrollment() {
```
grant-cleary, line 27: This feels not great, but it also feels marginally better than having a sub-object in here to surface `context` directly. I don't mind the idea that anyone adding a new context member should also be required to explicitly surface it somewhere in the editor instead of just calling it and hoping for the best. Basically, this feels like a more stable API to me. 🤷

components/context.js
```diff
@@ -0,0 +1,132 @@
+import { HtmlAttributeObserverController } from '@brightspace-ui/core/controllers/attributeObserver/htmlAttributeObserverController.js';
+
+let context;
+let contextKey;
+const contextAttribute = 'data-he-context';
+
+export class EditorContext {
+
+	constructor(component) {
+		this.component = component;
```
grant-cleary, line 10: General idea is that in a perfect world, this could become completely internal. Right now it isn't because there are a number of places where we fetch/set properties directly and I didn't really want to end up with a gigantic changeset right off the bat.

components/context.js
```diff
@@ -0,0 +1,132 @@
...
+
+		this._contextController = new HtmlAttributeObserverController(component, contextAttribute);
+		this._contextMenuPlugins = new Map();
+		this._host = '';
+
+		this._updateContext(false);
```
grant-cleary, line 16: I can probably write a different function that uses a better name (like `initContext` or something). The general idea is that an editor will only create exactly one of these objects. If this is the first object created, we need to make sure to populate `context` with the initial values surfaced from the context observer.

PR #811, opened 2022-05-02
Maintain cursor position when opening source code editor

grant-cleary:
We ideally want to maintain the cursor position within the HTML when opening the source code editor. This is a bit challenging to do because CodeMirror expects a numerical position representing the number of characters into the HTML string it should move the cursor to, whereas TinyMCE only works with DOM objects when it comes to manipulating cursor position.

How do we solve this? In short, by cheating. We can introduce an invisible span into the HTML at the cursor location at the moment the source code dialog is opened. We can then regex search for that specific span to identify its start position, and then immediately remove it from the editor document's internal DOM and from the HTML string we pass into the editor. It's not elegant, but it works.

I did explore some alternate solutions involving using range objects to recursively walk up the DOM up to the document root, calculating its offset along the way, but of course the inherent problem there is that the browser's DOM parsing ignores irrelevant whitespace while CodeMirror does not.


components/plugins/source-code.js
```diff
@@ -128,13 +161,13 @@ class SourceCodeDialog extends Localizer(RtlMixin(LitElement)) {
...
 
 		this.html = '';
+		this.openPosition = -1;
 		this.opened = false;
 	}
 
 	_handleOpen() {
 		if (this._codeEditor) {
 			this._codeEditor.focus();
-			this._updateCodeEditorTooltipPositioning();
```
grant-cleary, line 137: This was broken, as this function doesn't exist anymore. It wasn't causing actual issues because no JS runs after this, but I'm sure it will be polluting dev instance logs.

components/plugins/source-code.js
```diff
@@ -245,6 +284,9 @@ class SourceCodeDialog extends Localizer(RtlMixin(LitElement)) {
 
 		requestAnimationFrame(() => {
 			this._codeEditor.focus();
+			this._codeEditor.dispatch({
+				scrollIntoView: true
```
grant-cleary, line 288: There's no way to specify this when first creating the editor state. This seems strange, but the explanation is tied into how CodeMirror instantiates. Specifically, there's no requirement to specify an initial `doc` when first instantiating the editor. Ergo, at the time of state creation, there's no guarantee the DOM has been constructed yet so "scroll into view" has no meaning. Instead, it's expected for the consumer to manage this once the editor has been attached to the DOM.

components/plugins/source-code.js
```diff
@@ -7,6 +7,9 @@ import { Localizer } from '../../localizer.js';
 import { requestInstance } from '@brightspace-ui/core/mixins/provider-mixin.js';
 import { RtlMixin } from '@brightspace-ui/core/mixins/rtl-mixin.js';
 
+const cursorId = 'd2l-htmleditor-source-code-cursor-position';
+const cursorMarker = `<span id="${cursorId}" style="display: none;"></span>`;
```
grant-cleary, line 11: I have no attachment to this being a span. It just needs to be something inoffensive to the editor that's valid to introduce virtually anywhere and can take an ID/styles.

grant-cleary: I also suspect this may not be perfect. It wouldn't entirely surprise me if it was possible for a user to get the cursor somewhere that doesn't support any HTML element as a child/sibling. But in that case, I also have no idea what we could do anyways.

One small side effect here is that adding/removing the span does break up text nodes in an odd way if you inspect the DOM. However, the browser treats sequential text node siblings exactly the same as a single continuous text node, so I don't think this can cause issues.

dbatiste (approved):
Seems alright.

PR #814, opened 2022-05-03
Render d2l-dialogs within the editor container

grant-cleary:
Our `d2l-dialog`s have a z-index that's lower than the editor's fullscreen shim, so in situations where we don't use native `dialog` elements, any of our plugins that spawn `d2l-dialog`s will spawn the dialogs behind the editor when we're in fullscreen mode (the dialogs are siblings of the editor container).

Toying with the z-indexes is sketchy, so instead, let's just jam the dialogs into the editor container so their z-index will compute relative to the editor's fullscreen shim.


components/plugins/attributes.js
```diff
@@ -37,7 +37,9 @@ export function addAttributesPlugin() {
 
 		const openAttributesDialog = () => {
 			let dialog = component.shadowRoot.querySelector('d2l-htmleditor-attributes-dialog');
-			if (!dialog) dialog = component.shadowRoot.appendChild(document.createElement('d2l-htmleditor-attributes-dialog'));
+			if (!dialog) dialog = component.shadowRoot.querySelector('.d2l-htmleditor-dialog-container').appendChild(
```
grant-cleary, line 40: This extra query selector isn't ideal. I have a thought on how to make this better (or at least obfuscate this some), but that's a broader refactor and I want to resolve this bug for now.

grant-cleary:
Note that ideally, we should register all dialogs the same way. For the purpose of this change, however, I wanted to restrict things to the dialogs that are actually problematic here. LMS dialogs don't have this problem because their z-index is higher than the fullscreen shim.

htmleditor.js
```diff
@@ -575,7 +575,8 @@ class HtmlEditor extends HtmlEditorMixin(SkeletonMixin(LitElement)) {
 						${renderFooter && !this._fullscreen ? html`<div class=${classMap(shimClasses)}></div>` : ''}
 					</div>
 					${renderFooter ? this._renderFooter() : ''}
-				</div>
+				</div>			
+				<div class="d2l-htmleditor-dialog-container"></div>
```
Trandrew17, line 579: So, we are basically choosing to append the dialog here which has a z-index that is desired whereas before we were just appending it to the root?

grant-cleary: In fullscreen mode, the parent container we're adding this to gets a z-index added. This means that if a dialog is created as a descendant of that container, its own z-index is calculated relative to the container's. Basically even though the dialog still has a z-index that's > the fullscreen editor container's z-index, it still stacks over top because it treats the ancestor container as having a z-index of 0.

grant-cleary:
Alright, it's done. The tests all work now. So I'm going to close this and rebase (or rather, just start from scratch because there are so many commits in here) and then put up another PR.

Trandrew17 (approved):

PR #818, opened 2022-05-05
Don't try to init TinyMCE if we disconnected while adding TinyMCE's scripts

grant-cleary:
This was a fun one. It's possible for the editor to be disconnected from the DOM while it's loading the global TinyMCE scripts. If this happens, we try to init TinyMCE against a `textarea` that isn't attached to the DOM and TinyMCE hates that. Hence, check if we've disconnected after the scripts load, and if so, bail.


htmleditor-mixin.js
```diff
@@ -555,6 +555,9 @@ export const HtmlEditorMixin = superclass => class extends Localizer(LabelledMix
 		await this._loadTinyMCELangs();
 		await this._loadTinyMCE();
 
+		// Editor has been removed from the DOM during TinyMCE script loading. Don't try to initialize.
+		if (this._editorDisconnected) return;
```
grant-cleary, line 559: This is exclusively modified in the `htmleditor` component, which is unfortunate because we're accessing it here in the mixin. In general I think I want to re-evaluate what lives in the mixin and what doesn't because it's getting more and more challenging to keep the two separated.

htmleditor-mixin.js
```diff
@@ -555,6 +555,9 @@ export const HtmlEditorMixin = superclass => class extends Localizer(LabelledMix
 		await this._loadTinyMCELangs();
 		await this._loadTinyMCE();
 
+		// Editor has been removed from the DOM during TinyMCE script loading. Don't try to initialize.
+		if (this._initState === editorInitStates.UNINITIALIZED) return;
```
grant-cleary, line 559: This has the same behaviour as before. When we first start initializing, we change this state to `INITIALIZING`, but when we disconnect from the DOM we reset it to `UNINITIALIZED`. So if we get here and we're uninitialized, it means we have indeed disconnected from the DOM.

grant-cleary: Eh, actually, this does open a slim window in which this timing issue could be reintroduced, so I'm going to walk it back. `_editorDisconnected` is the last thing that gets set when the editor reconnects, so it's the safest thing to check.

PR #832, opened 2022-05-11
Fix flaky visual diff tests

grant-cleary:
Couple things here worth pointing out:
1. I flipped off `captureBeyondViewport` basically everywhere (thanks @dbatiste) to hopefully increase stability across the board.
2. I tried to convert `setTimeout`s to rAFs and rAFs to updateComplete/initializationComplete promises where possible to try to clean things up a bit (hopefully something more stable because of (1).
3. I moved the inline tests into their own fixture because the full tests were mucking up the viewport for Puppeteer's screenshots somehow. I could expend time and energy trying to figure out why, but I don't think that would be time well-spent, given that this does the job just fine and local testing shows that the editors clearly are visible on the page.

In concert with (3) above, I also re-enabled the core editor visual diff tests that were disabled before due to their horrible stability issues. Hopefully the above changes correct that, but I'll need to monitor these for a bit to know for sure.


grant-cleary:
Need a bit more of a wait after marking stuff as invalid, I think.

grant-cleary:
Oh shoot, wrong tests. It's the non-forced-invalid ones, I think because whether they mark invalid or not depends on if focus happens to leave the editor.

grant-cleary:
I don't think those tests really add value, so I'm going to remove them.

Trandrew17 (approved):

PR #839, opened 2022-05-16
Only use offset parent for context menu positioning if it's actually explicitly positioned

grant-cleary:
When the editor is contained within a table, we try to position the context menu relative to the parent table cell (as that's what `getOffsetParent` does if there's a table parent and no other explicitly positioned ancestor). We don't want this behaviour, however, since we're already calculating the document-relative position against the editor iframe. Hence if the editor is contained within a table, we double-count that offset and the context menu gets wildly out of position.

Solution: Make sure the offset parent we've found is _actually_ explicitly positioned before we try using it.


components/context-menu/context-menu.js
```diff
@@ -436,9 +436,17 @@ class ContextMenu extends Localizer(RtlMixin(LitElement)) {
...
+			return position && position !== 'static';
+		}
+
 		let parentTop;
 		let parentLeft;
-		if (offsetParent && offsetParent.tagName !== 'BODY') {
+
+		// We're only concerned with the offset parent if it's actually explicitly positioned, as we already
+		// calculate positioning against the editor's iframe container.
+		if (offsetParent && offsetParent.tagName !== 'BODY' && isExplicitlyPositioned(offsetParent)) {
```
grant-cleary, line 449: I can make this check better by only doing the explicit positioning check if we're dealing with a TH, TD, or TABLE element. Not sure why I didn't just do that on initial PR creation...

components/context-menu/context-menu.js
```diff
@@ -436,9 +436,23 @@ class ContextMenu extends Localizer(RtlMixin(LitElement)) {
...
 		let parentLeft;
-		if (offsetParent && offsetParent.tagName !== 'BODY') {
+
+		// We're only concerned with the offset parent if it's actually explicitly positioned, as we already
+		// calculate positioning against the editor's iframe container.
+		if (offsetParent
+			&& offsetParent.tagName !== 'BODY'
+			&& (
+				(offsetParent.tagName !== 'TH' && offsetParent.tagName !== 'TD' && offsetParent.tagName !== 'TABLE')
+				|| isExplicitlyPositioned(offsetParent)
```
grant-cleary, line 453: This does make me wonder if the offset parent is actually what we want here, but I'm pretty sure it is... The problem we're trying to solve here is a case like (for example) Rubrics, where there's some explicitly positioned ancestor that the bounding rect is being calculated against, hence we need to adjust our own positioning to reflect that ancestor's position.

Trandrew17 (approved):

PR #866, opened 2022-06-07
Learndl component base

lloydwaddell:
**merges into learndl/main**

### Context
Sets up the base component for the new learndl practice. Existing schema components are rendered with save and preview buttons. Template values for answers, attributes, and attribute values are displayed as json.

Out of scope for this PR:
- CSS to be revisited in future PR
- creating html to rendering answers, attributes, and values 

learndl-form-1


components/plugins/practice.js
```diff
@@ -608,6 +609,9 @@ class PracticeDialog extends RequesterMixin(Localizer(LitElement)) {
 
 			case 1:
 				return html`<d2l-sequencing-practice @activityUpdate="${this._activityUpdated}" @activitySave="${this._saveActivity}" @togglePreview="${this._togglePreview}" .activityData="${this.activityData}" ?display-feedback="${this._modifiers[0].value}"></d2l-sequencing-practice>`;
+
+			case 2:
```
louiscoste, line 613: oh... this is weird that it is about the index in the template and not about the type...

lloydwaddell: yeah, that was my initial impression too

components/practice/learndl.js
```diff
@@ -0,0 +1,292 @@
...
+			case 'number':
+				return html`
+				<div class="d2l-form-field">
+					<label class="d2l-input-label ${property.required ? 'd2l-input-label-required' : ''}" aria-hidden="true">${label}</label>
+					${property.help ? html`<d2l-button-icon id="help_${property.key}" class="d2l-help" text="${this.localize('practice:additionalInformation')}" icon="tier1:help" @click="${this._openHelp}"></d2l-button-icon><d2l-dialog title-text="${label} Information"><div class="d2l-help-content">${property.help}</div></d2l-dialog>` : ''}
+					<d2l-input-number id="${id}" data-property="${property.property}" data-type="number" .value="${live(this._getPropertyValue(property.property))}" placeholder="${property.placeholder}" ?required="${property.required === true}" label-hidden label="${label}" @change="${this._change}"></d2l-input-number>
+				</div>
+				`;
+
+			case 'checkbox':
```
louiscoste, line 145: For now could we just remove the cases we do not use? Just to reduce the code duplication as much as possible

components/practice/learndl.js
```diff
@@ -0,0 +1,292 @@
...
+	_showPreview() {
+		this.dispatchEvent(new CustomEvent(
+			'togglePreview', {
+				bubbles: true,
+				composed: true,
+				detail: ''
+			}));
+	}
+
+	_updateGlobalProperties() {
```
louiscoste, line 287: Doesn't look like this gets used

louiscoste (approved):

PR #869, opened 2022-06-07
learndl component 

lloydwaddell:
**merges into learndl/main**

### Context
Learndl main component is now rendered using d2l-table component.

- aria-labels are used for each d2l-input-text
- get / update property methods now support complex properties `answers:{index}:title`

Out of scope
- Add new column
- Add new row function
- better mapping between answer -> attribute values needs to be considered to ensure data integrity


components/practice/learndl.js
```diff
@@ -144,12 +154,36 @@ class LearndlPractice extends RequesterMixin(Localizer(LitElement)) {
...
+								`)}
+							</tr>
+						</thead>
+						<tbody>
+							${this.activityData.answers.map((answer, index) => html`
+								<tr>
+									<td>
+										<d2l-input-text id="${getUniqueId()}" data-property="answers:${index}:title" data-type="text" .value="${answer.title}" ?required="true" aria-label="${this.localize('practice:learndl:answers')}" @change="${this._change}"></d2l-input-text>
+									</td>
+									${this.activityData.attributeValues.map((attrValue, innerIndex) => {
```
lloydwaddell, line 175: I don't like this, there is no guarantee that attributeValues is in order. Potentially, this could be mapped on `attributes` and we search by answerId and attributeId. However, I'm leaning towards adding an additional property to `answers` for attributes.

Something like `answers.attributes[attributeId]` would ensure adding / deleting attributes would keep the table in sync.

components/practice/learndl.js
```diff
@@ -144,12 +154,36 @@ class LearndlPractice extends RequesterMixin(Localizer(LitElement)) {
...
-				<h4>Attribute Values</h4>
-				<div><pre>${JSON.stringify(this.activityData.attributeValues, null, 2)}</pre></div>
+				<d2l-table-wrapper type="light" no-column-border>
+  					<table class="d2l-table">
+						<thead>
+							<tr>
+								<th>${this.localize('practice:learndl:answers')}</th>
+								${this.activityData.attributes.map((attr, index) => html`
+									<th>
+										<d2l-input-text id="${getUniqueId()}" data-property="attributes:${index}:title" data-type="text" .value="${attr.title}" ?required="true" aria-label="${this.localize('practice:learndl:attributes')}" @change="${this._change}"></d2l-input-text>
```
louiscoste, line 164: nit:
```suggestion
										<d2l-input-text id="attribute-${attr.id}" data-property="attributes:${index}:title" data-type="text" .value="${attr.title}" ?required="true" aria-label="${this.localize('practice:learndl:attributes')}" @change="${this._change}"></d2l-input-text>
```

components/practice/learndl.js
louiscoste: I was about to make the same suggestion. If we move the attributes into the answers object then it closely ties the attribute values to the answers. Which will make it easier for the editing and validating the answers end users provide

components/practice/learndl.js
```diff
@@ -144,12 +154,36 @@ class LearndlPractice extends RequesterMixin(Localizer(LitElement)) {
...
+									</th>
+								`)}
+							</tr>
+						</thead>
+						<tbody>
+							${this.activityData.answers.map((answer, index) => html`
+								<tr>
+									<td>
+										<d2l-input-text id="${getUniqueId()}" data-property="answers:${index}:title" data-type="text" .value="${answer.title}" ?required="true" aria-label="${this.localize('practice:learndl:answers')}" @change="${this._change}"></d2l-input-text>
```
louiscoste: nit:
```suggestion
										<d2l-input-text id="answer-${answer.id}" data-property="answers:${index}:title" data-type="text" .value="${answer.title}" ?required="true" aria-label="${this.localize('practice:learndl:answers')}" @change="${this._change}"></d2l-input-text>
```

components/practice/learndl.js
lloydwaddell: I'm going to do this in a follow up PR, its being problematic.

components/practice/learndl.js
lloydwaddell: these should all be fixed now

louiscoste (approved):

PR #871, opened 2022-06-08
Updating Learndl data structure

lloydwaddell:
**merges into learndl/main**

### Context
The answer objects in d2l-practices now contain their own attribute values rather than being in a separate array. 

- Updated renderer to fetch values from their new home
- Method `_updateValue` now calls a recursive function for setting properties, old method was hard coded to a depth of 3.
- Cleaned up unnecessary data properties
- Removed lang term that is no longer used

### Quality
Verified values are being saved to `activityData`

**Attributes:**
attributes-save

**Answers:**
save-answers

**Values:**
save-values


components/practice/d2l-practices.js
```diff
@@ -287,28 +287,46 @@ const practiceTemplates = function(localize) {
...
-					{ 'id':0, 'attributeId':3, 'answerId': 2, 'value': localize('practice:learndl:attributeValues'), 'correctFeedback': '', 'incorrectFeedback': '' },
-					{ 'id':0, 'attributeId':0, 'answerId': 3, 'value': localize('practice:learndl:attributeValues'), 'correctFeedback': '', 'incorrectFeedback': '' },
-					{ 'id':0, 'attributeId':1, 'answerId': 3, 'value': localize('practice:learndl:attributeValues'), 'correctFeedback': '', 'incorrectFeedback': '' },
-					{ 'id':0, 'attributeId':2, 'answerId': 3, 'value': localize('practice:learndl:attributeValues'), 'correctFeedback': '', 'incorrectFeedback': '' },
-					{ 'id':0, 'attributeId':3, 'answerId': 3, 'value': localize('practice:learndl:attributeValues'), 'correctFeedback': '', 'incorrectFeedback': '' },
+					{
+						'id': 0,
+						'value': localize('practice:learndl:answers'),
+						'attributes': {
+							'a0': { 'attributeId': 0, 'value': '' },
```
lloydwaddell, line 294: key is formatted `a{attr.id}`

components/practice/d2l-practices.js
```diff
@@ -287,28 +287,46 @@ const practiceTemplates = function(localize) {
 					{ 'id': 3, 'value': localize('practice:learndl:attributes'), 'type': 0 }
 				],
 				'answers':[
-					{ 'id':0, 'value': localize('practice:learndl:answers'), 'correctFeedback': '', 'incorrectFeedback': '' },
-					{ 'id':1, 'value': localize('practice:learndl:answers'), 'correctFeedback': '', 'incorrectFeedback': '' },
-					{ 'id':2, 'value': localize('practice:learndl:answers'), 'correctFeedback': '', 'incorrectFeedback': '' },
-					{ 'id':3, 'value': localize('practice:learndl:answers'), 'correctFeedback': '', 'incorrectFeedback': '' }
-				],
-				'attributeValues': [
-					{ 'id':0, 'attributeId':0, 'answerId': 0, 'value': localize('practice:learndl:attributeValues'), 'correctFeedback': '', 'incorrectFeedback': '' },
```
lloydwaddell, line 296: IMO having the attributeValue as a default value is really annoying

components/practice/d2l-practices.js
louiscoste: Why the a in the key?
Why not just the id? The id should be unique so there should not be a problem with collision with keys

components/practice/learndl.js
```diff
@@ -243,14 +245,23 @@ class LearndlPractice extends RequesterMixin(Localizer(LitElement)) {
 	_updateValue(prop, value) {
 		if (prop.includes(':')) {
 			const propArr = prop.split(':');
-			this.activityData[propArr[0]][propArr[1]][propArr[2]] = value;
+			this._updateValueRecursive(this.activityData, propArr, value);
```
louiscoste, line 248: Why not also implement a for loop kind of solution like you did for `_getPropertyValue`

lloydwaddell: The for loop in _getPropertyValue passes by value and not reference, that was my initial approach. Passing the object to a method preserves the reference.

components/practice/d2l-practices.js
louiscoste: I would be surprised if we get an issue with that. I was just able to create an object with numeric properties on the console and it is behaving as I would expect it to. Do you remember what kind of issues you ran into?

lloydwaddell: no, I don't unfortunately, so it may not be an issue here. Having just a number as a property (in my mind variable) seems invalid to me.

lloydwaddell: another option could be `_{attr.id}`, since that would also be a valid variable name. 🤷 I'll let you be the judge 😄

louiscoste: Let's keep it as a number for now, if we run into issues because of that then it should not be too hard to fix it

lloydwaddell: updated in last commit

louiscoste (approved):

PR #872, opened 2022-06-08
Lwaddell/add rows cols

lloydwaddell:
**merges into learndl/main**

### Context
Implement add and remove buttons for both answers and attributes.

add-remove-component

add-remove-data


components/practice/learndl.js
```diff
@@ -115,6 +117,42 @@ class LearndlPractice extends RequesterMixin(Localizer(LitElement)) {
 		`;
 	}
 
+	_addAnswer() {
+		this.nextAnswerId = this.nextAnswerId ?? Math.max(...this.activityData.answers.map((answer) => answer.id));
+
+		this.activityData.answers.push({
+			'id': ++this.nextAnswerId,
+			'value': this.localize('practice:learndl:answers'),
+			'attributes': {
```
louiscoste, line 126: This is assuming a certain number of attributes

lloydwaddell: right, good catch

louiscoste (approved):

PR #878, opened 2022-06-10
Added input validation

louiscoste:
- Set answer and attribute default value to empty string
- Added check if entered answer or attribute is a duplicate
- Not saving duplicate values

image


components/practice/learndl.js
```diff
@@ -248,6 +279,26 @@ class LearndlPractice extends RequesterMixin(Localizer(LitElement)) {
...
+				this._updateActivity();
+				return;
+			} else {
+				this.activityData.answers[prop.split(':')[1]].duplicate = false;
+			}
+		}
+		if (event.target.classList.contains('learndl-attribute')) {
+			if (this.activityData.attributes.find(attribute => attribute.value.toLowerCase() === value.toLowerCase())) {
+				event.target.setAttribute('aria-invalid', true);
+				this.activityData.attributes[prop.split(':')[1]].duplicate = true;
```
lloydwaddell, line 294: if you split property at the start you could combine the duplicate checks
```suggestion
				this.activityData[propArr[0]][propArr[1]].duplicate = true;
```

lang/en.js
```diff
@@ -187,6 +187,8 @@ export default {
 	"practice:learndl:attributes": "Attribute",
 	"practice:learndl:deleteAnswer": "Delete Answer",
 	"practice:learndl:deleteAttribute": "Delete Attribute",
+	"practice:learndl:duplicateAnswer": "Duplicate answer",
+	"practice:learndl:duplicateAttribute": "Duplicate attribute",
```
lloydwaddell, line 191: nit: for consistency, though I fully expect there to be some kind of audit on these terms down the road. 
```suggestion
	"practice:learndl:duplicateAnswer": "Duplicate Answer",
	"practice:learndl:duplicateAttribute": "Duplicate Attribute",
```

components/practice/learndl.js
louiscoste: Good catch

lloydwaddell (approved):
looks good, the en.js update will need to be updated in all files

PR #882, opened 2022-06-13
Added types

louiscoste:
image


components/practice/learndl.js
```diff
@@ -207,6 +207,16 @@ class LearndlPractice extends RequesterMixin(Localizer(LitElement)) {
 												${this.localize('practice:learndl:duplicateAttribute')}
 											</d2l-tooltip>
 										` : html``}
+										<label>
+										<span class="d2l-input-label">${this.localize('practice:learndl:typeLabel')}</span>
+											<select
+												class="d2l-input-select"
+												data-property="attributes:${index}:type"
+												@change="${this._change}">
+												<option value="text" ?selected=${attr.type === 'text'}>${this.localize('practice:learndl:textType')}</option>
```
lloydwaddell, line 216: if we ever add more types, we might want to switch to an enum and iterate the types here.

lloydwaddell (approved):

components/practice/learndl.js
louiscoste: Ideally I would there to be only one enum/list of types for both this code and the iframe

PR #888, opened 2022-06-15
Improved import/export

louiscoste:


components/practice/learndl.js
```diff
@@ -363,7 +363,7 @@ class LearndlPractice extends RequesterMixin(Localizer(LitElement)) {
 
 		stringifier.write(
 			this.activityData.attributes.reduce(
-				(row, attribute) => [...row, attribute.value],
+				(row, attribute) => [...row, attribute.value + (attribute.type !== 'text' ? `:${attribute.type}` : '')],
```
lloydwaddell, line 366: nit: I think this should be delimited to prevent collisions, but I doubt it would cause issues right now.

PR #889, opened 2022-06-15
Added option to make user guess attributes instead of answers

louiscoste:


components/practice/learndl.js
```diff
@@ -48,7 +48,10 @@ class LearndlPractice extends RequesterMixin(Localizer(LitElement)) {
 			.d2l-title,
 			.d2l-short-description,
 			.d2l-instructions,
-			.d2l-input-checkbox {
```
lloydwaddell, line 51: was this intentional? we are still using checkboxes

louiscoste: actually I should be adding a class name to all of those that is common and change the css to only do this for that new class name

lang/en.js
```diff
@@ -184,7 +184,7 @@ export default {
 	"practice:learndl:allowInvalidAnswers": "Allow Invalid Answers",
 	"practice:learndl:answers": "Answers",
 	"practice:learndl:autocomplete": "Display Auto Completion",
-	"practice:learndl:attributes": "Attribute",
+	"practice:learndl:attributes": "Attributes",
```
lloydwaddell, line 187: think this needs to be manually updated in each lang file

louiscoste: No need to do that. the linter just wants `practice:learndl:attributes` in all the lang files. It does not matter what the text is because they will get translated anyways

lloydwaddell: 🤦‍♂️ fair

lloydwaddell (approved):

PR #950, opened 2022-08-05
Fix inconsistencies in accessing and saving code snippets

grant-cleary:
Two things are addressed in this PR:
1. Hitting enter when a code element was selected would wipe it out instead of opening the overlay.
2. Zeroing out code and then saving it was a no-op. It's expected that we would delete the snippet instead.


components/plugins/insert-code.js
```diff
@@ -178,7 +178,7 @@ export function addInsertCodePlugin() {
 			overlay.showing = true;
 
 			overlay.addEventListener('d2l-htmleditor-insertcode-overlay-close', async(e) => {
-				if (e.detail && e.detail.code && e.detail.lang) {
+				if (e.detail && e.detail.code !== undefined && e.detail.lang !== undefined) {
```
dbatiste, line 181: If I follow this correctly, `e.detail.code === null` when the user has closed the overlay without any code in it. Is that correct?

grant-cleary: No, it will be `''`. This condition shouldn't happen unless we somehow have the wrong event or something, so in that sense I think it's sort of redundant.

dbatiste: Cool.  Looks good.  Thanks for the clarification on the zero-length string resolving to `true` below in the conditional.

dbatiste (approved):

PR #981, opened 2022-09-02
Retrieve lang attribute and pass to practice learner view

rileyodonnelld2l:
This is necessary in order for localization in the learner view of a Practice to be pulled from the locale of the htmleditor.
PR for learner view: https://github.com/Brightspace/ls-practice-activity/pull/71


components/plugins/practice.js
```diff
@@ -55,12 +55,15 @@ export function addPracticePlugin() {
...
 			const context = requestInstance(editor.getElement(), 'context');
 			const component = context.component;
+			const editorDocument = new DOMParser().parseFromString(editor.getContent(), 'text/html');
 
 			let dialog = component.shadowRoot.querySelector('d2l-practice-dialog');
 			if (!dialog) dialog = component.shadowRoot.querySelector('.d2l-htmleditor-dialog-container').appendChild(
 				document.createElement('d2l-practice-dialog')
 			);
 
+			dialog._lang = editorDocument.querySelector('html').getAttribute('lang') ? editorDocument.querySelector('html').getAttribute('lang') : 'en-us';
```
grant-cleary, line 65: This can be simplified to:
```
dialog._lang = editorDocument.querySelector('html').getAttribute('lang') || 'en-us';
```
which would also avoid needing to invoke the selector twice.

grant-cleary: Also, nit: If `_lang` is being invoked like this, it should probably be public (i.e. `lang`).

components/plugins/practice.js
```diff
@@ -297,6 +300,7 @@ class PracticeDialog extends RequesterMixin(Localizer(LitElement)) {
 			_activityDataURL: { type: String },
 			_changeActivityTarget: { type: Object },
 			_context: { type: Object },
+			_lang: { type: String },
```
grant-cleary, line 303: You'll also need to update the property name here, in the constructor, and where it's used later on.

PR #990, opened 2022-09-12
Added in encoding / decoding of angled brackets for text inputs in In…

jschlumkoski:
…sert Element


components/plugins/lcs.js
```diff
@@ -740,6 +744,10 @@ class ElementEditorDialog extends Localizer(LitElement) {
 		this._loadedHTML = false;
 	}
 
+	_decodeString(str) {
+		return str.replace(/&lt;/g, '<').replace(/&gt;/g, '>');
```
grant-cleary, line 748: Is it only these characters you need to encode/decode? `encodeURIComponent` and `decodeURIComponent` are fairly robust, from what I've seen.

jschlumkoski: @grant-cleary I had initially tried doing this, but when encoded like this, these characters don't actually render properly on the page (see image below). This is what led me to go the HTML Entity route. For right now I only need these characters, but it may be expanded on in the future!

image

grant-cleary: Yeah, I guess using those functions would also assume these bits were always URI-encoded/decoded throughout the render chain.

grant-cleary (approved):

PR #1005, opened 2022-09-26
Remove redundant dev dependencies

grant-cleary:
Pruning out some peer dependencies we don't explicitly need here.


package.json
```diff
@@ -37,20 +37,13 @@
...
   "devDependencies": {
-    "@babel/eslint-parser": "^7",
     "@brightspace-ui/stylelint-config": "^0.7",
     "@open-wc/testing": "^3",
     "@web/dev-server": "^0.1",
     "@web/test-runner": "^0.14",
-    "babel-eslint": "^10",
     "chalk": "^5",
     "deepmerge": "^4.2.2",
-    "eslint": "^8",
```
dbatiste, line 48: There is an argument to be made that `eslint` itself should still be referenced since it's call directly.

grant-cleary: Ah yeah, I suppose it is.

PR #1016, opened 2022-10-06
Insert Element Check Page Validity

jschlumkoski:
Prevent Insert Element from opening (and showing a fallback message) if the content document contains CSS/scripts that aren't from the Creator+ Template or Practices.


components/plugins/lcs.js
```diff
@@ -111,6 +111,32 @@ export function addElementEditorDialogPlugin() {
 			}
 		};
 
+		const checkPageValidity = () => {
+			const editorDocument = new DOMParser().parseFromString(editor.getContent(), 'text/html');
+			const safelistCSS = /(https:\/\/templates.lcs.brightspace.com\/.*\/.*.css)|(https:\/\/practices.lcs.brightspace.com\/.*\/.*.css)/;
```
grant-cleary, line 116: This will fail if a Content Styler override is set (as an FYI).

jschlumkoski: @grant-cleary Content Styler used to attach a `data-override="override"` attribute to it's link tag because we needed something we could query for this (and practices) to make sure our CSS was added before the overrides (so they worked properly). In the check against the link tags, I only check tags that don't have that attribute value.

Is this not a thing anymore?

grant-cleary: Ah okay, no, that's fine then.

grant-cleary (approved):

PR #1027, opened 2022-10-13
Practice Updates

jschlumkoski:
* Modifying the valence controller logic to incorporate new parameter needed for Insert Element
* Modified urls to point into the new CDN locations


jschlumkoski:
@grant-cleary Just as a note about swapping the URLS, we have redirects set up on the old paths to the new paths, so anything published with the old URLs will still function!

components/plugins/practice.js
```diff
@@ -514,6 +515,8 @@ class PracticeDialog extends RequesterMixin(Localizer(LitElement)) {
 
 		this._context = this.requestInstance('context');
 
+		console.log(this._context);
```
grant-cleary, line 518: I think this made it in by accident.

grant-cleary: Hmm, did you push that change yet? I don't see a new commit.

jschlumkoski: It's pushed now oops

grant-cleary (approved):

PR #1036, opened 2022-10-19
Jprouse/telemetry events

JakeProuse:
https://trello.com/c/s6BoADic/133-data-creator-data


tonystecca (approved):

components/plugins/practice.js
```diff
@@ -59,6 +60,8 @@ export function addPracticePlugin() {
 			const component = context.component;
 			const editorDocument = new DOMParser().parseFromString(editor.getContent(), 'text/html');
 
+			sendOpenPractiseEditorEvent(context);
```
jschlumkoski, line 63: Probably better for this to be `sendOpenPracticeEditorEvent()`

jschlumkoski, line 63: Probably better for this to be sendOpenPracticeEditorEvent() for consistency

package.json
```diff
@@ -67,6 +67,7 @@
     "@codemirror/state": "^6",
     "@codemirror/view": "^6",
     "@lezer/highlight": "^1",
+    "d2l-telemetry-browser-client": "^1.1.0",
```
grant-cleary, line 70: @dbatiste, just wondering if Gaudi has any thoughts/context on this, since it doesn't appear that any of our core components use this library. I'm not sure if there's a standard we're looking for with component telemetry.

dbatiste: It sounds like this is indeed the correct library to use. It's not used in our core components, but it's possible we might want to one day.

We've wanted the ability to gather client-side telemetry events from the editor for a long time, to inform some of our design decisions. Like maybe we would want to know what languages people use for code samples, and do they choose light or dark theme. I think they require a bit of design thought to determine the right events to collect. It would be possible to gather some metrics from database queries and IIS logs, but not always. 

Nevertheless, currently the `telemetry.js` files contains a bunch of stuff that is specific to Creator+, Insert Element, and Practice Questions, even though it is being defined as a plugin such that it seems as though it should be generic. We should consider how we would want this factored to support telemetry events not just from these two plugins.

I'm not super familiar with the client and the data that's passed to it. There might be certain data/naming conventions we would want to follow to filter/group certain events.

JakeProuse: The data that would change between the creator+ events and some other htmleditor events would be the sourceId in the `sendTelemetryEvent` function. Other than that they'll likely pass data in a similar format like a JSON object, key-value pair `{ name: "", value: "" }` or one of the client's setters (I.e. `setAction`, `setUser`...). So there is definitely some common logic that could be seperated and kept in a more generic `telemetry.js` file (a stripped down `sendTelemetryEvent` and some common action language).

Would an events folder to store the events for each tool, and a generic handler in the plugins/telemetry file work? I.e events/creator-plus/insertElement for the IE events.

grant-cleary: I would personally be fine with (and would prefer) a structure like that. I don't really have any comment on what's actually expected by the client (w.r.t. data structure and the like), as I'm not super familiar with that bit.

grant-cleary (approved):

grant-cleary (approved):

package.json
JakeProuse: @grant-cleary Should I merge the PR or does @dbatiste have any comments?

dbatiste (approved):

PR #1047, opened 2022-10-25
Updating Translations 2211 10/25/2022

scowing:
Updating Translation Packs


lang/hi-in.js
```diff
@@ -0,0 +1,423 @@
+/* eslint quotes: 0 */
```
dbatiste, line 1: As mentioned in this [Slack thread](https://d2l.slack.com/archives/C0TF4KSGZ/p1666719158315249), this file was created in error due to incorrect configuration that should be corrected for next time. Any translations that we were expecting might be applied in this file, and we'll probably want them copied over to `hi.js`, but we won't want to merge this particular file.

grant-cleary: Yeah, I'll fix this up in a bit.

grant-cleary: Actually, I'm going to get to this tomorrow. Been a bit distracted today with some personal business I needed to take care of.

grant-cleary: Posted in that thread, but I think some terms were just missed for translation in the hi-in file as well.

grant-cleary (approved):

PR #1061, opened 2022-11-07
Migrate/remove legacy icons to avoid referencing static paths

grant-cleary:
These static image paths are going away, so we need to migrate editor usages away from them. I'll add notes inline in the PR to indicate what I'm doing, but broadly speaking, two areas were affected:
1. Error handling when using WIRIS to generate image previews.
2. A placeholder image used when rendering XSPL-protected iframes within the editor view.


components/plugins/xspl.js
```diff
@@ -24,30 +24,42 @@ export function convertIframeToImagePlaceholder(htmlDoc, host) {
 		// If there are nested iframes, ensure we replace them all recursively.
 		tempDoc.body.querySelectorAll('iframe.d2l-xspl-box').forEach(convertNestedIframes);
 
-		const img = document.createElement('img');
-		img.style = 'border:1px solid #333399;background-color: #ffffcc;padding: 75px 153px;';
-		img.className = 'd2l-xspl-image-placeholder';
-		img.src = `${host || ''}/d2l/img/lp/xspl/html_editor_media.png`;
-		img.dataset.d2lXsplOriginalHtml = tinymce.html.Entities.encodeAllRaw(tempDoc.body.innerHTML);
+		// SVG mark-up for the current insert-stuff icon. Suitable as a placeholder, but we
+		// may revisit this with a new icon in the future if necessary.
```
grant-cleary, line 28: This one's more thorny because we can't just nuke it. Instead, we (perhaps temporarily) just manually build an SVG that maintains roughly the original placeholder style, but uses the ISF icon. If we need a custom icon eventually, we can always address that in the future.

dbatiste: Is it not possible to call `svg.innerHTML = '...'`? Or did you not want to do it that way?

grant-cleary: I could, yeah, but I didn't feel there was a particular benefit to that approach. Either way involves some big blobs of text.

dbatiste (approved):
Seems reasonable. Building the `svg` is icky... too bad we can't write is as a `template` and just use `template.content.cloneNode(true)`.

components/plugins/xspl.js
dbatiste: Yep. Some icky code somewhere no matter what I think.

grant-cleary:
> Seems reasonable. Building the `svg` is icky... too bad we can't write is as a `template` and just use `template.content.cloneNode(true)`. Could the module...
> 
> ```js
> const svgTemplate = document.createElement('template');
> svgTemplate.innerHTML = '<svg ....>...</svg>';
> ```
> 
> Then later go...
> 
> ```js
> const svg = svgTemplate.content.cloneNode(true);
> ```

Mm, good question. I can see if this works.

dbatiste:
>Mm, good question. I can see if this works.

Like you said though, there will still be some ugly code either way.

grant-cleary:
> 
> Like you said though, there will still be some ugly code either way.

Yeah, I made a change that I think makes it a bit cleaner, though I opted to keep the XSPL-specific styles and attributes as separate additions after the fact, since I think it's easier to understand what's going on that way.

components/plugins/xspl.js
```diff
@@ -24,30 +24,35 @@ export function convertIframeToImagePlaceholder(htmlDoc, host) {
...
+		const svgTemplate = document.createElement('template');
+		svgTemplate.innerHTML = '<svg xmlns="http://www.w3.org/2000/svg" width="18" height="18" viewBox="0 0 18 18"><path d="M1 8a1 1 0 01-1-1V1A1 1 0 011.447.1l6 3a1 1 0 010 1.789l-6 3A1 1 0 011 8zM12 8a1 1 0 01-1-1V1a1 1 0 012 0v6a1 1 0 01-1 1zM16 8a1 1 0 01-1-1V1a1 1 0 012 0v6a1 1 0 01-1 1zM16 18h-4a2 2 0 01-2-2v-4a2 2 0 012-2h4a2 2 0 012 2v4a2 2 0 01-2 2zM4 18a4 4 0 112.828-1.172A4 4 0 014 18zm0-6a2 2 0 102 2 2 2 0 00-2-2z" fill="#494c4e"/></svg>';
```
dbatiste, line 30: I was thinking this could be done once someplace. If we're gonna construct the `template` every time, might as well just use a `div` and transplant the `svg` rather than incurring the cloning cost.

grant-cleary: Hmm, well at the very least it has to be somewhere editor-specific because this gets created in the context of TinyMCE's iframe. I suppose it could be created when TinyMCE inits?

htmleditor-mixin.js
```diff
@@ -720,11 +720,16 @@ export const HtmlEditorMixin = superclass => class extends Localizer(LabelledMix
 					this._handleFocusIn();
 				});
 
+				// SVG mark-up for the current insert-stuff icon. Suitable as a placeholder, but we
+				// may revisit this with a new icon in the future if necessary.
+				const xsplSvgTemplate = document.createElement('template');
+				xsplSvgTemplate.innerHTML = '<svg xmlns="http://www.w3.org/2000/svg" width="18" height="18" viewBox="0 0 18 18"><path d="M1 8a1 1 0 01-1-1V1A1 1 0 011.447.1l6 3a1 1 0 010 1.789l-6 3A1 1 0 011 8zM12 8a1 1 0 01-1-1V1a1 1 0 012 0v6a1 1 0 01-1 1zM16 8a1 1 0 01-1-1V1a1 1 0 012 0v6a1 1 0 01-1 1zM16 18h-4a2 2 0 01-2-2v-4a2 2 0 012-2h4a2 2 0 012 2v4a2 2 0 01-2 2zM4 18a4 4 0 112.828-1.172A4 4 0 014 18zm0-6a2 2 0 102 2 2 2 0 00-2-2z" fill="#494c4e"/></svg>';
```
grant-cleary, line 726: Eh, now that I look at this, I'm not clear why this works? How long do variables created inside an arrow function actually stay in scope?

PR #1066, opened 2022-11-10
Upgrade TinyMCE to 6.2.0

grant-cleary:
We want to bump TinyMCE to latest, which is now v6.2.0. This is a major version bump, so there are some inherent risks here, but I believe they're relatively limited in terms of where they overlap with our usage of their APIs.

Things we need to care about specifically here:
- fullpage plugin - This has been removed and we need to write our own.
- hr, noneditable plugins - These are now core so we can just remove them from our plugin list
- textpattern plugin - This is core, so remove it from our plugin list; also need to rename the associated init option
- imagetools plugin - Renamed to editimage (incl. all associated init options)
- Image upload API - Now promise-based instead of callback-based
- Plugin list can now be an array of plugin names instead of a string


components/plugins/image.js
```diff
@@ -48,54 +48,54 @@ export function getImage(editor, src) {
 	}
 }
 
-export function uploadImage(editor, blobInfo, success, failure) {
+export function uploadImage(editor, blobInfo) {
```
grant-cleary, line 51: This is easier to parse with whitespace ignored.

grant-cleary:
Visual diff failures are interesting...

grant-cleary:
Oop, and I think the attributes stuff needs some modifications given the fullpage changes (hence the CI failure).

components/plugins/formatting.js
```diff
@@ -10,6 +11,14 @@ export function addFormatPlugin() {
 		const context = requestInstance(editor.getElement(), 'context');
 		const component = context.component;
 
+		// TinyMCE 5 reflected the currently selected tag bag when querying the formatBlock command value.
+		// This no longer appears to be the case, so we need to do this ourselves for the format selector to work.
+		editor.addQueryValueHandler(cmds.formatBlock, () => {
```
grant-cleary, line 16: Apparently this is an undocumented change in TinyMCE 6? Might be a bug, so I'll submit a ticket asking as much.

grant-cleary:
This won't be merged until next Monday.

Trandrew17 (approved):

PR #1083, opened 2022-11-22
Stop content styler from adding urls to pages with extraneous CSS links

jschlumkoski:


components/plugins/contentstyler.js
```diff
@@ -4,40 +4,52 @@ const name = 'd2l-contentstyler';
...
 export function addContentStylerPlugin() {
 	tinymce.PluginManager.add(name, function(editor) {
-
 		const styleUrl = 'https://templates.lcs.brightspace.com/lib/assets/css/styles.min.css';
+		const safelistCSS = /(https:\/\/templates.lcs.brightspace.com\/.*\/.*.css)|(https:\/\/practices.lcs.brightspace.com\/.*\/.*.css)|(https:\/\/s.brightspace.com\/.*\/.*.css)/;
+
+		editor.on('SetContent', () => {
+			let requiresContentUpdate = false;
+
+			const editorDocument = new DOMParser().parseFromString(editor.getContent(), 'text/html');
```
grant-cleary, line 13: If you're going to do something like this, I would suggest using `editor.getDoc()` to fetch the editor's iframe document instead of `getContent`. The problem with using `getContent` is that it triggers any existing `GetContent` handlers. There are already workflows that trigger both `SetContent` and `GetContent` handlers (like opening the source code editor), so you sort of run the risk of creating a loop by doing this.

As a general rule, `GetContent` is invoked whenever HTML is serialized from the editor document into raw HTML; think of it as moving from the TinyMCE iframe to the actual HTML document we expect to save into a file. `SetContent` is invoked whenever we're going the other way around.

jschlumkoski: @grant-cleary I can't do this, because we're modifying whatever content we retrieve, and setting the editor to use that. This came up during Insert Element I believe because injecting CSS wasn't working properly.

jschlumkoski: If I do this, it injects all sorts of TinyMCE markup 😬

grant-cleary: Well for example, if you look at what this is doing, you're calling `setContent` in a `SetContent` handler, which immediately means this handler is always going to run twice any time the event is invoked. The second time shouldn't need to do any kind of update, but that's already a sign that we're misusing TinyMCE's events here. `SetContent` handlers shouldn't be calling `setContent` because that already creates a loop.

grant-cleary: One thing to note is that `GetContent` won't always return the entire editor document; nor should it. `GetContent` is invoked whenever any portion of the document is serialized, which includes small actions like changing the formatting of a selection. In a case like this, there shouldn't be a need to re-add the CSS, so I think it's reasonable to suggest that if, for example, you can't find the document head using some kind of regex on the event context, you can just bail without doing anything.

jschlumkoski: @grant-cleary With your second comment, it's the piece about how it doesn't return the entire document which is giving me problems when trying to stop the CSS from being injected when a template is added. `GetContent` fires after you add the template page, but since it doesn't return any sort of information about what's currently in the head of the document, you can't ever rely on it for the check because it will always fail to find extraneous CSS files we are checking for.

And I don't think that bailing without doing anything is viable, because there are situations where even if you added a template, you'd technically still want the plugin to do something. The immediate use case is that if someone adds a Creator+ template page, we still need the plugin to add the override CSS file if it exists. If we were disregard things if the head isn't included in `e.content`, then it seems like this will never work, because I've never had `e.content` give me anything other than an empty string when a template page is added.

grant-cleary: So yes, all of this is true because `GetContent` wouldn't be the appropriate event to listen to for this case, because it's the `SetContent` event that primarily gets fired when content is added to the editor (including when adding a new template).

The thing is that if a template is added to the editor, effectively setting the html in its entirety, we're going to get a `BeforeSetContent` event that _does_ include the entire HTML document in its event context, so that's the event we should be listening to. That also allows us to early bail, let's us avoid the use of the second `setContent`, and injects the CSS as appropriate.

I'll send you sample code that works with a quick mock-up in the demo; you may want to test it with the template but I believe it should also work there.

grant-cleary (approved):

PR #1085, opened 2022-11-23
Cherry pick + stripping out new fullpage import

jschlumkoski:


components/plugins/contentstyler.js
```diff
@@ -4,40 +4,54 @@ const name = 'd2l-contentstyler';
...
+			if (requiresContentUpdate) e.content = `<!DOCTYPE html>${editorDocument.documentElement.outerHTML}`;
 		});
-
 	});
-	/* tinyMCE's fullpage plugin MUST be spcified before this plugin to ensure it's GetContent
-	handler runs before this one, otherwise the html document created by this plugin will
-	be placed within the body of the document generated by tinyMCE's plugin */
-	return `fullpage ${name}`;
+
+	return name;
```
grant-cleary, line 56: We probably don't want this change. Or at least if it wasn't broken before, we probably don't want to play with fire by changing it, haha.

jschlumkoski: Got it, I'll fix it in both branches!

grant-cleary: Chatted offline and Justin found that it _did_ break stuff to keep this in there.

grant-cleary (approved):

PR #1087, opened 2022-11-23
Directly inject MathJax into TinyMCE's iframe for math rendering

grant-cleary:
With this PR, we now inject MathJax straight into TinyMCE's iframe for rendering equation previews, instead of using WIRIS' web services to render image previews and fetch alt text for them.


components/toolbar/menu-item.js
```diff
@@ -86,7 +86,7 @@ class MenuItem extends MenuItemSelectableMixin(ToolbarItemMixin(LitElement)) {
 
 		const editor = await getEditor(this);
 		editor.on('NodeChange', () => {
-			this._updateSelected(editor);
+			requestAnimationFrame(() => this._updateSelected(editor));
```
grant-cleary, line 89: There was a timing issue here that crops up when the `NodeChange` event fires on a noneditable element. Because TinyMCE loses its internal selection when the editor blurs while a noneditable element is selected, we need to manually move it back. This means we also need to wait a frame after this happens to make sure we get the right data back.

What I _don't_ like about this is that it might create queueing weirdness if a user is clicking around super rapidly. Because we only apply this within menus, which can't be open while a user is interacting with the editor window, I believe we're fine as long as the _last_ processed event corresponds to the last click.

grant-cleary: I'm open to alternate suggestions here.

components/plugins/equation.js
```diff
@@ -85,127 +90,68 @@ export function addEquationPlugin() {
...
-					text = response.ok ? await response.text() : '';
-				}
+				node.parentElement.insertBefore(mathContainer, node);
+				mathContainer.appendChild(node);
 
-				imgNode.setAttribute('alt', text);
+				await typesetMath(mathContainer, {
+					deferTypeset: true,
+					renderLatex: false,
+					outputScale: context.mathRenderScale || 1,
```
grant-cleary, line 136: This isn't yet provided, but it will be.

components/plugins/equation.js
```diff
@@ -85,127 +90,68 @@ export function addEquationPlugin() {
...
 				let math = match.match(/<math[^>]*>(.|\n|\r)*<\/math>/gi);
 				if (math !== null && math.length > 0) {
 					math = math[0];
 				}
-				return func(D2L.LP.Web.UI.Html.Math.AnnotateMathML('wiris', '', math, math));
+				return D2L.LP.Web.UI.Html.Math.AnnotateMathML('wiris', '', math, math);
 			};
 
+			// If we're dealing with applet math, convert to MathML
 			eventContext.content = eventContext.content.replace(/<applet[^>]*>(.|\n|\r)*?(<math[^>]*>(.|\n|\r)*?<\/math>)(.|\n|\r)*?<\/applet>/gi, appletFunc);
```
dbatiste, line 103: Maybe one day we can remove this. I've just been worried about breaking this conversion in case there is anyone trying to load old content containing an applet. It would have to be really really old I think.

components/plugins/equation.js
```diff
@@ -85,127 +90,68 @@ export function addEquationPlugin() {
 		editor.addQueryStateHandler(cmds.equationMathML, () => isEquationSelected(editorTypes.MathML));
 		editor.addQueryStateHandler(cmds.equationChemistry, () => isEquationSelected(editorTypes.Chemistry));
 
-		const trimMml = (mml, removeAnnotation, removeTitle) => {
```
dbatiste, line 88: Nice to remove this icky string manipulation code!

components/selection-helpers.js
```diff
@@ -0,0 +1,27 @@
+const bookmarks = new Map();
+
+export function moveToBookmark(editor, bookmarkId) {
+	const bookmark = bookmarks.get(bookmarkId);
```
dbatiste, line 4: It looks like the `Map` of bookmarks would get shared for all instances of the editor on the page. If there is more than one editor on the page, could this `bookmarks.get(...)` call return a bookmark for the wrong editor?

components/toolbar/menu-item.js
dbatiste: Do we get passed the node, and if so, can we detect this scenario easily and conditionally wait the frame only for this case? I'm generally not a fan of conditionally waiting a frame though... seems pretty hacky, and I probably prefer it the way you have it.

components/selection-helpers.js
grant-cleary: Ah, in theory yes, so I should probably map this to the editorId as well. In practice, no, because of how the bookmarks are used. The bookmarks live only for a very brief time when focus is removed from the editor. Still, it seems much cleaner to make this mapping editor-specific. TinyMCE actually offers an editor-specific key-value store now, so I may store this there.

components/plugins/equation.js
grant-cleary: Yeah, we did this just to keep the query string for the WIRIS requests as short as possible to avoid going over the character limit. That's not an issue anymore, so we don't need this anymore either! 🎉

components/plugins/equation.js
grant-cleary: Yeah, I'm reasonably sure the change I've made here will still function... I'm just switching up _when_ we convert from the applet code to make sure it happens before we start trying to process any equations.

components/toolbar/menu-item.js
grant-cleary: I believe we do get passed the node, yes. So in theory we could do something conditional here. I think I'm inclined to agree that I like the generic approach better. I didn't see anything go wrong otherwise with this code, so my inclination is to leave it as is for now, and then if we see this introduce weird bugs or anything, I can revisit this and make it conditional.

dbatiste (approved):

components/selection-helpers.js
```diff
@@ -0,0 +1,36 @@
+const bookmarks = new Map();
+
+function createCompositeBookmarkId(editor, bookmarkId) {
+	return `${editor.id}_${bookmarkId}`;
+}
+
+export function moveToBookmark(editor, bookmarkId) {
+	const id = createCompositeBookmarkId(editor, bookmarkId);
```
grant-cleary, line 8: I opted to do this in here so consumers don't need to care about it. 🤷

PR #1108, opened 2022-12-14
jprouse/US146554-change-request-method-to-come-from-same-origin

JakeProuse:
Rally Story: https://rally1.rallydev.com/#/?detail=/userstory/673386795721&fdp=true

This PR is to switch the htmleditor telemetry requests from being sent from inside the component using the d2l-telemetry-browser-client library, to be sent using window.d2lfetch helper which adds a correct auth token, relevant headers, and shows as coming from the same origin instead of cross-site. This should fix the current issue of receiving 502 responses from the telemetry endpoint.


components/plugins/telemetry.js
```diff
@@ -38,8 +36,13 @@ export function sendTelemetryEvent(sourceId, action, customProperties) {
 			.setSourceId(sourceId)
 			.setBody(eventBody);
 
-		client.logUserEvent(event);
-		return true;
+		// Take the formatted event body and use it in our own request.
+		const body = event.toJSON();
+
+		return window.d2lfetch.fetch(new Request(endpoint, {
```
grant-cleary, line 42: I don't know that we want this kind of depdendency on d2lfetch within the editor. @dbatiste, curious what your thoughts are here.

JakeProuse: Haha beat me to commenting first. It saves digging into localstorage for a fetch token, but that's an easy change if we don't want that dep.

JakeProuse: For reference, this would be the alternative
```javascript
const bearerToken = JSON.parse(localStorage['D2L.Fetch.Tokens'])['*:*:*'].access_token;

return fetch(endpoint, {
	method: 'POST',
	body,
	headers: {
		'Authorization': `Bearer ${bearerToken}`,
		'Content-type': 'text/plain;charset=UTF-8',
		'X-Csrf-Token': localStorage['XSRF.Token']
	}
});
```

grant-cleary: So there are a couple assumptions here that I don't think are valid. First of all, there's no guarantee that `window.d2lfetch` will be defined, since there's no guarantee the editor will run in the context of an LMS. While it's true that the current usage of this telemetry stuff can't work unless that's true (without some serious finagling), it's also technically not really a safe change because of that.

A similar problem occurs when it comes to fetching from localStorage... First of all, it assumes that localStorage will be accessible at all, and second of all, it assumes that the editor is going to be loaded in a context where the XSRF token is accessible. But that may not work as advertised, in, say, the Lessons FRA.

Broadly speaking, these are my concerns. When we bake in implicit reliance on LMS-specific code for functionality to work properly, we open the door to lots of this kind of weirdness. This is why I'm thinking we need to figure out a broad way to accomplish this for our core components, rather than just slap-dashing something into the editor.

JakeProuse: All good points Grant. Would checking if d2lfetch was available before sending an event work as an interm solution? 
```javascript
if (window.d2lfetch) ...
```
Doesn't fix the issue of having the extra dependency, and we'd only be collecting events from htmleditors that are in the LMS like in Content, but I think that would help as a start while we investigate a better approach for the core components like you mentioned.

dbatiste: My 2 cents... if `d2lfetch` is going to be used, it should probably be [imported as described](https://github.com/Brightspace/d2l-fetch). By importing it, `window.d2lfetch` should be initialized. As Grant mentioned, it's not safe to assume it will be present. For this to run in a FRA, we'd need some additional abstraction to proxy through ifrau services, but FRAs are going away so I think we can ignore that case as long as we fail gracefully and not break anything.

grant-cleary: Agreed with Dave... That would also mean importing the auth middleware, but I think that's still safer to be more explicit. I also think we should refine the `telemetryEndpoint` lookup to be something a bit simpler:
```document.documentElement.dataset.telemetryEndpoint```
and exit early if it doesn't exist. It's not a perfect check of whether we're in an LMS context, but it certainly gets us closer.

grant-cleary (approved):

components/plugins/telemetry.js
JakeProuse: Should I add d2l-fetch and d2l-fetch-auth as dependencies to the repo?

grant-cleary: They would need to be, yeah.

JakeProuse: Should I pin them to their major versions? I noticed that's what most of the deps were

grant-cleary: Yep, that's what we want.

PR #1115, opened 2022-12-19
jprouse/US146828-change-d2lfetch-reference-to-window-object

JakeProuse:
Rally Story: https://rally1.rallydev.com/#/?detail=/userstory/673785419229&fdp=true

Changes here are that I import the d2l-fetch file for side effects (setting the `window.d2lfetch` variable), and use the `addTemp()` method instead of `use()` so that the middleware is just used for my request instead of poluting the global instance.


JakeProuse (commented):
@dbatiste Hey Dave, mostly just fixing this bit where I imported the wrong file. That file just exports the d2lfetch method and doesn't set the window object, but d2l-fetch.js does (that's how their testing suite [is ran as well](https://github.com/Brightspace/d2l-fetch/blob/main/test/d2l-fetch/d2l-fetch.test.js#L1)).

Changing `use` for `addTemp` is to stop populting the window object if it didn't have the auth middleware for example.

I'm looking to get this onto the 20.23.1 release and was wondering if I'll need to make a release branch to do the CERT hotfix, since brightspace-bot has made some updates?

dbatiste (approved):

PR #1126, opened 2022-12-28
Make sure fullpage plugin is robust on initialization

grant-cleary:
The gist of this PR is to make the fullpage plugin more robust by ensuring the following:
1. Default settings are immediately respected when a new/empty document is created.
2. Only the body contents are passed on to other handlers. Once the fullpage plugin saves its state internally, we don't want to propagate the full HTML document (TinyMCE's handlers freak out).

This set of changes is intended to correct the following issues:
1. Default font settings were only being applied when the user interacted with the editor for the first time, leading to an inconsistent experience on first load.
2. Certain node types in the document head (e.g. comments) could be propagated into the document body because TinyMCE's internal handlers get a bit confused when they're passed a full HTML document.


components/plugins/fullpage.js
```diff
@@ -1,74 +1,143 @@
...
+	}
+
+	_updateStyles(body) {
+		this._fontColor = body.style.color || null;
+		this._fontFamily = body.style.fontFamily || null;
+		this._fontSize = body.style.fontSize || null;
+	}
+}
+
+const name = 'd2l-fullpage';
```
PaulKrznaric, line 119: What is the advantage of defining the name as a const in this case?

PaulKrznaric (approved):

components/plugins/fullpage.js
grant-cleary: Consistency, for the most part. Many of the plugins require that this name is used in more than one place (e.g. for registering commands), so setting the name as a constant facilitates that and keeps all of our plugins consistent.

davewagler-d2l (approved):

PR #1130, opened 2022-12-30
Update TinyMCE's iframe/body labels when editor's label is updated

grant-cleary:
Right now we only set TinyMCE's internal labels (on the iframe and document body) when TinyMCE first initializes. This PR adds some brief logic to allow these to be updated post-initialization if the editor's label changes.


dbatiste (approved):

dbatiste:
Would not be a bad idea to add a unit test for this case.

grant-cleary:
> Would not be a bad idea to add a unit test for this case.

Oh, yes. I put up this PR for now with the intent of running it by you when you were back from vacation to see what you thought of the idea, haha. If you're good with it (I assume so given you approved), I can add a test.

dbatiste (approved):

PR #1132, opened 2023-01-03
Fullpage plugin should load first and should allow stylesheets to be added by other plugins

grant-cleary:
There's a bit of an issue with our fullpage plugin implementation and Content Styler. First, Content Styler can end up clobbering the fullpage plugin with its scripts, so we need to load our fullpage plugin first. Second, we need a way for the fullpage plugin to be able to know and track which stylesheets other plugins may have added to the editor. To facilitate this, we expose a function that allows plugins to do this.


components/plugins/fullpage.js
```diff
@@ -20,6 +24,16 @@ class FullPageState {
 		this._docElementAttrs = new Map();
 		this._docType = document.implementation.createDocumentType('html', '', '');
 		this._head = undefined;
+
+		this._stylesheets = new Set();
+	}
+
+	addStyleSheets(content, stylesheets) {
+		for (const stylesheet of stylesheets) {
+			this._stylesheets.add(stylesheet);
```
dbatiste, line 33: Given that the `styesheet` is an object, are you not worried if it already exists in the `Set`? Each object will be a unique instance even their property values are the same.

components/plugins/fullpage.js
```diff
@@ -114,6 +132,26 @@ class FullPageState {
 		this._fontFamily = body.style.fontFamily || null;
 		this._fontSize = body.style.fontSize || null;
 	}
+
+	_updateStylesheets(doc, updateEditorDoc) {
+		for (const stylesheet of this._stylesheets) {
+			const styleUrl = stylesheet.url;
+			if (doc.head.querySelector(`link[href="${styleUrl}"]`)) continue;
```
dbatiste, line 139: ... continuation from my other comment, it looks like we'd skip it here if it's already in the document, data attribute differences aside. Though it seems like the `Set` could grow and grow.

grant-cleary: Yeah, in practice we don't get duplicates in the existing code because of these checks, but just to cover our bases I can adjust the data structure to make sure we're only ever storing one entry per URL.

components/plugins/fullpage.js
```diff
@@ -114,6 +132,26 @@ class FullPageState {
...
+
+			const elem = doc.createElement('link');
+			elem.rel = 'stylesheet';
+			elem.href = styleUrl;
+
+			for (const [attrName, attrValue] of stylesheet.dataAttrs) {
+				elem.data[attrName] = attrValue;
+			}
+
+			doc.head.appendChild(elem);
```
dbatiste, line 149: I think I might have forgotten, but what is the difference between `doc` and `this._editor.contentDocument`?

components/toolbar/toolbar-full.js
```diff
@@ -46,6 +46,8 @@ export function getTagName() {
 
 export function getPlugins(component) {
 	return [
+		// Always load the fullpage plugin first if relevant, as other plugins rely on it
```
dbatiste, line 49: I feel like there could be a better way to define plugins that depend on other plugins and then build this array accordingly. But I don't want to introduce unnecessary complexity if we can manage this simply with a comment. Something to think about going forward though.

demo/htmleditor.html
```diff
@@ -29,7 +29,7 @@
 	<body unresolved>
 		<d2l-demo-page page-title="d2l-htmleditor (full)">
 
-			<h2>Default</h2>
+			<!--<h2>Default</h2>
```
dbatiste, line 32: Did you intend to leave these comment tags here?

grant-cleary: Ah, shoot, no. Accidentally undid their removal before pushing.

components/toolbar/toolbar-full.js
grant-cleary: I agree. I have a longer term plan that's definitely better than this. I just didn't think it was really worth it for something we would be cert fixing. I wasn't exactly intending on this being a step towards defining better plugin APIs, but it ended up being a necessary requirement to make this plugin play nicely with Content Styler.

components/plugins/fullpage.js
grant-cleary: Ok, I switched this up to use a map keyed on the URL, so even if a consumer gets things wrong, we shouldn't get this growing unbounded. I also switched up the dataAttrs to be a Map to make access a bit easier.

dbatiste (approved):

PR #1139, opened 2023-01-09
Make sure existing stylesheets are applied when content is first loaded

grant-cleary:
Round 1536361 of fullpage compatibility fixes... This time we need to make sure that existing stylesheets are propagated into the editor document when loaded.


components/plugins/fullpage.js
```diff
@@ -134,21 +134,33 @@ class FullPageState {
 	}
 
 	_updateStylesheets(doc, updateEditorDoc) {
-		for (const [styleUrl, styleDataAttrs] of this._stylesheets) {
-			if (doc.head.querySelector(`link[href="${styleUrl}"]`)) continue;
+		// Handle existing stylesheets
+		doc.head.querySelectorAll('link').forEach(link => {
```
grant-cleary, line 138: The editor may not know about these stylesheets yet, in which case we need to make sure we grab and store them appropriately so they can be propagated into the editor.

components/plugins/fullpage.js
```diff
@@ -134,21 +134,33 @@ class FullPageState {
...
+			if (!updateLinkInEditor && doc.head.querySelector(`link[href="${styleUrl}"]`)) continue;
 
 			const elem = doc.createElement('link');
 			elem.rel = 'stylesheet';
 			elem.href = styleUrl;
 
-			for (const [attrName, attrValue] of styleDataAttrs) {
-				elem.dataset[attrName] = attrValue;
+			for (const [attrName, attrValue] of styleAttrs) {
+				elem.setAttribute(attrName, attrValue);
```
grant-cleary, line 159: I had to make this more generic because for arbitrary stylesheets we can't know for sure what the attributes may be.

dbatiste (commented):
Any unit tests to add for this?

grant-cleary:
> Any unit tests to add for this?

I probably could. Maybe a simple test that adds a script to the initial HTML and expects to find it afterwards.

dbatiste:
Unless there is a visual-diff test that covers it, I think we should try to cover it.

dbatiste (approved):

PR #1142, opened 2023-01-10
DE48821 - Revert styling on Blockquotes to match LMS

PaulKrznaric:
Currently, `blockquotes` when edited in the HTML Editor, do not visually match the LMS's styling on user-generated HTML.
This fix aims to address this issue.

### Before:
image
image

### After:
image
image


htmleditor-mixin.js
```diff
@@ -93,6 +93,16 @@ const contentFragmentStyles = css`
 		background-color: inherit;
 		outline: none !important;
 	}
+	.mce-content-body:not([dir=rtl]) blockquote {
+		border-left: revert;
+		margin-left: revert;
+		padding-left: revert;
+	  }
```
grant-cleary, line 100: Nit: whitespace is wrong on these. I suspect it may fail linting but we'll want to fix them regardless.

grant-cleary (approved):

ca2hill (approved):

PR #1144, opened 2023-01-11
DE48683 - Equation Editor Dialog Languification

ca2hill:
Include the orgunitid in the route to create the editor in a dialog.

Dependent on this change: [](https://github.com/BrightspaceUI/htmleditor/pull/1144)


components/plugins/equation.js
```diff
@@ -196,7 +202,7 @@ class EditorDialog extends LitElement {
 
 			const result = await openDialogWithParam(
 				getComposedActiveElement(),
-				'/d2l/lp/math/createeditor',
+				'/d2l/lp/math/createeditor/${this._orgUnitId}/',
```
dbatiste, line 205: I think this would need to be a template:
```suggestion
				`/d2l/lp/math/createeditor/${this._orgUnitId}/`,
```

grant-cleary (approved):

PR #1148, opened 2023-01-12
DE47689

ca2hill:
Make font sizes distinct in the list


grant-cleary (approved):

dbatiste (commented):
I think I am a bit disconnected from the issue. I don't quite follow how this fix relates to the defect (i.e. what the root cause was).

grant-cleary:
> I think I am a bit disconnected from the issue. I don't quite follow how this fix relates to the defect (i.e. what the root cause was).

If `defaultFontSize` happens to be identical to any other size in the font list, then the font size dropdown will display duplicates.

dbatiste:
Cool, thanks. I didn't clue into the `defaultFontSize` part and how it could be already an element in the array.

dbatiste (approved):

PR #1165, opened 2023-01-23
Make sure TinyMCE's iframe is tabbable

grant-cleary:
As of TinyMCE 6, TinyMCE seems to be reflecting the tabindex property from its internal textarea onto the editor iframe. Because we have this set to `tabindex="-1"` on the textarea element, TinyMCE in turn sets that on the editor iframe and thus removes it from the tab order. This means that currently the editor iframe and its contents are completely unreachable via keyboard nav.


htmleditor-mixin.js
```diff
@@ -947,7 +947,7 @@ export const HtmlEditorMixin = superclass => class extends Localizer(LabelledMix
 		};
 
 		return html`
-			<textarea id="${this._editorId}" class="${classMap(textAreaClasses)}" style="${styleMap(textAreaStyles)}" aria-hidden="true" tabindex="-1">${this._html}</textarea>
```
grant-cleary, line 950: As far as I can tell, whatever changed this behaviour in TinyMCE 6 appears to be undocumented. 😢

PaulKrznaric (approved):

PR #1174, opened 2023-01-25
Switch to DOMParser instead of regex for converting MathJax-rendered equations back to MathML

grant-cleary:
The regex previously used to convert images back to MathML for equations became a bit overzealous when we switched from looking for images to looking for spans instead because of a greedy execution that didn't stop when the closing span tag was identified. The simplest solution here is to just jettison the regex entirely and switch to a DOMParser-based strategy like we do elsewhere in the editor.


components/plugins/equation.js
```diff
@@ -104,15 +104,15 @@ export function addEquationPlugin() {
 		};
 
 		const convertToMathML = (eventContext) => {
-			const func = (match) => {
-				// find the MathML
-				const hackRegEx = /^<span[^>]*\s+data-d2l-mathml=\"([^\"]*)\"(.|\n|\r)*/i;
```
grant-cleary, line 109: I certainly didn't name these things; this was pulled largely as is from the LMS code and then modified, but the names may give you an idea of why I don't at all feel bad ditching this code. 😆

components/plugins/equation.js
```diff
@@ -104,15 +104,14 @@ export function addEquationPlugin() {
...
-				const hackResult = hackRegEx.exec(match);
-				const hack = tinymce.DOM.decode(hackResult[1]);
-				return hack;
-			};
+			const doc = new DOMParser().parseFromString(eventContext.content, 'text/html');
+			if (!doc || !doc.body) return;
+
+			doc.querySelectorAll('span[data-d2l-mathml]').forEach(span => {
+				span.outerHTML = span.dataset.d2lMathml || '';
```
grant-cleary, line 111: If this attribute is null or undefined for some reason, replace it with an empty string before wiping out the span. Otherwise we're going to end up with literal `undefined` or `null` strings being dumped into the editor.

klhuff: Is the M capitalized on purpose? My understanding is that this should match the data property, minus the "data-" and other "-".

davewagler-d2l (approved):

components/plugins/equation.js
klhuff: Ah, the docs do mention the name gets camelcased. cool

klhuff (approved):

components/plugins/equation.js
davewagler-d2l: Looks like it goes to camel case:
https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dataset

grant-cleary: Ah, it's the way attributes map to properties. Everywhere there's a `-`, it's removed and the next letter is capitalized.

grant-cleary: Oops, you beat me to it. Yes, haha.

PR #1180, opened 2023-01-27
Allow plain text to be passed in to the fullpage handlers (just wrap it appropriately)

grant-cleary:
There's an edge case that used to work in the editor in which you could open the source code editor in fullpage mode, delete all of the markup, and then enter plain text. TinyMCE would take that text, jam it in a `p` tag, and wrap it in the standard document markup. This PR tweaks our own fullpage implementation to do the same.


components/plugins/fullpage.js
```diff
@@ -72,8 +72,8 @@ class FullPageState {
 
 		if (!fullDocTestRegex.test(content)) {
 			this._updateEditorBody(this._editor);
-			this._updateHead(doc);
-			return '';
+			this._updateStylesheets(doc, true);
```
grant-cleary, line 75: We want to update any stylesheets as needed, but we don't want to wipe out the stored head if it exists.

components/plugins/fullpage.js
```diff
@@ -72,8 +72,8 @@ class FullPageState {
...
+			return content;
```
grant-cleary, line 76: This is the crux of the change. Instead of wiping out plain text coming in, just spit it right back out.

davewagler-d2l (approved):

PR #1187, opened 2023-01-30
DE50456 prevent linking to editor version of quicklink

mdulat:
[DE50456](https://rally1.rallydev.com/#/?detail=/defect/664062096087&fdp=true): Quizzes > 20.22.10 > In New Quiz Editing Experience, Editing Quiz Description and selecting Quicklink to Content file leads to Error Page Not Found

When a quicklink is added to the editor, it contains an orgUnitId template in it's href (i.e. ou={orgUnitId}) and is not intended to be linked to directly. If the user clicks on the link while the editor is not in edit mode (i.e. the inline editor is not in focus, toolbar not showing), then the link is followed when clicked, leading to a 404 not found page. We want to prevent following the link when the user clicks on it within the editor. It should only bring the editor into focus, allowing editing of the link.

See https://d2lmail-my.sharepoint.com/:v:/r/personal/jchapman_desire2learn_com/Documents/Shared%20with%20Everyone/module1%20-%20FACE%20opt-in%20(Competencies)%20-%20Google%20Chrome%202023-01-26%2017-23-14.mp4?csf=1&web=1&e=DRWVh6


grant-cleary:
Ah, I see. So when we switch into editing mode, we convert the replace string back into its original form, but that happens before the click event propagates down to the link. I _think_ I'm okay with the idea of the editor not propagating click events into its iframe when it's still transitioning into editing mode, but I would be curious about @dbatiste's thoughts. I would also like to take a bit to compare before/after scenarios to make sure nothing surprising is changing here.

dbatiste:
Yeah this seems reasonable.

grant-cleary (approved):

PR #1216, opened 2023-02-09
Only use MathJax to typeset the document once to avoid multiple reflows

grant-cleary:
Every time MathJax typesets math on the page, it needs to fetch some DOM properties that trigger reflows. This is problematic in the editor's edit mode when editing content that contains a large number of equations. To get around this, we can just apply the same sort of logic we do in Content and typeset the document exactly once. This comes with the caveat that this won't really work for us if we opt to support web components in the editor, but... We can cross that bridge later.


components/plugins/equation.js
```diff
@@ -22,7 +22,7 @@ export function addEquationPlugin() {
 		const context = requestInstance(editor.getElement(), 'context');
 
 		// bail if no LMS context
-		if (!context || !context.activeOrgEnrollment) return;
+		//if (!context || !context.activeOrgEnrollment) return;
```
grant-cleary, line 25: Oops. Added for testing, mistakenly kept.

PaulKrznaric (approved):

klhuff (approved):

PR #1217, opened 2023-02-09
Add property to toggle more efficient rendering

grant-cleary:
Sorry, I forgot we wanted this change to be toggle-able...


klhuff (approved):

grant-cleary:
As long as this property doesn't exist, then these checks will always return false, so we're safe here. The flag addition has to wind its way through the LMS, which is more onerous, so I'll get CI going on that PR and I can backport the editor changes in parallel.

PR #1228, opened 2023-02-13
Implement more performant versions of TinyMCE/MathJax workaround logic

grant-cleary:
TinyMCE's serializer filters aren't very specific (they're not full query selectors), so we have to do a bunch of unnecessary processing when we're trying to use them to parse content. Instead of using them, fall back to using standard query selectors so we can more specifically target the elements we care about.


htmleditor-mixin.js
```diff
@@ -786,7 +915,11 @@ export const HtmlEditorMixin = superclass => class extends Localizer(LabelledMix
 						convertImagePlaceholderToOriginalContent,
 						decodeSrcReplaceStrings,
 						restoreActionIds,
-						restoreGlobalEventHandlers
+						restoreGlobalEventHandlers,
+						restoreIframeChildren,
+						restoreInvalidElements,
+						removeMathJaxMarkUp,
+						restoreScripts
```
PaulKrznaric, line 922: So we're just able to pass these functions in, and the editor can run them rather than the functions that were set up below (and are now removed)?

PaulKrznaric (approved):

htmleditor-mixin.js
grant-cleary: It doesn't get passed into TinyMCE in this case; this is a custom workflow I built to allow arbitrary operations to be run HTML as it's pushed into or out of the editor. It uses the DOMParser to construct a temporary document and runs all of these converters against it before returning the parsed HTML.

PR #1235, opened 2023-02-16
Jprouse/us148311 layouts editor setup

JakeProuse:
Rally Story: https://rally1.rallydev.com/#/?detail=/userstory/684999723889&fdp=true

- Moving [checkDependencies](https://github.com/BrightspaceUI/htmleditor/blob/main/components/plugins/lcs.js#L36), [checkPageValidity](https://github.com/BrightspaceUI/htmleditor/blob/main/components/plugins/lcs.js#L117), and the [hook in SetContent](https://github.com/BrightspaceUI/htmleditor/blob/main/components/plugins/lcs.js#L197) from lcs.js into a common file where it can be importing to the layouts plugin as well (so it finds or puts in the template styling). Where should this file live? 

- Need a good name for the SetContent hook (setCreatorPlusTemplate right now).

**Purpose of PR**
- Setup layouts button
- Hide layouts behind feature flag
- Able to drop a layout from a selection
- Able to type in layout
- Move up/down and delete layouts


components/plugins/creator-plus.js
```diff
@@ -0,0 +1,155 @@
...
+					if (hasDefaultStyle && styleSheet.href.split('/').pop() === 'bootstrap.min.css') {
+						editor.contentDocument.head.insertBefore(elem, editor.contentDocument.head.querySelector('link[href*="styles.min.css"]'));
+					}
+					else {
+						if (!hasOverride) {
+							editor.contentDocument.head.appendChild(elem);
+						}
+						else {
+							const override = editorDocument.head.querySelector('link[data-override="override"]').href.split('/').pop();
+							editor.contentDocument.head.insertBefore(elem, editor.contentDocument.head.querySelector(`link[href*="${override}"]`));
```
grant-cleary, line 148: I would recommend using the `addStylesheets` API for this call and the one above.

components/plugins/layouts.js
```diff
@@ -0,0 +1,309 @@
...
+const name = 'd2l-layouts';
+let scripts, styleSheets;
+
+export function setLayoutsToNotEditable(htmlDoc) {
+	const layouts = htmlDoc.body.querySelectorAll('.d2l-cplus-layout');
+	if (!layouts || layouts.length === 0) return;
+
+	for (const layout of layouts) {
+		layout.parentElement.contentEditable = false;
+		layout.parentElement.removeAttribute('contenteditable');
```
grant-cleary, line 30: This seems redundant. Any reason we can't just keep this last line?

components/plugins/layouts.js
```diff
@@ -0,0 +1,309 @@
...
+			dialog.lang = editorDocument.querySelector('html').getAttribute('lang') || 'en-us';
+
+			sendLaunchLayoutMenuEvent(context);
+
+			component._dialogOpen = true;
+			dialog.opened = true;
+
+			dialog.addEventListener('d2l-layout-dialog-close', e => {
+				if (!e.detail || !e.detail.layout) {
+					editor.editorContainer.getRootNode().querySelector('.tox-tinymce-aux').innerHTML = '';
```
grant-cleary, line 57: Is this still necessary? I thought this bug was fixed at some point.

components/plugins/layouts.js
```diff
@@ -0,0 +1,309 @@
...
+		super.disconnectedCallback();
+		this._resizeObserver.disconnect();
+	}
+
+	firstUpdated() {
+		super.firstUpdated();
+
+		const editor = this.shadowRoot.querySelector('d2l-dialog').querySelector('.d2l-editor');
+		if (editor !== null) {
+			this._resizeObserver.observe(editor);
```
grant-cleary, line 203: Attaching the observer here, but detaching it in `disconnectedCallback` means if the dialog is ever detached and re-attached to the DOM, the observer will be wiped out. If that's an acceptable risk, then cool. Otherwise you may wish to move this to `connectedCallback`.

components/plugins/layouts.js
```diff
@@ -0,0 +1,309 @@
...
+			return html`
+			<d2l-dialog ?opened="${this.opened}" title-text="${this.validPage ? this.localize('layouts:heading') : this.localize('layouts:invalidPageTitle')}" @d2l-dialog-close="${this._handleClose}">
+				<div class="d2l-editor">
+				${content}
+				</div>
+				${!this.validPage ?  html`<d2l-button slot="footer" primary data-dialog-action>${this.localize('layouts:ok')}</d2l-button>` : ''}
+			</d2l-dialog>
+		`;
+		} catch (error) {
+			console.log(error);
```
grant-cleary, line 256: Not super keen on seeing this here. Is there a particular error you expect to catch here? I would rather let exceptions through unless you already know there's a likely case they can crop up.

components/plugins/layouts.js
```diff
@@ -0,0 +1,309 @@
...
+			}
+
+			this.dispatchEvent(new CustomEvent(
+				'd2l-layout-dialog-close', {
+					bubbles: true,
+					detail
+				}
+			));
+		} catch (error) {
+			return;
```
grant-cleary, line 290: What's the purpose of this try-catch?

icons/layouts.svg
```diff
@@ -0,0 +1,14 @@
+<svg width="21" height="21" viewBox="0 0 21 21" fill="none" xmlns="http://www.w3.org/2000/svg">
```
grant-cleary, line 1: Note that our toolbars are expecting 18x18 icons, so I think this is likely to be cut off some.

components/plugins/layouts.js
JakeProuse: Ah sorry I'd left that log in there by mistake, and the try/catches are something our team has been using more in our own stuff to try to reduce the amount of breaking errors. If that isn't wanted for the editor that's okay, I'll remove them 👍

components/plugins/layouts.js
JakeProuse: That makes sense, thanks!

components/plugins/layouts.js
JakeProuse: Justin just tested this with InsertElement and had no issue when removing this line and the .collapse() underneath, so I'll remove it 👍

components/plugins/creator-plus.js
grant-cleary: The stylesheet reference constructed via `addStylesheets` is ordered by insertion, so as long as you add them in the correct order, it should work.

grant-cleary: I mean there _shouldn't_ be a case where an override is set, but the base URL isn't, no?

JakeProuse: So even if there was already an existing override link in the head, if called addStyleSheets with `[ templateCSS, contentStylerOverride ]` then the head would get re-arranged and our templateCSS would get put first?

JakeProuse: Yeah Justin said when adding a template we'd want an override without setting the base URL. This is already how lcs.js adds our links, I'm just moving this into a common function that we can share for now. Justin just mentioned he'll also going to be making an update next week (needs to time it with our CDN release) that entirely removes lines 42-66 and the entire `setCreatorPlusTemplate()` function, so we wouldn't need the addStyleSheet adjustment anyway. This would stop InsertElement and Layouts from needing to add CSS at all. This in mind, is it okay to leave it as is for now?

grant-cleary: Sure, that's fine then. If this is all going away anyways then I have no problem with it staying as is for now.

I have a plan to heavily restrict what access plugins will have to TinyMCE's APIs and iframe directly, and will instead look to surface APIs within the core editor that plugins will use instead (basically to reduce the surface area of places we need to care about when updating TinyMCE). Hence why I'm interested in preventing new access to direct manipulations to the editor document like this.

components/plugins/layouts.js
JakeProuse: Looks like I don't need this function, the layouts are uneditable on save without it. This is from when I'd been adding insert element classes I believe.

grant-cleary: Yay! Always nice when code turns out to be redundant and removable, haha.

components/plugins/creator-plus.js
JakeProuse: Thanks Grant! And those sound like great changes, will you be doing a similar method for adding scripts to the page like with `addStyleSheets()`?

grant-cleary: Yeah, the idea will be to enable, at minimum, the same functionality that currently exists. The intent isn't to curtail existing functionality, but to just make it easier to track what stuff might break when TinyMCE changes their APIs.

grant-cleary (approved):

icons/layouts.svg
JakeProuse: Updated 👍 I also updated the color to match the other icons, is #4A4C4E a brightspace color or something chosen for the htmleditor buttons?

dbatiste: The other icons appear to be `#494c4e`, which is `--d2l-color-tungsten`, and is consistent with core icons. Is that the color you meant? I can't find any reference to `#4A4C4E`.

dbatiste: I would run it through the steps [described here for core icons](https://github.com/BrightspaceUI/core/tree/main/components/icons#updating-or-contributing-new-icons).

JakeProuse: Weird, I see `#4A4C4E` when using a color picker on the icons. I see that the button fill property is set as the --d2l-color-tungsten

dbatiste: I find the eye dropper thingy can give slightly different results depending on aliasing and where it samples from. I only sample colors that way when I have no other way to get the hex.

PR #1236, opened 2023-02-16
Use spans for XSPL placeholder HTML to ensure greater compatibility with various parent elements

grant-cleary:
Iframes can be considered both flow and phrasing content. Thus, to ensure compatibility with as many possible parent elements as possible when we replace them with our XSPL placeholders, those placeholders should also meet that criterion. Thus, we should use a `<span>` here.


components/plugins/xspl.js
```diff
@@ -26,8 +26,9 @@ export function convertIframeToImagePlaceholder(htmlDoc) {
 
 		// Sets a CSS background image for the current insert-stuff icon. Suitable as a placeholder, but we
 		// may revisit this with a new icon in the future if necessary.
-		const placeholder = document.createElement('p');
+		const placeholder = document.createElement('span');
```
grant-cleary, line 29: Can't use a `<p>` or `<div>` tag here because the placeholder's immediate parent could be, for example, a `<p>` tag itself (and that's actually quite likely). Hence, we need a placeholder that shares the same content categories with the `<iframe>` element.

PaulKrznaric (approved):

PR #1238, opened 2023-02-16
fix: add state to cancel practice event

JakeProuse:
Rally Story: https://rally1.rallydev.com/#/?detail=/userstory/685773148695&fdp=true

Merging to main, then backporting to 20.23.3


components/events/creator-plus/practices/practices.js
```diff
@@ -30,7 +31,7 @@ export function sendCancelPracticeEvent(id, context) {
 export function sendCancelEditPracticeEvent(id, context) {
 	const customProperties = {
 		Tool: tools.practices,
-		Id: id,
+		Id: `${id}`,
```
grant-cleary, line 34: Interesting. These don't get massaged into strings somewhere downstream?

grant-cleary (approved):

components/events/creator-plus/practices/practices.js
JakeProuse: Good catch grant, they do. I'll update that!

PR #1247, opened 2023-02-21
Set display properly on math placeholder so TinyMCE's noneditable outline shows when focused

grant-cleary:
When block math is created with the equation editor, the containing span remains an inline element so TinyMCE's noneditable styles (primarily the focus outline) don't show properly. This PR sets the containing span to use a block rendering context if the underlying math is block math, solving our problem.


htmleditor-mixin.js
```diff
@@ -56,12 +56,40 @@ const a11yCheckerRuleUrlMap = {
 	'T4C': 'https://www.w3.org/WAI/WCAG21/Techniques/html/H63.html'
 };
 
-const fullPageStyles = css`
+const commonStyles = css`
```
grant-cleary, line 59: This is a partially unrelated refactor. I needed this change to apply both to fullpage and non-fullpage modes and while I was factoring that out into a common CSS declaration, I realized that this should also apply to the other TinyMCE-specific workarounds we've applied.

PaulKrznaric (approved):

PR #1252, opened 2023-02-22
Extend invalid element replacement workaround to pattern elements

grant-cleary:
This PR generalizes the workflow added to work around some of TinyMCE's invalid element filtering within the editor to allow it to work for arbitrary elements. We're also adding the `<pattern>` element to the list because it's another known problem case.


htmleditor-mixin.js
```diff
@@ -818,57 +818,77 @@ export const HtmlEditorMixin = superclass => class extends Localizer(LabelledMix
...
+						}
+						return undefined;
+					};
+
+					const selector = Array.from(elementReplacementMapping.values()).map(tagName =>
+						tagName.concat('.d2l-invalid-element-placeholder')
+					).join();
+					doc.body.querySelectorAll(selector).forEach(elm => {
+						const originalTagName = getOriginalElmTagName(elm);
+						if (!originalTagName) return;
```
PaulKrznaric, line 869: Should this `continue` instead of `return`?

PaulKrznaric (approved):

htmleditor-mixin.js
grant-cleary: No, `forEach` is iterative, so you can think of its inner function as a callback that gets executed against each element of the array rather than, say, a for...of loop. So `return` exits the callback early but successive callbacks still run against the other elements.

PR #1253, opened 2023-02-22
Extend invalid element replacement workaround to pattern elements (backport)

grant-cleary:
This is **not** a backport, but is meant to convey the same logic as this PR: https://github.com/BrightspaceUI/htmleditor/pull/1252

A clean backport wasn't possible because I refactored the way we do these replacements in `main` and I didn't want to bring those changes back.


htmleditor-mixin.js
```diff
@@ -976,6 +981,32 @@ export const HtmlEditorMixin = superclass => class extends Localizer(LabelledMix
 								node.replace(contentNode);
 							});
 						});
+						editor.serializer.addNodeFilter('img', (nodes) => {
```
grant-cleary, line 984: Not really interested in finagling with TinyMCE's APIs in a way that would let us consolidate these. There's no code duplication here in `main` so I'm unconcerned about there being a bit of it here in the backports.

PaulKrznaric (approved):

PR #1256, opened 2023-02-23
Jprouse/ DE52190-content styler enabled on templates

JakeProuse:
Rally Story: https://rally1.rallydev.com/#/?detail=/defect/686321892739&fdp=true

The defect is that content styler is being added to any page, and we only want it to be applied to pages that are not a template page. I'll be adding logic to the template button in the LMS and BrightspaceHypermediaComponents/Activities to set this prop to true when initializing the editor.

@grant-cleary I wanted to double check with you about the name of the prop.

This merge shouldn't affect anything yet, since the isTemplate prop will not be defined.


htmleditor-mixin.js
```diff
@@ -202,6 +202,11 @@ export const HtmlEditorMixin = superclass => class extends Localizer(LabelledMix
 			 * @type {boolean}
 			 */
 			insertPractice: { type: Boolean, attribute: 'insert-practice' },
+			/**
+			 * Whether the htmleditor was initialized using a template.
+			 * @type {boolean}
+			 */
+			isTemplate: { type: Boolean, attribute: 'is-template' },
```
grant-cleary, line 209: I'm not sure I would create a new property for this. My thought was more that when adding a template, we could flip the existing `contentStyler` property off.

JakeProuse: Ah, so adding a line to the [content styler enabled variable](https://search.d2l.dev/xref/lms/lp/framework/web/D2L.LP.Web.UI/Desktop/Controls/HtmlEditor/HtmlEditorContextHtmlElementAttributeProvider.cs?r=f5b8a195#147)?

grant-cleary: Unfortunately it's not quite that easy because the editor already exists when the template is added, so the context object can't really be changed at that point. Rather, I was thinking that we could call something like `editor.contentStyler = false` when adding a template, and then your plugin short-circuit would just be:
```if (!component.contentStyler) return;```

JakeProuse: That makes sense, I'll make the change 👍

grant-cleary: Basically in the same place you would have needed to set `isTemplate`, we just toggle off `contentStyler` instead.

JakeProuse: Updated 🎉

grant-cleary (approved):

PR #1269, opened 2023-02-28
When trying to find out which block formatting context we're in, traverse the DOM tree until we find a valid context

grant-cleary:
This is a more robust detection method for identifying whether we're in a formatting block. If the immediate selection is not in a block, walk up the DOM until we find the nearest block context.


components/plugins/formatting.js
```diff
@@ -4,19 +4,26 @@ import { requestInstance } from '@brightspace-ui/core/mixins/provider-mixin.js';
...
+			else return getBlockElement(elm.parentNode);
+		};
+
 		// TinyMCE 5 reflected the currently selected tag bag when querying the formatBlock command value.
 		// This no longer appears to be the case, so we need to do this ourselves for the format selector to work.
 		editor.addQueryValueHandler(cmds.formatBlock, () => {
 			const selectedNode = editor.selection && editor.selection.getNode();
-			if (selectedNode && selectedNode.nodeType === Node.ELEMENT_NODE) return selectedNode.tagName.toLowerCase();
-			else return 'p';
+			return getBlockElement(selectedNode) || 'p';
```
grant-cleary, line 26: If we never find a block context, return a default value to display.

components/plugins/formatting.js
```diff
@@ -4,19 +4,26 @@ import { requestInstance } from '@brightspace-ui/core/mixins/provider-mixin.js';
...
 
 	tinymce.PluginManager.add(name, function(editor) {
 
 		const context = requestInstance(editor.getElement(), 'context');
 		const component = context.component;
 
+		const getBlockElement = elm => {
+			if (!elm) return undefined;
+			if (elm.nodeType === Node.ELEMENT_NODE && blockNodes.has(elm.tagName.toLowerCase())) return elm.tagName.toLowerCase();
+			else return getBlockElement(elm.parentNode);
```
ca2hill, line 19: I hate these `else`s, but acknowledge that this maybe more acceptable in JS.

Dan-DeAraujo (approved):

components/plugins/formatting.js
grant-cleary: Yes, this is pretty standard.

PR #1273, opened 2023-03-01
Remove FocusVisiblePolyfill and replace with pseudo-class selector

grant-cleary:
Pretty much does what it says on the tin, ditching the `FocusVisiblePolyfillMixin` for a more robust replacement.


dbatiste (approved):
Looks good to me!

@eKoopmans fyi... in case you are interested given you are tackling this story for core.

components/button-styles.js
```diff
@@ -59,16 +60,13 @@ export const buttonStyles = css`
 		z-index: -1;
 	}
 	button:hover > .d2l-htmleditor-button-background,
-	button:focus > .d2l-htmleditor-button-background,
-	:host(:not([data-js-focus-visible])) button:focus > .d2l-htmleditor-button-background,
-	button.focus-visible > .d2l-htmleditor-button-background {
+	button.focus-visible > .d2l-htmleditor-button-background,
```
eKoopmans, line 63: Just making sure I understand - we're keeping the `.focus-visible` styles here because `forceFocusVisible` is being used, and that relies on adding the class still, correct?

dbatiste: That's correct.

PR #1276, opened 2023-03-02
Expand invalid HTML workaround to include multimedia/heading elements, as well as recursive replacements

grant-cleary:
There are two main goals here:
1. Add `<heading>` and `<multimedia>` elements to the temporary workaround.
2. Make the workaround recursive so it can replace nested invalid elements.


htmleditor-mixin.js
```diff
@@ -820,36 +820,46 @@ export const HtmlEditorMixin = superclass => class extends Localizer(LabelledMix
...
-						tempElm.dataset.d2lInvalidOriginalElement = originalTagName;
-						elm.replaceWith(tempElm);
-					});
+							tempElm.classList.add('d2l-invalid-element-placeholder');
+							tempElm.dataset.d2lInvalidOriginalElement = originalTagName;
+							elm.replaceWith(tempElm);
+
+							// Recurse through any children to ensure we catch
+							// nested invalid elements.
+							replaceElementsWithPlaceholders(tempElm);
```
grant-cleary, line 858: This is necessary because the results returned from `(document|element).querySelectorAll(...)` are static. When we modify the DOM to do one of these replacements, the node list becomes stale and we end up skipping over any children of the newly-replaced placeholder.

grant-cleary: As a sidenote: Yes, this does introduce scary performance concerns if there's enough nesting here. No, there's not really anything we can do about it. This functionality is not being enabled across the board.

joshhoey: Ick.  But also, okay. :|

joshhoey (approved):

PR #1277, opened 2023-03-02
Expand invalid HTML workaround to include multimedia/heading elements, as well as recursive replacements

grant-cleary:
Conceptually the same as this PR: https://github.com/BrightspaceUI/htmleditor/pull/1276


htmleditor-mixin.js
```diff
@@ -918,11 +918,13 @@ export const HtmlEditorMixin = superclass => class extends Localizer(LabelledMix
...
 					});
 
-					if (this._context.allowInvalidElements) {
+					if (!this._context.allowInvalidElements) {
 						const elementReplacementMappings = new Map([
 							['content', 'div'],
 							['pattern', 'img'],
-							['background', 'div']
+							['background', 'div'],
+							['header', 'div'],
```
grant-cleary, line 926: This is actually slightly different, but in versions <= 20.23.4, making this a `<span>` replacement would interfere with the script workaround.

joshhoey (approved):

PR #1282, opened 2023-03-03
Content Styler: Moving template CSS so that it gets added before the override CSS

jschlumkoski:


jschlumkoski:
@grant-cleary This update just makes sure the template CSS is added before the overrides (if necessary) so they apply properly.

grant-cleary (approved):

PR #1297, opened 2023-03-14
Jprouse/us148311 layouts editor setup

JakeProuse:
Rally Story: https://rally1.rallydev.com/#/?detail=/userstory/684999723889&fdp=true

Some updates to the UI. The menu is now responsive and is closer to the design.


JakeProuse (commented):
@dbatiste Hi Dave, Justin passed along that you'd sent a website to him previously to help clean up SVGs to match the core SVGs. Do you have that on hand? Or are these SVGs are fine? I'll be adding more for the context menu in another branch.

dbatiste:
> @dbatiste Hi Dave, Justin passed along that you'd sent a website to him previously to help clean up SVGs to match the core SVGs. Do you have that on hand? Or are these SVGs are fine? I'll be adding more for the context menu in another branch.

Yep... svgomg, some notes and link to it can be found here:
https://github.com/BrightspaceUI/core/tree/main/components/icons#updating-or-contributing-new-icons

dbatiste:
The doc I linked is intended for icons. I know that these aren't icons, and you'll probably have to ignore the color guideline. I might suggest inquiring with your designer if there are colors from [our Daylight palette](https://daylight.d2l.dev/style-elements/colors/) that can be used, so at least these illustrations will be "on palette". Even though these aren't icons, I'd still try to follow the guidelines to keep the markup clean.

components/d2l-layouts.js
```diff
@@ -22,7 +22,7 @@ const layoutsHtml = function(localize) {
 			id: 2,
 			label: localize('layouts:twoThirdPanels:label'),
 			description: localize('layouts:twoThirdPanels:description'),
-			image: html`<svg slot="illustration" xmlns="http://www.w3.org/2000/svg" width="184" height="71" viewBox="0 0 184 71" fill="none"><rect x="0.5" y="0.5" width="183" height="70" rx="5.5" stroke="#CCCCCC"/><rect x="6.5" y="6.5" width="107" height="58" rx="5.5" fill="#E1E9EB" stroke="#CCCCCC"/><rect x="120.5" y="6.5" width="57" height="58" rx="5.5" fill="#E1E9EB" stroke="#CCCCCC"/></svg>`,
+			image: html`<svg xmlns="http://www.w3.org/2000/svg" width="184" height="71" viewBox="0 0 184 71" fill="none"><rect x="0.5" y="0.5" width="183" height="70" rx="5.5" stroke="#CCCCCC"/><rect x="6.5" y="6.5" width="107" height="58" rx="5.5" fill="#E1E9EB" stroke="#CCCCCC"/><rect x="120.5" y="6.5" width="57" height="58" rx="5.5" fill="#E1E9EB" stroke="#CCCCCC"/></svg>`,
```
dbatiste, line 25: What do the actual layouts look like when using a RTL language? Would these illustrations need to be flipped in RTL to match? 

I think we probably would want the image to match the rendered layout with whatever writing direction the user happens to be using. If the layout is insensitive to the writing direction (like float-based layouts) then we would not want it flipped. If the layout is sensitive to the writing direction (like flex or grid based layouts), then I think we would want it flipped. 

If using `d2l-icon-custom`, we can handle that automagically for you when the SVG has the `mirror-in-rtl="true"` attribute. Of course, there are other ways to deal with it.

**LTR language:**
image

**RTL language:**
image

JakeProuse: Yes all our layouts when they're dropped are using css grid so we'd want them to be flipped. I'll add that to the root svg elements 👍 Good catch Dave

JakeProuse: I've changed to using d2l-icon-custom and just overriding the size with css 👍 I've also updated some of the icons to be 18x18 instead of 20x18 like some of them were. Our designer had some issues making our 3-panel icon as a 18x18 that wasn't blurry, so we're going to be using a 17x18 instead for the moment until it's revisited. But all the others should be sized correctly and with daylight colours now!

JakeProuse: When I change the dir from ltr to rtl on the root <html> tag I see my SVGs in the drop menu don't flip 🤔 Am I missing an extra piece? Also I've targeted the margin of the heading/supporting info with the rule `:host([dir="rtl"]) .d2l-layout-list-item-info-label` but it doesn't get applied when I change the dir to rtl. Is that how I should be referencing it?

Context, I'm switching it without reloading the page, so I'm not sure if that changes things.
image
image

components/plugins/layouts.js
```diff
@@ -153,6 +153,44 @@ class LayoutsDialog extends RequesterMixin(Localizer(LitElement)) {
...
+			display: flex;
+			width: 100%;
+		}
+
+		.d2l-layout-list-item {
+			flex: 0 0 30%;
+		}
+
+		.d2l-layout-list-item-info-label {
+			margin-left: 24px;
```
dbatiste, line 167: Normally this would require that this component extends `RtlMixin`, and this margin would need to be flipped in RTL.

```css
:host([dir="rtl"]) .d2l-layout-list-item-info-label {
    margin-left: 0;
    margin-right: 24px;
}
```
But you can just use the flex box's `gap` property to manage the spacing between flex elements.

components/plugins/layouts.js
```diff
@@ -153,6 +153,44 @@ class LayoutsDialog extends RequesterMixin(Localizer(LitElement)) {
...
+				flex: 1;
+				justify-content: flex-start;
+			}
+
+			.d2l-layout-list-item-image {
+				flex: 0 0 50%;
+			}
+
+			.d2l-layout-list-item-info-label {
+				font-size: 18px;
```
dbatiste, line 187: We should use our typography styles. That involves:

* importing the exported styles from [typography/styles.js](https://github.com/BrightspaceUI/core/blob/main/components/typography/styles.js) 
* including the required block of CSS in this component's style block
* and applying either the `d2l-body-standard`, `d2l-body-compact`, `d2l-body-small` class to these elements depending on which size we want 

These styles will already have the responsive styles baked in for smaller viewports.

components/plugins/layouts.js
```diff
@@ -178,11 +216,6 @@ class LayoutsDialog extends RequesterMixin(Localizer(LitElement)) {
 	connectedCallback() {
 		super.connectedCallback();
 		this._context = requestInstance(this, 'context');
-
```
dbatiste, line 181: What was the reason for moving this? I think the original idea was that if the component was ever disconnected and then reconnected to the DOM, the observe would still work.

JakeProuse: Yeah I talked briefly about this with Grant after noticing it didn't attatch the observer, and the dialog doesn't exist at that point like it does in firstUpdated so it throws an error. For now he said we can accept that removing the component from the DOM will break the resizing.

components/plugins/layouts.js
```diff
@@ -206,25 +243,27 @@ class LayoutsDialog extends RequesterMixin(Localizer(LitElement)) {
...
-									<div slot="supporting-info">${layouts[layoutName].description}</div>
-								</d2l-list-item-content>
-							</d2l-list-item-button>
-						</d2l-list>
-					</div>
-				</div>`))}
+					${ this._mobileView ? html`<p><strong class="d2l-layout-mobile-instructions">${this.localize('layouts:mobile')}</strong></p>` : '' }
+					<d2l-list>
+						${Object.keys(layouts).map((layoutName) => (html`
+						<d2l-list-item-button @d2l-list-item-button-click="${this._handleClose}" key="${layoutName}" label="${layouts[layoutName].label}" value="${layoutName}">
```
dbatiste, line 255: This is sort of a minor thing, but it looks like some of the conditional logic in `_handleClose` could be simplified with a little refactoring and separate event handlers, `_handleClose` and `_handleListItemButtonClick`.

components/plugins/layouts.js
dbatiste: Nit: I think it would be good to choose a different class name other than `d2l-editor` to avoid confusion with the html-editor itself.

components/plugins/layouts.js
dbatiste, line 1: Is this needed? I don't see `d2l-focus-trap` imported or used anywhere in this component. Maybe this is left over from some previous changes?

```css
d2l-focus-trap {
	width: 100%;
}
```

JakeProuse: Yeah that's a leftover from using InsertElement as a base, removed 👍

components/d2l-layouts.js
dbatiste: > I'm switching it without reloading the page, so I'm not sure if that changes things.

Yeah I think it does. I think the `RtlMixin` (which these components need to extend in order to use `:host([dir="rtl"])`) doesn't observe subsequent changes to `dir`. You might be able to trick it into updating itself by also subsequently updating the language of the page. Usually I am working in the demo pages where it's easy to just this this before loading the page. The LMS provides a way to change the language, but it's not quite as convenient. Let me know if you still have troubles testing this and I can help you with it.

JakeProuse: Ah yes changing the language did it after I extended it with RtlMixin, and the SVGs flipped 🎉

dbatiste (approved):

PR #1298, opened 2023-03-14
Remove forceFocusVisible

dlockhart:
We're not a super huge fan of `forceFocusVisible`, and would like to remove it entirely. It's forcing us to maintain a duplicate set of styles for the `.focus-visible` CSS class separate from the correct `:focus-visible` pseudo-class styles.

Inside visual-diff tests, it's no longer necessary now that `focusWithKeyboard` exists -- this PR switches HTML Editor's visual-diff tests to use that instead.

Outside of visual-diff tests, as long as the user is interacting with the keyboard it also shouldn't be necessary, as calls to the regular `focus()` will show the `:focus-visible` styles. I've switched the few places that were using it over to `focus()` directly.


components/focus-helpers.js
```diff
@@ -19,7 +17,7 @@ export function resetActiveFocusable(focusables) {
 export function setFocus(focusables) {
 	const item = focusables.find(item => item.activeFocusable);
 	if (!item) return;
-	forceFocusVisible(item);
```
dlockhart, line 22: In my testing, I actually couldn't get this method to ever get called. It's used in a few of the toolbars' `focus()` method, but I'm guessing those toolbars themselves never receive focus directly?

components/focus-helpers.js
```diff
@@ -28,9 +26,7 @@ export function setFocusOnKeyDown(focusEvent, focusables, dir) {
 	const setActiveFocusable = async focusable => {
 		focusable.activeFocusable = true;
 		await focusable.updateComplete;
-		requestAnimationFrame(() => {
-			forceFocusVisible(focusable);
-		});
+		requestAnimationFrame(() => focusable.focus());
```
dlockhart, line 29: A keyboard event is already the source of this, so `focus()` works just fine.

components/footer/word-count-button.js
```diff
@@ -5,11 +5,12 @@ import { getEditor, getEditorComponent } from '../editor-adapter.js';
 import { buttonStyles } from '../button-styles.js';
 import { classMap } from 'lit/directives/class-map.js';
 import { cmds } from '../commands.js';
+import { FocusMixin } from '@brightspace-ui/core/mixins/focus-mixin.js';
 import { footerCountType } from '../plugins/wordcount.js';
 import { Localizer } from '../../localizer.js';
 import { RtlMixin } from '@brightspace-ui/core/mixins/rtl-mixin.js';
 
-class WordCountButton extends Localizer(RtlMixin(LitElement)) {
+class WordCountButton extends FocusMixin(Localizer(RtlMixin(LitElement))) {
```
dlockhart, line 13: One difference is that `forceFocusVisible` will search for focusable things inside, so the word count needed to redirect focus to its button.

htmleditor.js
```diff
@@ -551,7 +551,7 @@ class HtmlEditor extends HtmlEditorMixin(SkeletonMixin(LitElement)) {
 			}
 		}
 		const firstFocusable = getNextFocusable(this.shadowRoot.querySelector('.d2l-htmleditor-fullscreen-trap-start'));
-		forceFocusVisible(firstFocusable);
```
dlockhart, line 554: I couldn't ever get this code to run, since the target was always a composed ancestor and the block above always ran.

htmleditor.js
```diff
@@ -538,7 +538,7 @@ class HtmlEditor extends HtmlEditorMixin(SkeletonMixin(LitElement)) {
 	_handleTrapEndFocusIn() {
 		// user is tabbing forwards...
 		const firstFocusable = getNextFocusable(this.shadowRoot.querySelector('.d2l-htmleditor-fullscreen-trap-start'));
-		forceFocusVisible(firstFocusable);
+		firstFocusable.focus();
```
dlockhart, line 541: Keyboard is the source when the focus trap is engaged, so `focus()` works.

eKoopmans (commented):
Looks reasonable to me, any thoughts why the visual-diffs are failing?

components/focus-helpers.js
grant-cleary: They may not anymore, but I think it's likely we're going to need this for some changes being planned in the near-ish future.

grant-cleary: For some of our workflows we may need the ability to programmatically reset focus to the toolbar instead of the editor window when certain buttons are pressed, more or less. Whether it would end up using this function specifically is something I'm less certain about, since I haven't really delved too far into exploring solutions.

dlockhart: As long as that focus "reset" originates from a keyboard event in the first place (i.e. the user hitting a key on the keyboard), `focus()` on its own works great.

grant-cleary: Oh, yeah, sorry, I didn't mean I thought we needed `forceFocusVisible` here, just that this _function_, even if it isn't hit, will probably be used shortly.

dlockhart: Ah ok cool, no problem them!

test/toolbar.visual-diff.js
```diff
@@ -43,38 +43,27 @@ describe('d2l-htmleditor-toolbar', () => {
...
+			{ name: 'arrow-right', focusableIndex: 2, action: () => page.keyboard.press('ArrowRight') },
+			{ name: 'arrow-left', focusableIndex: 2, action: () => page.keyboard.press('ArrowLeft') },
+			{ name: 'wrap-to-last', focusableIndex: 0, action: () => page.keyboard.press('ArrowLeft') },
+			{ name: 'wrap-to-first', focusableIndex: 5, action: () => page.keyboard.press('ArrowRight') },
+			{ name: 'home', focusableIndex: 2, action: () => page.keyboard.press('Home') },
+			{ name: 'end', focusableIndex: 2, action: () => page.keyboard.press('End') }
 		].forEach(info => {
 			it(info.name, async function() {
+				await focusWithKeyboard(page, '#toolbar > d2l-htmleditor-toolbar-test');
+				await page.keyboard.press('Home');
```
dlockhart, line 59: This "resets" focus to the first element, then arrows right however many times we need to get to the desired "focusableIndex". We can then press whatever key we're interested in testing.

eKoopmans (approved):

grant-cleary (approved):
LGTM. Thanks Dave!

PR #1319, opened 2023-03-23
Protect fullpage plugin from trying to set bad attributes

grant-cleary:
If someone sets an invalid attribute on some HTML element (e.g. `<div \="val"></div>`), then `div.attributes` will happily return the invalid attribute in the returned `AttributeList`. Unfortunately, trying to then set that attribute elsewhere via `setAttribute` will throw, as the attribute name is invalid. As such, we need to make sure to wrap these cases in a try-catch to prevent these bad values from propagating.


dbatiste (approved):
LGTM!

PR #1323, opened 2023-03-24
Force a new minor version to be created on the first commit after an LMS update

grant-cleary:
Does what it says on the tin. Forcing a minor release on the first commit post-LMS-update makes backports easier to manage with very little downside.


.github/workflows/release.yml
```diff
@@ -7,7 +7,6 @@ on:
       - '[0-9]+.[0-9]+.x'
 jobs:
   release:
-    if: "!contains(github.event.head_commit.message, 'skip ci')"
```
grant-cleary, line 10: Nothing used this and I don't think it has any value here.

dbatiste (approved):

PR #1327, opened 2023-03-27
Fix: removing console log

jschlumkoski:


components/plugins/creator-plus.js
```diff
@@ -155,7 +155,6 @@ export const setCreatorPlusTemplate = (editor, styleSheets) => {
 		const editorCSSUrl = 'https://templates.lcs.brightspace.com/lib/assets/css/editor.css';
 
 		if (!editor.contentDocument.head.querySelector(`link[href="${editorCSSUrl}"]`)) {
-			console.log('adding');
```
grant-cleary, line 158: I'm actually a bit surprised this made it through linting. Maybe we relaxed our rules around logging to console.

PR #1341, opened 2023-03-31
Jprouse/us148303 layouts fable testing

JakeProuse:
Rally Story: https://rally1.rallydev.com/#/?detail=/userstory/684998877825&fdp=true

Solving issues from our Fable Testing.


components/plugins/layouts.js
```diff
@@ -155,9 +156,23 @@ export function addLayoutsPlugin() {
 
 			for (const layout of layouts) {
 				layout.contentEditable = false;
+				layout.setAttribute('tabindex', '-1');
 
 				for (const column of layout.children) {
 					column.contentEditable = true;
+					column.setAttribute('tabindex', '-1');
+					column.onkeydown = e => {
+						if (e.keyCode !== 9) return;
```
grant-cleary, line 165: Does shift-tab behave as expected in here?

JakeProuse: Should shift-tab bring you to the toolbar? Looks like I'm capturing that so I'll need to add a case to ignore it

JakeProuse: Ah similar to the original tab when I ignore the shift key, shift tab will bring you above the layout, then you'd need to shift-tab again to hit the toolbar.

JakeProuse: Also able to intercept that so it behaves as normal. I hate to code around things like this, it does make it kind of a pain if we added more shortcuts in the future.

grant-cleary: Yeah, hopefully we can find a better way to get this working in the future. Right now this is just the limitation of what we have available.

grant-cleary (approved):

PR #1346, opened 2023-04-03
Use update html-block API for displaying HTML

grant-cleary:
The html-block component has been updated with an `html` attribute that can/should be used to render HTML internally. This PR swaps over the editor's implementation (both when in `disabled` mode and when in inline readonly mode) to use this new attribute.


htmleditor.js
```diff
@@ -611,9 +610,7 @@ class HtmlEditor extends HtmlEditorMixin(SkeletonMixin(LitElement)) {
 	_renderDisabledHtmlBlock() {
 		return html`
 			${this._renderLabel()}
-			<d2l-html-block class="d2l-skeletize">
-				${this._renderProcessedHtml()}
-			</d2l-html-block>`;
+			${guard([this._html], () => this._renderProcessedHtml(false))}`;
```
grant-cleary, line 613: I left in the use of `guard` but made it specific to each place this renders, since we now have to gate the rendering of the entire html-block component, not just the template within it. I could have centralized this, but I also don't want to update on more (or fewer) properties than needed.

htmleditor.js
```diff
@@ -730,9 +727,7 @@ class HtmlEditor extends HtmlEditorMixin(SkeletonMixin(LitElement)) {
 					${this.localize('inline:button', { name: this.label || '' })}
 				</button>
 				<div class="${classMap(htmlBlockClasses)}" style="${styleMap(htmlBlockStyles)}" @click="${this._onInlineEditClick}">
-					<d2l-html-block class="d2l-skeletize" ?compact="${this.compact}">
-						${this._renderProcessedHtml()}
-					</d2l-html-block>
+					${guard([this._html, super.compact], () => this._renderProcessedHtml(this.compact))}
```
grant-cleary, line 730: Interestingly, `super` here seems to be required. The `guard` directive doesn't consider updates to `this.compact` as fulfilling the watch condition.

htmleditor.js
```diff
@@ -751,8 +746,11 @@ class HtmlEditor extends HtmlEditorMixin(SkeletonMixin(LitElement)) {
 			: '';
 	}
 
-	_renderProcessedHtml() {
-		return html`${guard([this._html], () => until(this._getProcessedHtml(), ''))}`;
+	_renderProcessedHtml(compact) {
+		return html`${runAsync(this._html, () => this._getProcessedHtml(), {
```
grant-cleary, line 750: Switching to use core's `runAsync` directive instead of rolling my own solution with Lit's `until` directive. `until` was difficult to work with because it expects a list of templates to render in priority order, so the logic to have it wait on `this._getProcessedHtml()` was a bit messy.

PR #1353, opened 2023-04-05
adjust to new location of mixins

dlockhart:
We've recently moved mixins in core into their own directories. We've left the old files in place to not break anything, but this is the only place outside of core using the `theme-mixin.js` and `visible-on-ancestor-mixin.js` so I've fixed that up here so that we can just remove its redirect.


package.json
```diff
@@ -47,7 +47,8 @@
     "eslint-config-brightspace": "^0.20.1",
     "lit-analyzer": "^1",
     "messageformat-validator": "^2.0.2",
-    "stylelint": "^15"
+    "stylelint": "^15",
+    "typescript": "^5"
```
dlockhart, line 51: Had to add this hack in temporarily [because of this issue](https://d2l.slack.com/archives/C0PHG3QB0/p1680714255203899).

grant-cleary: Oh, I was just planning on killing the analyzer because it's so old, but this works too. 😆

grant-cleary (approved):

package.json
dlockhart: Definitely rip it out if you're not really using it for anything!

PR #1355, opened 2023-04-06
Fix editor resizer a11y for wcag 2.2.

dbatiste:
[DE51626](https://rally1.rallydev.com/#/?detail=/defect/675766826327&fdp=true)

This PR updates the `d2l-htmleditor-resizer` component to render the `d2l-button-move` component when in keyboard mode. This addresses two issues - it enables keyboard users to resize the editor, and it also enables mouse users to resize the editor without dragging. When the user focuses on the resizer or clicks on it, it will switch to the non-dragging mode.

A couple points worth mentioning...

* the `d2l-button-move` will be enabled/disabled when the editor reaches its min/max dimensions
* hitting Escape from keyboard mode will move focus to the editor

**Dragging mode:**
image

**Non-dragging mode (light):**
image

**Non-dragging mode (dark):**
image


components/footer/resizer.js
```diff
@@ -1,14 +1,27 @@
+import '@brightspace-ui/core/components/icons/icon.js';
```
dbatiste, line 1: This was a missing import.

components/footer/resizer.js
```diff
@@ -17,72 +30,120 @@ class Resizer extends ThemeMixin(RtlMixin(LitElement)) {
 				fill: var(--d2l-htmleditor-toolbar-dark-button-text, var(--d2l-color-regolith));
 			}
 			.d2l-htmleditor-resize-handle {
-				cursor: nwse-resize;
+				cursor: ns-resize;
```
dbatiste, line 33: Unrelated to my changes - this was spotted by someone on my team... should just be up-down cursor since it's no longer possible to resize in the horizontal direction.

components/footer/resizer.js
```diff
@@ -17,72 +30,120 @@ class Resizer extends ThemeMixin(RtlMixin(LitElement)) {
...
 				margin-left: -1px;
 				margin-right: 0;
 				padding-left: 0;
 				padding-right: 6px;
 			}
-			.d2l-htmleditor-resize-handle-fullscreen {
-				display: none;
+			d2l-button-move {
+				bottom: 5px;
+				position: absolute;
```
dbatiste, line 47: This is absolute because we don't want it to cause the footer to get taller. We know there is an edge case where this could cover a few characters of text in the editor - we are "ok" with that since this mode can only be active when the user has focus on it.

components/footer/resizer.js
```diff
@@ -17,72 +30,120 @@ class Resizer extends ThemeMixin(RtlMixin(LitElement)) {
...
+		}
+	}
 
-		return html`
-			<div class="${classMap(resizeHandleClasses)}" @mousedown=${this._handleResizeStart} @touchstart=${this._handleTouchResizeStart}>
+	render() {
+		let content;
+
+		if (this._keyboardActive) {
+			const minHeight = this._context.component._getMinHeight();
```
dbatiste, line 78: We use these values below to toggle the `d2l-button-move`'s fine-grained `disabled-*` properties.

components/footer/resizer.js
```diff
@@ -17,72 +30,120 @@ class Resizer extends ThemeMixin(RtlMixin(LitElement)) {
...
-		window.removeEventListener('mouseup', this._handleResizeEnd, false);
 	}
 
-	_handleResizeStart(e) {
+	_handleDraggerMouseUp(e) {
 		e.preventDefault();
-
-		window.addEventListener('mouseup', this._handleResizeEnd, false);
-
+		if (e.pageY === this._mouseY) this._activateKeyboard = true;
```
dbatiste, line 137: The click event will be dispatched, but we don't want to toggle into keyboard mode if the user has dragged the mouse.

htmleditor.js
```diff
@@ -349,7 +350,12 @@ class HtmlEditor extends HtmlEditorMixin(SkeletonMixin(LitElement)) {
 			const inlineButton = this.shadowRoot.querySelector('.d2l-htmleditor-inline-button');
 			inlineButton.focus();
 		} else {
-			super.focus();
+			if (this._overlayOpen) {
```
dbatiste, line 353: This appears to have been a gap in the `focus` logic. When an overlay is open and `focus` is called on the editor, I think we want the focus directed to the overlay. I'm not sure if there is a better way to deal with this - currently it assumes it can focus the first element child of the overlay container, which appears to work nicely for the code sample editor.

htmleditor.js
```diff
@@ -376,6 +382,12 @@ class HtmlEditor extends HtmlEditorMixin(SkeletonMixin(LitElement)) {
 		this.focus();
 	}
 
+	_getHeight() {
```
dbatiste, line 385: This was just factored out so it could be called from a couple spots.

htmleditor.js
```diff
@@ -753,6 +760,15 @@ class HtmlEditor extends HtmlEditorMixin(SkeletonMixin(LitElement)) {
 		}, { pendingState: false })}`;
 	}
 
+	_setHeight(height) {
```
dbatiste, line 763: Just factored out so it could be called from more than one spot.

components/footer/resizer.js
```diff
@@ -17,72 +30,120 @@ class Resizer extends ThemeMixin(RtlMixin(LitElement)) {
...
+				?disabled-up="${minHeight && minHeight + 2 >= height}"
+				?disabled-down="${maxHeight && height >= maxHeight - 2}"
+				@focusout="${this._handleKeyboardFocusOut}"
+				@keydown="${this._handleKeyboardKeydown}"
+				@mousedown="${this._handleKeyboardMouseDown}"
+				text="${this.localize('resize')}"
+				theme="${this.theme}">
+			</d2l-button-move>`;
+		} else {
+			content = html`<div
```
grant-cleary, line 83: So the idea is that we only use the move button for keyboard nav explicitly?

components/footer/resizer.js
```diff
@@ -17,72 +30,120 @@ class Resizer extends ThemeMixin(RtlMixin(LitElement)) {
...
-		this.dispatchEvent(new CustomEvent(
-			'd2l-htmleditor-resize-end', {
-				bubbles: true,
-				composed: true
+	willUpdate(changedProperties) {
+		if (changedProperties.has('_keyboardActive')) {
+			if (this._keyboardActive) {
+				this._context = requestInstance(this, 'context');
+			} else {
+				this._context = null;
```
grant-cleary, line 107: So we're using the `_context` object to determine whether we're in keyboard mode for some of the event handling?

htmleditor.js
grant-cleary: Yeah, it's a bit of a weird case. In inline mode it's moot because the overlay closes as soon as focus is removed from the editor. For full mode I don't think this use case was considered mostly because the only consumers that apply programmatic focus to the editor tend to do it either on init or post-save.

grant-cleary: I'm not opposed to handling it this way, but if consistency is the game, other places use the `focus()` function on the overlay component itself: https://github.com/BrightspaceUI/htmleditor/blob/main/components/overlay/overlay.js#L228

components/footer/resizer.js
grant-cleary: Ah I see, because the button component is larger. Yeah, I agree with this. Shifting the footer's height could end up creating a scrollbar in the editor, which would then also shift the editor's horizontal layout, etc.

components/footer/resizer.js
dbatiste: There are the two use cases:
* keyboard user
* mouse user who cannot or does not want to drag - they click to switch to the `d2l-button/move`, and then they can click the arrows inside the button to resize/move up or down.

components/footer/resizer.js
dbatiste: We're just using the context object in order to get the dimensions so that we can set `disabled-up` or `disabled-down` on `d2l-button-move` to disable those specific interactions.

dbatiste: Oh... and we're also using it to apply focus back to the editor when escape is pressed.

grant-cleary: Ah yeah, I asked only because I was curious why we were requesting it (or nulling it out) every time this property changed, instead of just fetching it once and then using the property to figure out what to do.

grant-cleary (approved):

PR #1374, opened 2023-04-18
Remove Number wrapper from test input

grant-cleary:
This wrapper is redundant and likely to conflict with a new incoming linting rule.


test/htmleditor.test.js
```diff
@@ -106,7 +106,7 @@ describe('d2l-htmleditor', () => {
 		[
 			{ name: 'undefined', html: undefined },
 			{ name: 'null', html: null },
-			{ name: 'not a string', html: new Number(10) }
+			{ name: 'not a string', html: 10 }
```
PaulKrznaric, line 109: I think 11 is better

PaulKrznaric (approved):

test/htmleditor.test.js
grant-cleary: You're fired.

PR #1408, opened 2023-05-05
Add non-breaking space within XSPL placeholder span so TinyMCE won't strip it out

grant-cleary:
Under some circumstances, TinyMCE appears to strip empty span tags. One way I can reliably reproduce this is via directly inserting one into an otherwise empty editor, though I've found the behaviour to be a bit inconsistent. As a workaround, a quick and dirty solution is to force in a non-breaking space so TinyMCE will think the span has something in it.


components/plugins/xspl.js
```diff
@@ -30,6 +30,7 @@ export function convertIframeToImagePlaceholder(htmlDoc) {
 		placeholder.classList.add('d2l-xspl-image-placeholder');
 		placeholder.style.display = 'inline-block';
 		placeholder.contentEditable = 'false';
+		placeholder.innerHTML = String.fromCharCode(160); // &nbsp; so TinyMCE won't remove the empty span.
```
grant-cleary, line 33: I'm a bit concerned about the a11y implications here for screenreaders, but in hindsight, none of this XSPL stuff is or has been accessible from the start... So I think that's a separate conversation we should have, but it's not necessarily relevant to this PR.

grant-cleary:
I'm still a bit iffy about using a workaround this sketchy. I did try more legitimate solutions like putting a `br` tag in there, or another set of empty `span` tags, but unfortunately TinyMCE just strips those out too.

dbatiste (approved):

PR #1420, opened 2023-05-11
Jprouse/us151629 remove old wrapper in iframe srcdoc for preview

JakeProuse:
Rally Story: https://rally1.rallydev.com/#/?detail=/userstory/698303613313&fdp=true


components/plugins/lcs.js
```diff
@@ -800,7 +815,7 @@ class ElementEditorDialog extends Localizer(ProviderMixin(LitElement)) {
 
 				const templateHTML = this._buildHTML(template, selector);
 
-				const iframeStr = `<div class="container-fluid template-preview"><div class="row"><div class="col-12">${templateHTML}</div></div></div>${scriptsStr}${resizeScript}`;
+				const iframeStr = `<div class="container-fluid template-preview">${templateHTML}</div>${scriptsStr}${resizeScript}`;
```
tonystecca, line 818: We don't want the container-fluid class or an intermediate div. HTML content is expecting to go in the root of the body tag.

tonystecca (changes requested):

grant-cleary (approved):

PR #1429, opened 2023-05-15
fix: Hotspot improvements

jschlumkoski:


jschlumkoski:
@grant-cleary This should hopefully be the last of the things needing backported for 20.23.6

grant-cleary (approved):

PR #1434, opened 2023-05-17
fix: Hotspot Improvements

jschlumkoski:


components/insertElement/hotspot.js
```diff
@@ -794,7 +799,7 @@ class HotspotEditor extends Localizer(LitElement) {
 		let val;
 
 		if (e.target.tagName === 'D2L-HTMLEDITOR') {
-			val = e.target.html;
+			val = e.target.html.replace(/([\u2700-\u27BF]|[\uE000-\uF8FF]|\uD83C[\uDC00-\uDFFF]|\uD83D[\uDC00-\uDFFF]|[\u2011-\u26FF]|\uD83E[\uDD10-\uDDFF])/g, '');
```
dbatiste, line 802: Nit: this might be worth a comment for those not knowing what these characters are.

dbatiste (approved):

PR #1442, opened 2023-05-18
Add aria-label and aria-role to XSPL iframe placeholder span

grant-cleary:
Our XSPL iframe placeholder isn't currently accessible to screenreader users, so we should give it a proper ARIA role and label so there's some contextual information about what this empty span actually represents in the editor.


lang/en.js
```diff
@@ -573,5 +573,6 @@ export default {
 	"wordcount:footerSelector:wordCount": "Word Count",
 	"wordcount:footerSelectorLabel:character": "Footer Displays: Character Count",
 	"wordcount:footerSelectorLabel:noCount": "Footer Displays: Nothing",
-	"wordcount:footerSelectorLabel:word": "Footer Displays: Word Count"
+	"wordcount:footerSelectorLabel:word": "Footer Displays: Word Count",
+	"xspl:iframeLabel": "Placeholder for an iframe with source URL: {src}"
```
dbatiste, line 577: I'm not sure what the best verbiage is. The "iframe" is kinda technical jargon, but maybe users will understand what this means. 🤷‍♂️

dbatiste (approved):

lang/en.js
grant-cleary: Yeah, Elaine gave this a thumbs up for now. I'm also not sure how best to identify this... I figure a screenreader user is at least going to have the name of the element read out to them under normal usage so perhaps it's a bit more reasonable? Unsure.

PR #1445, opened 2023-05-19
Bump TinyMCE lib version to get new Edit Image compatible skin and add a basic framework for flagging TinyMCE lib updates

grant-cleary:
We need to bump the custom skin we're using for TinyMCE to account for changes made within TinyMCE 6 (specifically to unbreak the Edit Image crop workflow). We also want this change to be controllable via a feature flag, so this PR also adds a basic framework to toggle between versions from a server-side context object.


htmleditor-mixin.js
```diff
@@ -349,7 +347,7 @@ export const HtmlEditorMixin = superclass => class extends Localizer(LabelledMix
 		this.noInsertCode = false;
 		this.noLiveValidation = false;
 		this.noSpellchecker = false;
-		this.pasteLocalImages = false;
+		this.pasteLocalImages = true;//false;
```
grant-cleary, line 350: Oops, forgot this was here for debugging.

htmleditor-test-context.js
```diff
@@ -0,0 +1,5 @@
+console.warn('Using htmleditor test context, this is intended for demo pages and tests only');
+
+document.getElementsByTagName('html')[0].dataset.heContext = JSON.stringify({
+	useNewTinyMCEVersion: !(window.location.search.indexOf('oldVersion=true') !== -1)
+});
```
grant-cleary, line 5: We want all tests to use the new version, and we want to be able to have both available on demo pages.

grant-cleary:
Visual diff tests are failing here. I'm going to have to poke at them to figure out why.

htmleditor-test-context.js
grant-cleary: Though actually, I think there's a good argument for having tests run against both sets to guard against breaking the old version in case we have to toggle it off. I will likely do that in a separate PR, however, since I think that's likely to be a beefier change.

tinymce-lib-version.js
```diff
@@ -1 +1,4 @@
-export const version = '1.0.0';
```
dbatiste, line 1: Remind me why this doesn't match the tinyMCE version?

grant-cleary: Conveniently, this bump is a good example of why it can't. This change updates the skin we use, but TinyMCE's version remains unchanged. There are other types of changes that can be made in the TinyMCE library repo that don't involve TinyMCE version bumps.

grant-cleary:
Oh, the visual diff failures are because TinyMCE is now applying a very faint border around the editing space. I'll need to override that within the editor.

dbatiste (approved):

PR #1463, opened 2023-05-24
Run htmleditor tests against both old and new versions of TinyMCE library

grant-cleary:
This PR sets up a nice and easy way to guarantee tests will run by default against both old and new versions of our TinyMCE library. I envision the need for a state where we want different tests to run against new and old versions, and such a framework would also be very easy to add. That said, I don't really want to over-engineer this specific piece; rather, that work can be easily done when it's needed.


test/tinymce-version-helper.js
```diff
@@ -0,0 +1,12 @@
+export function runTestsWithTinyMCEVersioning(test) {
```
grant-cleary, line 1: This is the big change here. Basically we just manually set the context object before running the tests in order to capture changes. I'm unsure whether this will work as expected for visual diff tests, but those are easy to handle differently.

grant-cleary: Okay no, the visual diff tests do not like this. I'll change those up a bit.

grant-cleary:
Hmm. Well the visual diff tests now generate goldens (and also I can drop the timeout back down), but because I've renamed all of the cases, it's almost impossible to work out whether anything has changed. I think I'm going to change up how the test naming works to avoid generating a new name in the case that we're using the new version.

PaulKrznaric:
Are visual diff tests failing to run or failing?

grant-cleary:
> Are visual diff tests failing to run or failing?

They're failing to complete because they're taking too long, but they'll still fail once they do complete. The issue is that all of the "old version" stuff needs to generate new goldens and that causes a failure.

That said, I'm a bit nervous how long this is taking even with retries off (and even generating so many goldens). The editor's visual diff tests took ~2 minutes before this change and now they're taking more than 30 minutes. I'm going to try increasing the timeout until we can get new goldens generated and then see if the run-time becomes more reasonable once those are merged in... But realistically, I'm starting to think we might not be able to cover the "old version" cases in visual diff.

grant-cleary:
Hmm, though on closer inspection, I think the tests might actually be hanging.

grant-cleary:
Oh, I think I see what's happening here. Regardless of where Mocha's `before` hook is defined, it only runs once per _file_, and it runs before any tests have run. So I think the tests are being bogged down because the `before` hook setting the URL (including the query parameter for enabling the old version) is only running once per file.

PR #1469, opened 2023-05-26
Retain focus on fullscreen button when entering/exiting fullscreen in inline mode

grant-cleary:
When we're in inline mode, it's not sufficient for us to wait for the toolbar to finish rendering to apply focus, as the toolbar isn't visible (and thus isn't focusable) until it's measured. Thus, we need to apply an additional check to make sure this has occurred before we can interact with the toolbar focus programmatically.

While this change applies to both the full and the inline toolbar, it has no real effect on the full case. The extra promise resolves immediately.


components/toolbar/toolbar-mixin.js
```diff
@@ -184,10 +192,10 @@ export const ToolbarMixin = superclass => class extends Localizer(RtlMixin(super
 
 		const container = this.shadowRoot.querySelector('.d2l-htmleditor-toolbar-container');
 		this._resizeObserver = new ResizeObserver(async entries => {
+			this._measuresResolve = undefined;
```
grant-cleary, line 195: Oh, whoops. This was in there for testing and shouldn't have carried through.

components/toolbar/toolbar-mixin.js
```diff
@@ -198,6 +204,7 @@ export const ToolbarMixin = superclass => class extends Localizer(RtlMixin(super
 				// wait to set measured to avoid FOUC
 				requestAnimationFrame(() => {
 					this._measured = true;
+					this._measuresResolve();
```
grant-cleary, line 207: This won't get reset if the toolbar resizes at any point, but in the cases we care about, that turns out not to be relevant, because the toolbar fully re-renders when toggling into or out of fullscreen. If we're concerned about future use of this promise leading to weird behaviour, I can reset it on resize as appropriate.

components/plugins/fullscreen.js
```diff
@@ -33,6 +33,7 @@ export function addFullscreenPlugin() {
 			await component.updateComplete;
 			const toolbar = component.getToolbar();
 			await toolbar.updateComplete;
+			await toolbar.measuresComplete;
```
dbatiste, line 36: Is there some reason we can't do this inside of the toolbar's `setFocus()` method. Seems reasonable that it should be able to wait. That way `measuresComplete` doesn't have to be surfaced on the toolbar's "public" API. That's kinda what we do in core with [FocusMixin](https://github.com/BrightspaceUI/core/blob/main/mixins/focus/focus-mixin.js) when we need to ensure that the shadowDOM has rendered before trying to apply focus internally.

components/toolbar/toolbar-mixin.js
dbatiste: On that note, is it really the "measures" being complete that we care about? i.e. does that best describe what we are waiting for? Would `firstRender` or `renderItemsComplete` (I know past tense would be inconsistent) be more what we care about? Like the measures being complete just happens to coincide. More just thinking about the name if this does need to be surfaced on the public API.

components/plugins/fullscreen.js
grant-cleary: Nope! This could totally be done in there. I was just thinking of how we implemented `focus` in the editor where we were a bit leery about making that async. Then again, that was a while ago. Easy enough for me to shove that in there.

components/toolbar/toolbar-mixin.js
grant-cleary: So... Yeah, technically it's a bit of both? What we're _actually_ waiting for is for the toolbar to become visible, which happens to occur once we're done measuring. I think `firstRender` is probably more accurate in that sense, since that sort of encapsulates the whole process of rendering, taking the measures, and then making the toolbar visible.

components/plugins/fullscreen.js
```diff
@@ -32,8 +32,7 @@ export function addFullscreenPlugin() {
 
 			await component.updateComplete;
 			const toolbar = component.getToolbar();
-			await toolbar.updateComplete;
-			toolbar.setFocus(cmds.fullscreen);
+			await toolbar.setItemFocus(cmds.fullscreen);
```
grant-cleary, line 35: I changed the name here because I realized that the toolbar mixin was already importing a `setFocus` function and then using it internally. I'm not entirely sure why that didn't explode anything, but :shrug:.

dbatiste (approved):

PR #1472, opened 2023-05-26
Make improvements to inline mode skeleton

grant-cleary:
The inline editor looked all sorts of weird when skeleton was enabled. See below for before/after images. Beyond looking strange, the inline buttons were enabled and the onclick functioned when skeletized, so it was possible to open the editor behind the skeleton. I couldn't easily get a screenshot of this because the loss of focus while snipping also set the editor back to readonly mode, but suffice it to say it looks pretty zany.

The solution here was to modify where the skeleton class was applied (and some of the underlying CSS) to more correctly block out the whole editor container, and then to disable the accessible button and onclick when the editor is skeletized.

I'm not 100% sure about disabling the button and I'm somewhat tempted to just not render it. I'm not sure what would seem more reasonable to a screenreader user, however, so I'm open to suggestions.

Before:
Screen Shot 2023-05-26 at 3 58 14 PM

After:
Screen Shot 2023-05-26 at 3 56 18 PM


dbatiste (approved):
Seems reasonable, and looks much better.
> I'm not 100% sure about disabling the button and I'm somewhat tempted to just not render it. I'm not sure what would seem more reasonable to a screenreader user, however, so I'm open to suggestions.

Disabled buttons are not focusable, so screen reader users would only be able to discover the button using the virtual cursor, which is valid/acceptable in this case. Not rendering the button would also be fine too.

PR #1476, opened 2023-05-31
Fix Previously Created C+ Pages With Elements

jschlumkoski:
@grant-cleary I'm looking into a user story we have for Insert Element where when a page is being edited, it checks for our template JS file being included in the body, and if so, it removes it and adds it to the head so that it's consistent with our template pages and helps with our FoUC situation when swapping markup into components. It's known that it'll only take if the page is saved again.

The main issue I'm up against is we're already using the TinyMCE `SetContent` event listener for Insert Element to tell TinyMCE to load our editor specific CSS. I was trying to use the part to update the script, but it puts me in a situation where I have the `SetContent` listener calling a function that could end up calling `SetContent` again. In theory, it would only end up calling `SetContent` once more because the second time it wouldn't need to do anything, but I understand that it's certainly not an ideal situation. I'm not sure how to avoid this. I _think_ listening for `BeforeSetContent` instead won't help, it just shifts where the problem is. Wondering if there's a better was to be tackling this? Is there a better way to intercept when the editor initially opens?

I know Content Styler listens for `BeforeSetContent` and uses your full page plug-in to add CSS. But as far as I'm aware, there's nothing existing to do this with scripts?


grant-cleary:
> The main issue I'm up against is we're already using the TinyMCE `SetContent` event listener for Insert Element to tell TinyMCE to load our editor specific CSS. I was trying to use the part to update the script, but it puts me in a situation where I have the `SetContent` listener calling a function that could end up calling `SetContent` again. In theory, it would only end up calling `SetContent` once more because the second time it wouldn't need to do anything, but I understand that it's certainly not an ideal situation. I'm not sure how to avoid this. I _think_ listening for `BeforeSetContent` instead won't help, it just shifts where the problem is. Wondering if there's a better was to be tackling this? Is there a better way to intercept when the editor initially opens?

If you're in a situation where one operation needs to happen first, using `BeforeSetContent` can be helpful. I believe there's a parameter that can be passed into `BeforeSetContent` handlers to tell TinyMCE not to invoke any `SetContent` listeners on its contents. That may not be suitable for what you're trying to do though.

> I know Content Styler listens for `BeforeSetContent` and uses your full page plug-in to add CSS. But as far as I'm aware, there's nothing existing to do this with scripts?

There isn't, but such a thing could certainly exist! Ideally if we could do this via the full-page plugin I think that would be ideal.

grant-cleary:
I'll also add that if this is an urgent change, what you have now is likely fine. Ideally we want to minimize directly playing with TinyMCE's DOM, but given that the new plugin framework stuff isn't really fleshed out, I also don't really expect you to wait on work I'm doing to get basic stuff done. If all you need is a similar mechanism for scripts as we currently use for CSS, that's probably something I can whip up fairly quickly for you to play around with.

jschlumkoski:
> I'll also add that if this is an urgent change, what you have now is likely fine. Ideally we want to minimize directly playing with TinyMCE's DOM, but given that the new plugin framework stuff isn't really fleshed out, I also don't really expect you to wait on work I'm doing to get basic stuff done. If all you need is a similar mechanism for scripts as we currently use for CSS, that's probably something I can whip up fairly quickly for you to play around with.

It's not super urgent, it's more of a case where it's better to do it ASAP because the longer we wait, the more pages there will be to try and correct! For the most part, something that functioned like what we have for CSS would be fine, it's just that this specific use case needs to do some clean up. 😅

jschlumkoski:
@grant-cleary I've adjusted this PR with a change that I think makes this functionality a bit safer until we can replace it with something in the full-page plugin (like what CSS is doing).

I've changed it so that if / when we need to call `setContent()`, we now pass in an arguments object with an `ignore` property, and in our event listener we ignore doing the function call if it's set.

grant-cleary (approved):

PR #1486, opened 2023-06-06
Set iframe body label ourselves for consistency

grant-cleary:
We already set the iframe title directly here, so we should be consistent and set the body `aria-label` attribute the same way. This also aligns with how we set these values when the label is updated after first render.


mpharoah-d2l:
Looks like we both put up PRs at the same time. I think this one won't update the body if the label is changed after initialization.

mpharoah-d2l (approved):

PR #1487, opened 2023-06-06
US153429 -- Avoid using insecure iframe_aria_text property of tinymce

mpharoah-d2l:
https://rally1.rallydev.com/#/57732444928d/custom/373260458992?detail=%2Fuserstory%2F703481205741

`iframe_aria_text` is vulnerable to XSS, so set it ourselves instead


grant-cleary:
Ah, thank you for the PR, but I think ideally we want to implement this a bit differently: https://github.com/BrightspaceUI/htmleditor/pull/1486

PR #1491, opened 2023-06-08
Creator+ RTL Updates

jschlumkoski:


components/insertElement/hotspot-icons.js
dbatiste, line 1: Not related to your changes, but I find I can get 15-30% savings (more in a couple cases) running most of these icons through svgomg. Let me know if you are interested in the settings I am using (which is what we use for core icons as well)

components/plugins/lcs.js
```diff
@@ -175,10 +177,11 @@ export function addElementEditorDialogPlugin() {
 	return name;
 }
 
-class ElementEditorDialog extends Localizer(ProviderMixin(LitElement)) {
+class ElementEditorDialog extends Localizer(ProviderMixin(RtlMixin(LitElement))) {
 	static get properties() {
 		return {
 			opened: { type: Boolean },
+			directionality: { type: String },
```
dbatiste, line 184: Is this set by anything else, or is this meant to be internal. If internal, it can be defined as `_directionality: { state: true }`, indicating that it is a private reactive property. 

But in this case, do we need this when we already have `this.dir`? I would typically check `this.dir === 'rtl'` where needed, because when we're in RTL, the `dir` is guaranteed to be set, which is not the case for LTR.

components/insertElement/hotspot.js
```diff
@@ -33,6 +33,7 @@ class HotspotEditor extends Localizer(LitElement) {
 		return {
 			activityData: { type: Object },
 			context: { type: Object },
+			directionality: { type: String },
```
dbatiste, line 36: It appears the value would always be coming from `RtlMixin` in the consuming component. Normally we would just extend `RtlMixin` on this component as well, avoiding the need for adding directionality property to the public API of this component.

components/practice/sorting.js
```diff
@@ -539,7 +555,7 @@ class SortingPractice extends RequesterMixin(Localizer(LitElement)) {
 				</div>
 				<div class="d2l-categories">
 					<d2l-scroll-wrapper>
-						<div class="d2l-category-list" role="list" style="width: ${(this.activityData.categories.length * itemWidth) - 1}px;">
+						<div class="d2l-category-list ${this.dir && this.dir === 'rtl' ? 'rtl' : 'ltr'}" role="list" style="width: ${(this.activityData.categories.length * itemWidth) - 1}px;">
```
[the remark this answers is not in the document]
jschlumkoski: @dbatiste This is just adding a class to help assist with fixing a border situation. It accompanies this piece:
https://github.com/BrightspaceUI/htmleditor/blob/2007cf849e99b621ff9906fe32f033dbd099f6a4/components/practice/sorting.js#L124-L130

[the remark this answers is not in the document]
dbatiste: I see. Sorry, I glossed over the class.

Because this is extending `RtlMixin`, the `dir` will be sprouted and we would normally write this:
```css
.d2l-category:last-child {
    // LTR style
}
:host([dir="rtl"]) .d2l-category:last-child {
    // RT: style
}
```

components/insertElement/hotspot.js
jschlumkoski: @dbatiste I imported it in `lcs.js` initially so all the elements could benefit from the property value. I was going to extend it in `hotspot.js` also, but it seemed a bit odd to do it just to get a value I technically already have (hence the property expression approach). But if the preferred approach is to extend it again, I can swap it.

dbatiste: Cool. Yeah, I'd keep the public API as clean as possible.

components/plugins/lcs.js
jschlumkoski: @dbatiste I did it as a property because of the piece that you mentioned around how the property is `undefined` if it's not set to RTL, and my preference would be for the property to always have a value, which is why I set it to be LTR as a fallback in my constructor. If the `undefined` nature of the property is expected, and the general preference is to just use the built in property, I can swap it.

components/insertElement/hotspot-icons.js
jschlumkoski: @dbatiste Yes, please pass along the settings and I can take a look!

components/practice/sorting.js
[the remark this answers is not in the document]
jschlumkoski: @dbatiste I was also unsure if we should be setting the `dir` attribute on DOM elements, I could swap this to utilize the attribute instead if that's preferred!

components/plugins/lcs.js
grant-cleary: Our components are generally built around the idea that `dir` can be undefined, yes. I think it's ideal to just rely on the sprouted property.

jschlumkoski: This should be fixed in my latest push!

components/insertElement/hotspot.js
jschlumkoski: This should be fixed in my latest push!

grant-cleary (approved):
Approved, but I would definitely look into running those SVGs through SVGOMG before merging. I usually use the options defined in this block: https://github.com/BrightspaceUI/core/tree/main/components/icons#svg-format

Unsure if Dave has any other suggestions there as well.

components/practice/sorting.js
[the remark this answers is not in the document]
dbatiste: Either way works, but we normally just target the existing `:host([dir="rtl"]) ...` since `RtlMixin` sprouts it already, then it's not necessary to render any special `dir` attributes within the component.

dbatiste (approved):

jschlumkoski:
@grant-cleary @dbatiste Just ran all the icons through SVGOMG using the settings Dave shared, so they should all be good to go now!

PR #1492, opened 2023-06-08
US147904 - Retain filenames when editing images via custom upload

PaulKrznaric:
When users edit image files in the HTML Editor, the editor retains the images' filename. 
Previously, the editor would replace the edited image's filenames with `editimage0_GUID.jpg`, files will now be of the format `originalname_timestamp.jpg`

This work is based on the addition of [this feature to TinyMCE 6](https://www.tiny.cloud/docs/tinymce/6/file-image-upload/#images_reuse_filename); however, due to the way that temp files work in the LMS, we were unable to retain filenames without writing a custom uploader. 
i.e. a temp file has the URL: `/d2l/lp/files/temp/GUID.jpg;image.jpg/View`, and TinyMCE cannot extract the correct filename from this format as it relies on the URL. 

The files save when a 'blur' event is registered, i.e. the Editor is unfocused. Auto uploads are no longer turned on.

This code also touches the `blobCache` of TinyMCE, meaning it can be quite fragile to version updates. The property is not meant to be modified by any actor other than TinyMCE. 
This code requires the LMS to verify functionality, making it not unit testable. UI Tests in the future could help smoke test this code. 

This feature is hidden behind [this flag](https://github.com/Brightspace/lms-launch-darkly-flags/blob/89e582cfe54dad62b83316b980b0639b795bacf2/features/exterminator/f20696-htmleditor-image-enhancements.feature.json#L3).  
Related LMS PR: https://github.com/Brightspace/lms/pull/37291


htmleditor-mixin.js
```diff
@@ -666,6 +666,7 @@ export const HtmlEditorMixin = superclass => class extends Localizer(LabelledMix
 		if (this._fraContext) editImageConfig.editimage_fetch_image = img => getImage(this, img.src);
 
 		tinymce.init({
+			automatic_uploads: false,
```
grant-cleary, line 669: Nit: These should be in alpha order, so this property should come just before `base_url`.

htmleditor-mixin.js
```diff
@@ -739,8 +740,44 @@ export const HtmlEditorMixin = superclass => class extends Localizer(LabelledMix
...
+								name: () => scannedBlobInfo.name(),
+								filename: () => fileName,
+							};
+							blobCache.removeByUri(uploadReadyBlobInfo.blobUri());
+							blobCache.add(uploadReadyBlobInfo);
+
+						}
+					});
+
+					//replace the files and take the images off the stack?
```
grant-cleary, line 778: I don't think this comment is strictly necessary here.

htmleditor-mixin.js
```diff
@@ -739,8 +740,44 @@ export const HtmlEditorMixin = superclass => class extends Localizer(LabelledMix
 					editor.undoManager.clear();
 				});
 
-				const persistEditorContents = () => {
-					if (this.pasteLocalImages) editor.uploadImages();
+				//might need to be propogated out since it's not always used (generalization)
+				const persistEditorContents = async() => {
```
grant-cleary, line 744: Now that this is async, remember that we need to `await` it in the two places it's used. That's actually probably where the race condition is coming from with the validation logic.

htmleditor-mixin.js
```diff
@@ -739,8 +740,44 @@ export const HtmlEditorMixin = superclass => class extends Localizer(LabelledMix
...
+								filename: () => fileName,
+							};
+							blobCache.removeByUri(uploadReadyBlobInfo.blobUri());
+							blobCache.add(uploadReadyBlobInfo);
+
+						}
+					});
+
+					//replace the files and take the images off the stack?
+					await editor.uploadImages();
```
grant-cleary, line 779: The `if (this.pasteLocalImages)` check is still necessary here.

htmleditor-mixin.js
grant-cleary: I would also set this to `automatic_uploads: !this.pasteLocalImages` just to reduce the blast radius here. We only play around with these settings when it's relevant to us.

PaulKrznaric: So I should move `automatic_uploads` before `base_url` in order to follow alpha ordering?

htmleditor-mixin.js
```diff
@@ -739,8 +740,44 @@ export const HtmlEditorMixin = superclass => class extends Localizer(LabelledMix
...
-				const persistEditorContents = () => {
-					if (this.pasteLocalImages) editor.uploadImages();
+				//might need to be propogated out since it's not always used (generalization)
+				const persistEditorContents = async() => {
+					const scannedImages = await editor.editorUpload.scanForImages();
+					const blobCache = editor.editorUpload.blobCache;
+
+					scannedImages.forEach(scannedImage => {
+						const scannedBlobInfo = scannedImage.blobInfo;
+						if (scannedBlobInfo.filename().startsWith('editimage')) {
```
grant-cleary, line 750: Nit: You can invert this logic with
```
if (!scannedBlobInfo.filename().startsWith('editimage')) return;
```
Which gets rid of some nesting and makes this more readable.

htmleditor-mixin.js
PaulKrznaric: Ah I understand now, 1 before u

grant-cleary: That's right.

htmleditor-mixin.js
```diff
@@ -739,17 +740,51 @@ export const HtmlEditorMixin = superclass => class extends Localizer(LabelledMix
...
+						if (!scannedBlobInfo.filename().startsWith('editimage')) return;
+
+						const uri = scannedBlobInfo.uri();
+						const regexMatchFileSystem = uri.match(/\/([^/]+\.(jpg|jpeg|png|gif|bmp))$/i);
+						const tempFilePostfix = '/View';
+						let fileName = scannedBlobInfo.filename();
+
+						if (regexMatchFileSystem) {
+							fileName = regexMatchFileSystem[1];
+						} else if (uri.endsWith(tempFilePostfix) && uri.includes('/temp/')) {
```
grant-cleary, line 759: This could also be `/immutabletemp/`.

htmleditor-mixin.js
```diff
@@ -739,17 +740,51 @@ export const HtmlEditorMixin = superclass => class extends Localizer(LabelledMix
...
+
+					if (this.pasteLocalImages) await editor.uploadImages();
+
 					this._html = editor.getContent();
 				};
 
 				editor.on('detach', () => {
-					persistEditorContents();
+					persistEditorContents().Result;
 				});
```
grant-cleary, line 784: I'm not sure what this is doing, but this isn't the appropriate way to await a promise. Instead:
```
editor.on('detach', async() => {
   await persistEditorContents();
});
```

htmleditor-mixin.js
```diff
@@ -739,17 +740,51 @@ export const HtmlEditorMixin = superclass => class extends Localizer(LabelledMix
...
 
 				editor.on('focusout', () => {
-					persistEditorContents();
+					persistEditorContents().Result;
 					this._handleFocusOut();
 				});
```
grant-cleary, line 789: Same as above:
```
editor.on('focusout', async() => {
   await persistEditorContents();
   this._handleFocusOut();
});
```

htmleditor-mixin.js
```diff
@@ -756,7 +756,7 @@ export const HtmlEditorMixin = superclass => class extends Localizer(LabelledMix
 
 						if (regexMatchFileSystem) {
 							fileName = regexMatchFileSystem[1];
-						} else if (uri.endsWith(tempFilePostfix) && uri.includes('/temp/')) {
+						} else if (uri.endsWith(tempFilePostfix) && (uri.includes('/temp/') || uri.includes('/immutabletemp'))) {
```
grant-cleary, line 759: Ah, sorry, I think I led you astray here. I don't think immutable temp files have their own URI stem, so that additional check is in fact pointless.

htmleditor-mixin.js
```diff
@@ -671,6 +671,7 @@ export const HtmlEditorMixin = superclass => class extends Localizer(LabelledMix
 			a11ychecker_issue_url_callback: ruleId => a11yCheckerRuleUrlMap[ruleId],
 			allow_html_in_named_anchor: true,
 			allow_script_urls: true,
+			automatic_uploads: !this.pasteLocalImages,
```
grant-cleary, line 674:
```suggestion
			automatic_uploads: !this.pasteLocalImages && (!this._context || !this._context.useNewTinyMCEVersion),
```

htmleditor-mixin.js
```diff
@@ -739,17 +740,53 @@ export const HtmlEditorMixin = superclass => class extends Localizer(LabelledMix
 					editor.undoManager.clear();
 				});
 
-				const persistEditorContents = () => {
-					if (this.pasteLocalImages) editor.uploadImages();
+				const persistEditorContents = async() => {
+
+					if (this._context?.useNewTinyMCEVersion) {
```
grant-cleary, line 745:
```suggestion
					if (this._context && this._context.useNewTinyMCEVersion) {
```
Unfortunately null chaining isn't supported in all of our maintenance browsers yet.

grant-cleary: Though actually, this does technically bleed out of the feature flag, since this call wasn't async before. Even if there's nothing being awaited in an async method, the browser still schedules its run-time differently than if it wasn't async. So I think the flag check should control two versions of this method and the handlers below; flag off === synchronous and the original code, and flag on === async and the new code.

PaulKrznaric: I made some changes to the way things are flagged. I'm unsure if this is the best way to go about flagging since the methods can be synchronous vs asynchronous, but it made sense to me.

grant-cleary (approved):

htmleditor-mixin.js
```diff
@@ -739,19 +740,76 @@ export const HtmlEditorMixin = superclass => class extends Localizer(LabelledMix
...
+				};
+
+				if (this._context && this._context.useNewTinyMCEVersion) {
+					persistEditorContents = async() => {
+						const scannedImages = await editor.editorUpload.scanForImages();
+						const blobCache = editor.editorUpload.blobCache;
+
+						scannedImages.forEach(scannedImage => {
+							const scannedBlobInfo = scannedImage.blobInfo;
+							if (scannedBlobInfo.filename().startsWith('editimage') === false) return;
```
grant-cleary, line 766:
```suggestion
							if (!scannedBlobInfo.filename().startsWith('editimage')) return;
```
No need for strict equality here.

components/plugins/image.js
```diff
@@ -67,7 +67,7 @@ export function uploadImage(editor, blobInfo) {
 			));
 		}
 
-		const fileName = blobInfo.filename().replace('blobid', 'pic');
+		const fileName = decodeURIComponent(blobInfo.filename().replace('blobid', 'pic'));
```
PaulKrznaric, line 70: Due to the LMS's encoding, we need to de-encode the URL in order for us to be able to properly interact with the files.

grant-cleary (approved):

PR #1501, opened 2023-06-13
DE53130 - Terminate uploadImages early if editor not initialized

PaulKrznaric:
When quickly switching between users in the New Content Experience, it was noticed that the HTML Editor was throwing an error for `uploadImages()` due to the editor not being fully initialized. 

This change adds a check to ensure that the editor is initialized before any other work is done to prevent an error from occurring and unnecessary code from executing.


grant-cleary (approved):

htmleditor-mixin.js
```diff
@@ -758,6 +758,9 @@ export const HtmlEditorMixin = superclass => class extends Localizer(LabelledMix
 
 				if (this._context && this._context.useNewTinyMCEVersion) {
 					persistEditorContents = async() => {
+						if (this._initState !== editorInitStates.INITIALIZED) {
```
grant-cleary, line 761: Technically there's no real cost to also applying this to the flag off case, but I'm also not super concerned about that. This wasn't a particularly high priority issue in the first place; just generating some log noise under very specific circumstances.

ca2hill (approved):

PR #1505, opened 2023-06-14
Adding LMS language detection to mirror settings in IE preview

jschlumkoski:


grant-cleary:
I'm curious why this is necessary. The localizer should have access to this language code everywhere, so what is it being used by such that we need to do something custom?

jschlumkoski:
> I'm curious why this is necessary. The localizer should have access to this language code everywhere, so what is it being used by such that we need to do something custom?

@grant-cleary The goal is to have Insert Element respect the directionality and language that's set on the content topic they're included on. Right now if you were editing in a RTL language (e.g., Arabic) and you were to preview it, the preview would show as LTR because the document doesn't have a direction set. The preview iframe just uses `srcdoc` to render the component markup and some supplementary scripts, so that's why we're passing them in and adding them.

The lang piece isn't entirely necessary right now, but would become relevant as we flip our elements to components, and they need to fetch lang terms for default verbiage for a specific language.

jschlumkoski:
@grant-cleary If it's preferable to just have the injected script grab the attribute values from the top level document so they're not redundantly tied into the component, I could do that instead?

grant-cleary:
Ooooh, I see. No, that's alright, I think this solution is fine. I forgot that you have that preview functionality and need to be able to pass this information into an iframe. I think fetching the lang from the component makes sense; I'm always a little nervous when iframes start reaching out to the parent document. 😆

grant-cleary (approved):

PR #1514, opened 2023-06-21
Convert all ref URLs to absolute before trying to compare them

grant-cleary:
It turns out generalizing the code for adding refs actually introduced a regression here. For reasons unknown, if I have a ref relative to the current domain, `stylesheet.href` returns the absolute href for that stylesheet, whereas `stylesheet.getAttribute('href')` returns the domain-relative href. So, yeah, the solution here is to just make sure we're converting both sides of the comparison to absolute URLs before trying to compare.


components/plugins/fullpage.js
```diff
@@ -220,10 +220,12 @@ class FullPageState {
 		});
 
 		const getExistingRef = (doc, url) => {
+			const baseUrl = this._editor.options.get('document_base_url');
+			const absoluteUrl = new URL(url, baseUrl).href;
```
grant-cleary, line 224: This didn't need to be in the inner loop. I'm sure the cost of doing this multiple times is trivial, but this is still better.

components/plugins/fullpage.js
```diff
@@ -220,10 +220,12 @@ class FullPageState {
...
 			const existingRefs = doc.head.querySelectorAll(type);
 			for (const ref of existingRefs) {
-				const absoluteUrl = new URL(url, this._editor.options.get('document_base_url')).href;
-				if (ref.getAttribute(urlAttr) === absoluteUrl) {
+				const absoluteRefUrl = new URL(ref.getAttribute(urlAttr), baseUrl).href;
```
grant-cleary, line 227: If `ref` is already an absolute link, then the URL constructor will ignore the baseUrl we're passing in.

test/htmleditor.test.js
```diff
@@ -281,6 +281,21 @@ describe('d2l-htmleditor', () => {
...
+				el.html = `<html><head>${testStylesheet}</head><body></body></html>`;
+				await el.updateComplete;
+
+				// Get and set html to force fullpage GetContent/SetContent handlers to run
+				const newHtml = el.html;
+				el.html = newHtml;
+				await el.updateComplete;
+
+				expect(el.html.match(/href="whatever.css"/g) || []).to.have.length(1);
+			});
```
grant-cleary, line 298: This test case should prevent future regressions here, as long as we don't change the events used by the fullpage plugin (the chance of that is vanishingly small).

JakeProuse (approved):

PR #1516, opened 2023-06-21
Switch to new testing framework

grant-cleary:
Switch over to use the centralized `@brightspace-ui/testing` test framework. This required a little bit of refactoring but will make life a lot easier moving forward.


test/htmleditor.test.js
```diff
@@ -24,17 +24,11 @@ describe('d2l-htmleditor', () => {
...
-		await component.initializationComplete;
 		return component;
 	};
 
 	runTestsWithTinyMCEVersioning(() => {
 		describe('accessibility', () => {
-			it('should fail axe tests with missing label', async() => {
-				const el = await createEditorComponent({}, false);
-				await expect(el).not.to.be.accessible();
-			});
```
grant-cleary, line 36: This test was broken and also useless, since the editor now throws an error and won't render if its label is missing.

test/htmleditor.test.js
```diff
@@ -49,7 +43,7 @@ describe('d2l-htmleditor', () => {
 			it('should pass all axe tests with labelled-by', async() => {
 				const el = await createEditorComponent({
 					labelledBy: 'myLabel'
-				}, true);
+				}, false);
```
grant-cleary, line 46: This test was also broken, as this boolean was adding an implicit label.

test/htmleditor.test.js
```diff
@@ -24,17 +24,11 @@ describe('d2l-htmleditor', () => {
 			${attrs.labelledBy ? html`<span id="${attrs.labelledBy}">This is a label!</span>` : ''}
 		`);
 
-		await component.initializationComplete;
```
grant-cleary, line 27: There's a weird timing issue here where with this present, the first test in the fixture always fails.

grant-cleary:
Hmm, I see that a test fails in CI that doesn't fail locally. I do love that.

grant-cleary:
The editor's testing framework has changed so much since this PR was first put up that it's easier to just close this and start fresh.

PR #1524, opened 2023-06-27
Added toast messages for practice add/delete/move actions

rileyodonnelld2l:
This ensures there is feedback for when a practice answer/category/step is added/deleted/moved. It also ensures the focus moves along with an item when it is moved.


components/practice/multiple-choice.js
```diff
@@ -455,6 +471,15 @@ class MultipleChoicePractice extends RequesterMixin(Localizer(LitElement)) {
...
+		}
+		else if (e.target.getAttribute('data-target') === 'delete') {
+			this._deletedAnswer = false;
+		}
+		else if (e.target.getAttribute('data-target') === 'move') {
+			this._movedAnswer = false;
+		}
 		else {
 			this._noAnswers = false;
 		}
```
grant-cleary, line 485: Nit: At this point I would strongly consider a switch statement here instead.

grant-cleary (approved):

PR #1572, opened 2023-07-26
Introduce new framework for implementing HTML editor plugins

grant-cleary:
The plugin framework originally built for the HTML editor assumed that any/all plugins would be sourced from within the editor, as that matched existing behaviour. However, we now have a need to source plugins from other repos, which necessitates refactoring the plugin framework to support this.

The main goals of this refactor are as follows:
1. Provide predictable APIs for internal consumers to create and register plugins, as well as any UI elements they may need (dialogs, context menus, overlays, toolbar buttons).
2. Restrict access to the editor component and TinyMCE's internal APIs, to ensure we can update TinyMCE or the component's internals in the future without needing to scour all of our repos for necessary changes.
3. In general, standardize how plugins interact with the editor to make maintenance simpler.

This particular PR builds out the core framework and migrates the existing internal plugins. This is an already massive change, but deliberately omits the Maverick-managed plugins and toolbar rendering, as there are complexities in these two cases that make them far more complex to jam into this PR. I will also be adding a series of tests covering the new API surfaces in a future PR.


components/selection.js
```diff
@@ -0,0 +1,227 @@
...
+		return this.$range.getBoundingClientRect();
+	}
+
+	toString() {
+		return this.$range.toString();
+	}
+
+}
+
+export class EditorSelection {
```
grant-cleary, line 193: The intent here is to prevent a plugin from grabbing the current node from TinyMCE's selection object and using it to walk the DOM up to the document. The general idea here is that plugins should only have the ability to manipulate the editor iframe's document body, not the document or the head.

htmleditor-mixin.js
```diff
@@ -631,6 +668,18 @@ export const HtmlEditorMixin = superclass => class extends Localizer(LabelledMix
...
 		if (this._editorDisconnected) return;
 
+		const useNewPluginFramework = this._context && this._context.useNewPluginFramework;
+
+		let pluginList = [];
+		if (useNewPluginFramework) {
+			const pluginRegistry = new PluginRegistry(this, this._context);
+			const newPlugins = await pluginRegistry.loadPlugins();
+
+			pluginList = newPlugins.concat(getPlugins(this, true));
```
grant-cleary, line 678: On the face of it, it may seem weird that this can be done before TinyMCE initializes. However, TinyMCE's plugin registration is global, so as long as TinyMCE has been loaded, it doesn't actually matter if an editor has initialized. Plugins can safely be registered prior to initialization.

htmleditor-mixin.js
```diff
@@ -615,6 +632,26 @@ export const HtmlEditorMixin = superclass => class extends Localizer(LabelledMix
 		editorContainer.style.height = `${newHeight}px`;
 	}
 
+	_initializeFullPageState(editor) {
```
grant-cleary, line 635: After careful thought, I now believe that `full-page` and `context-menu` should not be plugins. They're core to the editor experience and now that we have our nice centralized wrappers and registries, we have no need to treat them like plugins any longer.

htmleditor-mixin.js
```diff
@@ -663,7 +712,7 @@ export const HtmlEditorMixin = superclass => class extends Localizer(LabelledMix
 		stylesConfig.body_class = bodyClass;
 
 		const editImageConfig = {};
-		if (this._fraContext) editImageConfig.editimage_fetch_image = img => getImage(this, img.src);
+		if (this._fraContext) editImageConfig.editimage_fetch_image = useNewPluginFramework ? img => getImage(this, img.src, this._context) : img => getImage(this, img.src);
```
grant-cleary, line 715: Because the new plugin framework passes the context object around manually, we can eventually stop propagating it via `provideInstance`. In preparation for that, I have tried to short-circuit other places where we fetch it via requestors. Note that I haven't really made any effort to be thorough here; these were just a couple pieces of low-hanging fruit. We can't remove the context object from the provider until the flag for this change is removed regardless.

htmleditor.js
```diff
@@ -472,6 +458,10 @@ class HtmlEditor extends HtmlEditorMixin(SkeletonMixin(LitElement)) {
 				this._editing = false;
 			}
 		}
+
+		if (e.detail && !e.detail.skipFocus && this._context && this._context.useNewTinyMCEVersion) {
+			requestAnimationFrame(this.focus.bind(this));
```
grant-cleary, line 463: This was being done on a component-by-component basis. Well, only `insert-code` uses it, so it wasn't _really_ boilerplate. Still, it makes more sense to handle this centrally instead of relying on the plugins to handle it.

components/event-registry.js
```diff
@@ -0,0 +1,74 @@
...
+	}
+
+	registerBeforeSetContentEvent(handler) {
+		this.$editor.on('BeforeSetContent', e => {
+			const content = e.content;
+
+			const result = handler(content, {
+				format: e.format,
+				selection: e.selection,
+				sourveView: e.source_view
```
grant-cleary, line 23: Once again, we don't actually want to expose the event target to plugins because they could then just walk the DOM. In practice, no plugins require access to the event target.

components/event-registry.js
```diff
@@ -0,0 +1,74 @@
...
+
+	registerPostProcessEvent(handler) {
+		this.$editor.on('PostProcess', () => handler());
+	}
+
+	registerSetContentEvent(handler) {
+		this.$editor.on('SetContent', () => handler());
+	}
+
+	registerShortcutKeyEvent(handler) {
```
grant-cleary, line 59: A number of plugins set keyboard shortcuts so it made sense to me to just centralize this into one common helper method.

components/plugin-registry.js
```diff
@@ -0,0 +1,213 @@
...
+}
+
+function addTinyMCEPlugin(plugin, pluginName) {
+	tinymce.PluginManager.add(pluginName, function(editor) {
+		if (typeof plugin.registerPlugin !== 'function') {
+			throw new Error(`d2l-htmleditor: Configuration error in plugin with name ${pluginName}. Expect registerPlugin function.`);
+		}
+
+		const context = contextMap.get(editor.id);
+		const registries = buildRegistries(editor, privilegedDefaultPlugins.includes(pluginName));
```
grant-cleary, line 175: This is a bit tricky. We need to rely on global maps here because TinyMCE's plugin registration is global. This was never an issue before because any context being passed into the plugins was done via the `editor` object that TinyMCE supplies. We want to pass in our registries instead, however, so we need to build them dynamically. The editor-specific mappings exist so that we don't need to construct the registries from scratch for every plugin.

components/command-registry.js
```diff
@@ -0,0 +1,152 @@
...
+			const dialogPromise = dialogFactory(dialog);
+			dialog.opened = true;
+			this.$component._dialogOpen = true;
+
+			await dialogPromise;
+
+			requestAnimationFrame(() => {
+				dialog.opened = false;
+				this.$component._dialogOpen = false;
+				dialog.remove();
```
grant-cleary, line 116: Destroying the dialog when it closes incurs a bit more cost each time it's reopened, but also removes the need for plugins to handle resetting their own state manually. I think it's worth the trade-off.

components/command-registry.js
```diff
@@ -0,0 +1,152 @@
...
+
+		this.$editor.addCommand(command, async() => {
+			const overlay = this.$component.shadowRoot.querySelector('.d2l-htmleditor-overlay-container').appendChild(
+				document.createElement(`d2l-htmleditor-${overlayName}-overlay`)
+			);
+
+			overlay.allowOverflow = this.$component.type !== 'full' && !this.$component._fullscreen && !this.$component._percentageBasedHeight();
+			overlay.floating = this.$component.type !== 'full' && !this.$component._fullscreen;
+			overlay.maxHeight = this.$component.maxHeight;
+			overlay.noResize = this.$component._fullscreen;
```
grant-cleary, line 135: I'm not as keen on this. The overlay component requires all of these to be set, so plugins creating overlays need to explicitly pass them through. I want to rethink whether this could be better handled with a mixin or something similar, but this works, only one plugin uses it right now, and realistically I don't think it's particularly high priority.

components/api-wrapper.js
```diff
@@ -0,0 +1,262 @@
...
+	setContent(content, isSelection, noEvents) {
+		if (isSelection) {
+			if (!this.$editor.selection) return;
+			else this.$editor.selection.setContent(content, { no_events: noEvents });
+		} else {
+			this.$editor.setContent(content, { no_events: noEvents });
+		}
+	}
+
+	/* VERIFY THIS */
```
PaulKrznaric, line 192: Is this verified?

components/command-registry.js
```diff
@@ -0,0 +1,152 @@
...
+	alignLeft: 'justifyLeft',
+	alignRight: 'justifyRight',
+	bold: 'bold',
+	character: 'mceShowCharmap',
+	emoticons: 'mceEmoticons',
+	fontFamily: 'fontname',
+	foreColor: 'forecolor',
+	formatBlock: 'formatBlock',
+	formatPainter: 'mceToggleFormatPainter',
+	fontSize: 'fontsize',
```
PaulKrznaric, line 13: nit: Inconsistent camel casing

components/context-menu/context-menu.js
```diff
@@ -31,119 +32,132 @@ export function addContextMenuPlugin() {
 		const context = requestInstance(editor.getElement(), 'context');
 		const component = context.component;
 
-		const getSelectionInfo = () => {
+		initializeContextMenus(component, editor);
```
PaulKrznaric, line 35: By default are we not using the new framework?

components/command-registry.js
grant-cleary: These are internal TinyMCE commands; we don't get a say over their casing, unfortunately.

components/api-wrapper.js
grant-cleary: No. This is only used by Maverick's plugins, which haven't been migrated yet, so I'm still uncertain if it's going to be needed (they're doing a bit of a refactor themselves). I'm probably going to move this into the new selection object regardless.

components/context-menu/context-menu.js
grant-cleary: From the perspective of the component, no. If `context.useNewPluginFramework` is false or doesn't exist, the editor will keep existing behaviour. That particular object is going to be controlled by a feature flag in the LMS that doesn't yet exist.

components/selection.js
JakeProuse: This looks really good looking at it from a consumption aspect, will the existing AddStylehseet/Script API handlers still exist for us to add stuff to the head?

htmleditor.js
JakeProuse: This is a nice touch for us in the future if we ever do something that uses inline editing outside of our modals!

components/selection.js
grant-cleary: Yup! In the new framework I've surfaced those here as part of the API wrapper: https://github.com/BrightspaceUI/htmleditor/pull/1572/files#diff-37976b23ccee2d9669eace413c17b35006d3b5a05b14d16eb17d77066f3a5f18R23

Because this doesn't account for content-styler, layouts, practices, or insert-element yet, there may be APIs you're reliant on that I've missed. I tried to be reasonably thorough and plan ahead to see what will be needed to move those over, but if I missed some I can rectify that later.

components/plugins/a11ychecker.js
```diff
@@ -2,6 +2,25 @@ import { cmds } from '../commands.js';
...
+		return name;
+	}
+
+	isAvailable(inlineLimited) {
+		return !inlineLimited;
+	}
+
+	registerPlugin(context, helpers) {
+		helpers.commandRegistry.registerCommand(
+			cmds.a11yChecker,
```
grant-cleary, line 17: My feeling is that this centralized exported list of commands can and should eventually go away, but we're not there yet. If I can get to a point where plugins can define everything that needs to use their own commands, then the actual command names can be entirely self-contained in each plugin, with the framework only caring if it encounters duplicates. I'm unaware of a use case where a plugin needs to call another plugin's commands.

components/event-registry.js
```diff
@@ -0,0 +1,74 @@
...
+		this.$editor.on('BeforeSetContent', e => {
+			const content = e.content;
+
+			const result = handler(content, {
+				format: e.format,
+				selection: e.selection,
+				sourveView: e.source_view
+			});
+
+			if (result) e.content = result;
```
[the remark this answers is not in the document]
JakeProuse: should this be `sourceView: e.source_view` intead of `sourveView`?

components/event-registry.js
grant-cleary: ... Oops. Yes indeed. 😆

components/event-registry.js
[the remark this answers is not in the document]
grant-cleary: Also yes. 😆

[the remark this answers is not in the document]
grant-cleary: I'm trying to make these APIs more wacky, okay?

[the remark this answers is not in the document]
JakeProuse: What's an API without some easter eggs 🤣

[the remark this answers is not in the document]
grant-cleary: Exactly! Thanks for catching these. 😆

components/api-wrapper.js
```diff
@@ -0,0 +1,257 @@
...
+		});
+	}
+
+	toggleA11yAudit() {
+		if (this.#editor.plugins && this.#editor.plugins.a11ychecker) this.#editor.plugins.a11ychecker.toggleaudit();
+	}
+
+}
+
+export class PrivilegedInternalAPIWrapper extends InternalAPIWrapper {
```
dbatiste, line 216: Do you think that all privileged uses will be internal to this repo? If so, we can consider using `exports` in `package.json` to prevent the ability of others from importing this class outside of this repo. Or, would that be too limiting?

components/command-registry.js
```diff
@@ -0,0 +1,152 @@
...
+		this.#editor = editor;
+
+		this.#commands = new Set();
+	}
+
+	get builtInCommands() {
+		return builtInCommands;
+	}
+
+	$addCommandReference(command) {
```
dbatiste, line 69: Guessing you meant this to be `#...` ?

components/command-registry.js
dbatiste: Agreed.

components/event-registry.js
```diff
@@ -0,0 +1,76 @@
...
+		});
+	}
+
+	registerInputEvent(handler) {
+		this.#editor.on(inputEventTypes.join(' '), () => handler());
+	}
+
+	registerPostProcessEvent(handler) {
+		this.#editor.on('PostProcess', () => handler());
```
dbatiste, line 54: Is this wrapping the handler call so that `handler` isn't passed tinyMCE's event args?

grant-cleary: That's right, yeah.

components/api-wrapper.js
grant-cleary: Hmm, that's a good point. I think there's limited use someone could get out of using this, on the grounds that they would need to instantiate it with a TinyMCE instance (which if they have one, they can just do all of these things anyways). That said, I think restricting what can be imported to avoid confusion might still be good.

dbatiste: That's definitely something we're going to start leveraging more so that we can limit the use of internal components.

components/event-registry.js
```diff
@@ -0,0 +1,76 @@
...
+				selection: e.selection,
+				sourceView: e.source_view
+			});
+
+			if (result) e.content = result;
+		});
+	}
+
+	registerInputEvent(handler) {
+		this.#editor.on(inputEventTypes.join(' '), () => handler());
```
dbatiste, line 50: Nit: looks like this is the only place `inputEventTypes` is used - could write those out inline here to make the code a little easier to follow.

components/api-wrapper.js
grant-cleary: Yeah, it feels like a no-brainer to do that here. For the most part I think very little within the editor actually needs to be exported, so I'll make it a point to investigate what we can lock down. I think I'll likely do that in a later PR though.

components/event-registry.js
grant-cleary: Good point, yeah.

components/api-wrapper.js
grant-cleary: Sorry, I also realize I didn't answer one of your original questions. No, I think it's safe to say that if a plugin needs privileged access, it needs to live in the editor repo. If there's a reason that can't happen (circular dependencies), then we should consider a tightly controlled second integration repo of sorts where these can be controlled.

components/api-wrapper.js
```diff
@@ -0,0 +1,264 @@
...
+		return this.#component.noFilter;
+	}
+
+	getSelection() {
+		if (!this.#editor.selection) return null;
+		return new EditorSelection(this.#editor.selection);
+	}
+
+	insertContent(content, skipFocus) {
+		if (skipFocus) this.#editor.execCommand('mceInsertContent', false, content, { skip_focus: skipFocus });
```
grant-cleary, line 126: See https://github.com/BrightspaceUI/htmleditor/pull/1586. It turns out the `insertContent` API doesn't allow focus to be bypassed in this way, so we need to execute the command directly.

components/plugins/insert-code.js
```diff
@@ -71,6 +71,202 @@ export function convertCodeSamplesToHtml(htmlDoc) {
...
+		const formatCodeSnippet = async elem => {
+			setContentEditable(elem, false);
+			if (elem.tagName === 'PRE') {
+				const codeNode = elem.querySelector('code');
+				if (!codeNode) return;
+
+				setContentEditable(codeNode, false);
+				codeNode.dataset.originalCode = tinymce.DOM.encode(codeNode.textContent);
+			} else {
+				elem.dataset.originalCode = tinymce.DOM.encode(elem.textContent);
```
grant-cleary, line 94: Oops. It occurs to me that these should have been swapped to the helpers. Will fix.

components/plugins/insert-code.js
```diff
@@ -71,6 +71,202 @@ export function convertCodeSamplesToHtml(htmlDoc) {
...
+
+							if (e.detail.inline) {
+								setProperties(codeElem);
+								helpers.internalAPIWrapper.insertContent(codeElem.outerHTML, e.detail.skipFocus);
+							} else {
+								const preElem = document.createElement('pre');
+								preElem.appendChild(codeElem);
+								setProperties(codeElem, preElem);
+
+								helpers.internalAPIWrapper.insertContent(preElem.outerHTML, e.detail.skipFocus);
```
grant-cleary, line 230: See https://github.com/BrightspaceUI/htmleditor/pull/1586. I'm not going to fix the old code specifically here, but will merge main into this PR after the fix merges.

components/api-wrapper.js
JakeProuse: Yeah I'm thinking we'll still need this command for now 👍

components/context-menu/context-menu.js
```diff
@@ -31,119 +32,132 @@ export function addContextMenuPlugin() {
...
 			return {
-				range: editor.selection && editor.selection.getRng(),
-				node: editor.selection && editor.selection.getNode()
+				range: selection.getRng(),
+				node: selection.getNode()
 			};
+		}
+
+		return {
+			range: editor.selection && editor.selection.getRng(),
```
njostonehouse, line 53: Does Javascript have the `editor.selection?.getRng()` syntax?

components/context-menu/link/link-edit.js
```diff
@@ -45,9 +53,14 @@ export class LinkEditContextMenuPlugin {
 		else return node.getBoundingClientRect();
 	}
 
-	updateContentElement(contentElm, range, node) {
-		if (node.href) contentElm.linkLocation = node.href;
-		if (node.title) contentElm.linkTitle = node.title;
+	updateContentElement(contentElm, range, node, useNewPluginFramework) {
+		if (useNewPluginFramework) {
+			if (node.hasAttribute('href')) contentElm.linkLocation = node.getAttribute('href');
```
njostonehouse, line 58: nit: don't we prefer to be explicit about single-line conditional blocks, by using `{...}`?

components/plugin-registry.js
```diff
@@ -0,0 +1,216 @@
...
+
+		componentMap.set(component._editorId, component);
+		contextMap.set(component._editorId, context);
+	}
+
+	#comparePlugins(plugin, otherPlugin) {
+		const pluginHasDependencies = this.#hasDependencies(plugin);
+		const otherPluginHasDependencies = this.#hasDependencies(otherPlugin);
+
+		if (!pluginHasDependencies && !otherPluginHasDependencies) return 0;
```
njostonehouse, line 53: Is this line necessary?

components/context-menu/context-menu.js
grant-cleary: It does, but unfortunately not in some of the versions of Safari we still support... And unfortunately Babel doesn't transpile it in BSI, so we can't use it. 😢

components/context-menu/link/link-edit.js
```diff
@@ -21,13 +21,21 @@ export class LinkEditContextMenuPlugin {
 		return elementName.toUpperCase();
 	}
 
-	createContentElement(range, node) {
+	createContentElement(range, node, useNewPluginFramework) {
 		const contentElm = document.createElement(elementName);
 
-		const anchorNode = findComposedAncestor(node, node => node.nodeType === Node.ELEMENT_NODE && node.tagName === 'A');
+		const anchorNode = useNewPluginFramework
+			? node.findComposedAncestor(node => node.nodeType === Node.ELEMENT_NODE && node.tagName === 'A')
```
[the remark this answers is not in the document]
grant-cleary: I can change that to make it explicit.

components/context-menu/link/link-edit.js
grant-cleary: Not in our front-end JS, no. The single-line stuff is preferred unless it's going to lead to a ridiculously long line.

components/command-registry.js
```diff
@@ -0,0 +1,155 @@
...
+
+			const dialogPromise = dialogFactory(dialog);
+			dialog.opened = true;
+			this.#component._dialogOpen = true;
+
+			await dialogPromise;
+
+			requestAnimationFrame(() => {
+				dialog.opened = false;
+				this.#component._dialogOpen = false;
```
njostonehouse, line 118: Do we intend to remove `this.#component.focus()` part from some consumers of this?

components/plugins/color.js
```diff
@@ -43,6 +47,29 @@ const getComputedStyle = (node, propertyName) => {
...
+
+	registerPlugin(context, helpers) {
+		helpers.commandRegistry.registerCommand(cmds.foregroundColor, async() => {
+			const contextNode = helpers.internalAPIWrapper.getSelection().getNode();
+			if (!contextNode) return;
+
+			const backgroundColor = getComputedStyle(contextNode, 'background-color', true);
+
+			const color = await helpers.uiRegistry.openLmsDialog(`/d2l/lp/htmleditor/foregroundcolor?backgroundColor=${backgroundColor}`);
+			if (color) helpers.commandRegistry.executeInternalCommand(helpers.commandRegistry.builtInCommands.foreColor, color);
```
njostonehouse, line 68: Will is `''` falsy here? There used to be a check for `color !== ''`

components/plugins/equation.js
```diff
@@ -16,6 +16,160 @@ const editorTypes = {
...
+		const launchEditor = async editorType => {
+			const contextNode = getSelectedMathContainer();
+
+			let mathML = '';
+			if (contextNode && getEditorTypeForMathContainer(contextNode) === editorType) {
+				mathML = decodeURIComponent(contextNode.attributes.getNamedItem('data-d2l-mathml').value);
+			}
+
+			const html = await helpers.uiRegistry.openLmsDialogWithParam(
+				`/d2l/lp/math/createeditor/${context.orgUnitId}/`,
```
njostonehouse, line 70: Where did this URI come from? Why wasn't it specified before?

njostonehouse: Nevermind, found it
https://github.com/BrightspaceUI/htmleditor/blob/680bd88b95a0433841837b856f2b809879d12714/components/plugins/equation.js#L373

components/plugins/isf.js
```diff
@@ -25,6 +25,379 @@ export const isfStyles = css`
...
+				content = sb.ToString();
+			}
+			return content;
+		};
+
+		helpers.eventRegistry.registerBeforeSetContentEvent(content => {
+			return convertToImages(content);
+		});
+
+		helpers.eventRegistry.registerGetContentEvent(content => {
```
njostonehouse, line 394: Shouldn't this be `registerPostProcessEvent`?

njostonehouse (approved):

components/plugins/isf.js
grant-cleary: I'm not convinced `PostProcess` was the right event to use here. Specifically, TinyMCE doesn't provide any `content` param as part of that event, so I suspect this just... Never did anything? It's weird.

components/plugins/color.js
grant-cleary: Empty strings are falsy, yeah, so that check was redundant.

components/command-registry.js
grant-cleary: Plugins under the new framework can't access the component, so they definitionally also can't call this. So I think we're okay here and nothing needs to be removed.

components/plugins/equation.js
grant-cleary: Yeah, so for some plugins (Insert Equation, Insert Image, ISF), the LMS dialogs were thinly wrapped with a different custom component. This isn't necessary and doesn't buy us anything now that we can wrap all of that dialog-specific handling code in these registries anyways. So I opted to just drop the wrapper and invoke the LMS dialogs directly instead.

dbatiste: The idea with the thin wrappers is that we will eventually have these dialogs defined in lms-core as components. The internals will launch the LMS dialogs, but the idea is that eventually these things can be updated to use hypermedia (or whatever) and consumers would not need to know anything about the implementation, the end-point, the response, etc. We just didn't have a better home for them at the time, so they were implemented here with the idea that they might eventually be moved. The new file work that I am starting will define the QuickLinks dialog there.

grant-cleary: That's right. At the time I don't think the architectural direction on how to handle this sort of LMS middleware was particularly sorted, so we used a couple different patterns across the editor to isolate the LMS layer more. I think now that we have a better idea of how that work is going to play out, the wrappers will end up a bit redundant.

PR #1591, opened 2023-08-08
GenerativeAI Merge

JamesVaChao:
Leaving as a preview to see changes


components/toolbar/toolbar-full.js
```diff
@@ -372,6 +379,17 @@ class Toolbar extends ToolbarMixin(RequesterMixin(LitElement)) {
 					</d2l-menu>
 				</d2l-dropdown-menu>
 			</d2l-htmleditor-button-menu>
+
+			${showInsertPractice && this.context.generativeAIEndpoint ? html`
```
grant-cleary, line 383: The html-editor doesn't have room for new buttons outside the `+` menu. Adding this here will break the toolbar. Given that `Insert Practice` is inside the `+` menu, I would suggest also placing this button there.

components/context-menu/formatting/formatting.js
```diff
@@ -58,18 +58,17 @@ export class FormattingContextMenuPlugin {
...
 			<d2l-htmleditor-button-toggle cmd="${cmds.italic}" icon="italic" text="${this.localize('format:italic')}"></d2l-htmleditor-button-toggle>
 			<d2l-htmleditor-separator no-resize></d2l-htmleditor-separator>
 			<d2l-htmleditor-button-toggle cmd="${cmds.linkOpenMenu}" @click="${this._handleLinkAddEditClick}" icon="link" text="${this.localize('link:addEdit')}" no-execute no-resize></d2l-htmleditor-button-toggle>
 			<d2l-htmleditor-button-toggle cmd="${cmds.blockQuote}" icon="blockquote" text="${this.localize('format:blockquote')}" no-resize></d2l-htmleditor-button-toggle>
-			<d2l-htmleditor-separator no-resize></d2l-htmleditor-separator >
+			<d2l-htmleditor-separator no-resize></d2l-htmleditor-separator>
 			<d2l-htmleditor-button-toggle cmd="${cmds.formatBlock}" value="h1" icon="heading1" text="${this.localize('format:heading1')}" no-resize></d2l-htmleditor-button-toggle>
 			<d2l-htmleditor-button-toggle cmd="${cmds.formatBlock}" value="h2" icon="heading2" text="${this.localize('format:heading2')}" no-resize></d2l-htmleditor-button-toggle>
-		`);
+			`);
```
grant-cleary, line 71: Weird whitespace change here. There are a number of files in this PR that are unrelated to this PR but have changes made to them. Please consider undoing those changes before merge.

components/plugins/insert-code.js
```diff
@@ -343,6 +343,11 @@ class InsertCodeOverlay extends ThemeMixin(LitElement) {
 			</d2l-htmleditor-overlay>`;
 	}
 
+	async updated(changedProperties) {
+		super.updated(changedProperties);
+		if (!changedProperties.has('opened')) return;
+	}
```
grant-cleary, line 349: Unrelated change; please remove.

components/plugins/practice.js
```diff
@@ -47,8 +52,103 @@ const cdnDomain = 'https://practices.lcs.brightspace.com/latest';
...
+		text += `${cur.textContent  }\n`;
+		cur = cur.nextElementSibling;
+	}
+
+	return text;
+}
+
 export function addPracticePlugin() {
 	tinymce.PluginManager.add(name, function(editor) {
+		outerEditor = editor;
```
grant-cleary, line 150: This is dangerous if multiple editors are on the same page. The editor object shouldn't leave the scope of this function.

components/plugins/practice.js
```diff
@@ -820,14 +1330,23 @@ class PracticeDialog extends RequesterMixin(Localizer(LitElement)) {
...
+					if (LlmController.loading) {
+						LlmController.cancelStream(this.localize('practice:llm:cancel')); // might be in the middle of question generation..
+					}
+					LlmController.customized = false;
+					LlmController.customizedSourceText = '';
+					LlmController.clearFields(this._templateIndex, this.shadowRoot);
+					LlmController.generated = false;
+					LlmController.error = false;
+					LlmController.loading = false;
+					LlmController.meaningfulResult = true;
```
grant-cleary, line 1342: Why do all of these properties need to be static? Is there a reason these can't just be instance properties and the controller manages its own state? The reliance on globals here makes me nervous in the case where multiple editors could be loaded on the same page. While that's not a thing in Content, it's not safe to assume that will always be the case.

components/plugins/practice.js
```diff
@@ -820,14 +1330,23 @@ class PracticeDialog extends RequesterMixin(Localizer(LitElement)) {
...
 					this._handleClose();
 				}
 			}, { once: true });
 		} else {
 			this._handleClose();
 		}
 	}
-
```
grant-cleary, line 830: Some of these whitespace removals should be undone.

components/plugins/insert-code.js
JamesVaChao: @erickang21 did you have a justification for this, I know you added it for a feature

icons/robot.svg
```diff
@@ -0,0 +1,39 @@
+<?xml version="1.0" standalone="no"?>
```
grant-cleary, line 1: It would be good to run these SVGs through SVGOMG as per https://github.com/BrightspaceUI/core/tree/main/components/icons#svg-format. Ideally, we want to keep the SVG code as minimal as possible.

components/toolbar/toolbar-full.js
ianmclean2011: @grant-cleary yep, we'll get rid of this

icons/robot.svg
JamesVaChao: Removed, as not used. Noted for future svgs

components/plugins/practice.js
```diff
@@ -49,6 +53,7 @@ const name = 'd2l-practice-dialog';
 
 export function addPracticePlugin() {
 	tinymce.PluginManager.add(name, function(editor) {
+		LlmController.editor = editor;
```
grant-cleary, line 56: I realize we've already talked about this separately, but I just want to highlight why this is a real issue. `tinymce.PluginManager.add` runs once _per editor on the page_ for every loaded plugin. This means that if two editors on the page were to include Practices, this would run twice and whichever editor loads last will win. Some of the other static assignments are a little less worrisome because they only run when, e.g., the Practices dialog opens, which can't happen across two editors simultaneously. This one, however, will _always_ conflict if multiple Practices-enabled editors are present on the page.

components/plugins/practice.js
```diff
@@ -322,6 +328,15 @@ export function addPracticePlugin() {
...
 		});
 
+		editor.ui.registry.addButton('selection-llm', {
+			icon: 'chevron-down',
+			tooltip: 'Practice',
+			onAction: function() {
+				LlmController.selectedText = editor.selection.getContent({ format: 'text' });
+				launchPracticeEditor();
+			}
+		});
```
grant-cleary, line 338: With the button removed, is this used anywhere? Also, the use of `chevron-down` as an icon here seems very odd.

components/plugins/practice.js
```diff
@@ -603,6 +744,10 @@ class PracticeDialog extends RequesterMixin(Localizer(LitElement)) {
 		let content;
 		const template = templates[this._templateIndex];
 
+		LlmController.genAIEndpoint = this?._context?.generativeAIEndpoint;
```
grant-cleary, line 747: We support versions of Safari that are too old for optional chaining. You'll need to change this to something like `this && this._context && this.context.generativeAIEndpoint`.

components/practice/dropdown-blanks.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable sort-class-members/sort-class-members */
```
grant-cleary, line 1: This should be removed. If the code doesn't pass the linter, it should be fixed such that it does. Exceptions should be very rare.

components/practice/dropdown-blanks.js
```diff
@@ -434,14 +510,14 @@ class DropdownBlanksPractice extends RequesterMixin(Localizer(LitElement)) {
 										${answer.correctAnswers.indexOf(idx) > -1 ? '' : html`
 											<div class="d2l-blank-list-item" role="listitem">
 												<div class="d2l-form-field d2l-blank-text">
-													<d2l-input-text id="${getUniqueId()}" data-property="questions:0:answers:${index}:answers:${idx}:text" data-type="text" .value="${this._getPropertyValue(`questions:0:answers:${index}:answers:${idx}:text`)}" required label-hidden label="${this.localize('practice:incorrectAnswer')}" @change="${this._change}"></d2l-input-text>
+													<d2l-input-text ?disabled=${disabled} id="${getUniqueId()}" placeholder="Enter the question." data-property="questions:0:answers:${index}:answers:${idx}:text" data-type="text" .value="${this._getPropertyValue(`questions:0:answers:${index}:answers:${idx}:text`)}" required label-hidden label="${this.localize('practice:incorrectAnswer')}" @change="${this._change}"></d2l-input-text>
```
grant-cleary, line 513: The placeholder text here should be localized. We shouldn't have unlocalized English strings anywhere.

components/practice/dropdown-blanks.js
```diff
@@ -552,6 +628,132 @@ class DropdownBlanksPractice extends RequesterMixin(Localizer(LitElement)) {
...
+				this.questionText = contentFormatted;
+
+			} else if (fieldToFill === fields.answer) {
+				const answerIndex = this.activityData.questions[0].answers.map(function(answer) { return answer.id; }).indexOf(answerId);
+				if (answerIndex > -1) {
+					this.activityData.questions[0].answers[answerIndex].answers[0].text = contentFormatted;
+					this._updateValue(`questions:0:answers:${answerIndex}:answers:0:text`, contentFormatted);
+					this._updateActivity();
+				} else {
+					console.warn(`Unable to update answer. llm suggestion answer placeholder id of '${answerId}' not found`);
```
grant-cleary, line 684: This should be removed.

components/practice/fill-blanks.js
```diff
@@ -354,6 +428,99 @@ class FillBlanksPractice extends RequesterMixin(Localizer(LitElement)) {
...
+				this.questionText = contentFormatted;
+
+			} else if (fieldToFill === fields.answer) {
+				const answerIndex = this.activityData.questions[0].answers.map(function(answer) { return answer.id; }).indexOf(answerId);
+				if (answerIndex > -1) {
+					this.activityData.questions[0].answers[answerIndex].answers[0].text = contentFormatted;
+					this._updateValue(`questions:0:answers:${answerIndex}:answers:0:text`, contentFormatted);
+					this._updateActivity();
+				} else {
+					console.warn(`Unable to update answer. llm suggestion answer placeholder id of '${answerId}' not found`);
```
grant-cleary, line 490: Should be removed.

components/practice/fill-blanks.js
```diff
@@ -477,7 +652,7 @@ class FillBlanksPractice extends RequesterMixin(Localizer(LitElement)) {
 									${answer.answers.map((blank, idx) => html`
 										<div class="d2l-blank-list-item" role="listitem">
 											<div class="d2l-form-field d2l-blank-text">
-												<d2l-input-text id="${getUniqueId()}" data-property="questions:0:answers:${index}:answers:${idx}:text" data-type="text" .value="${this._getPropertyValue(`questions:0:answers:${index}:answers:${idx}:text`)}" required label-hidden label="${this.localize('practice:correctAnswer')}" @change="${this._change}"></d2l-input-text>
+												<d2l-input-text placeholder="Enter the question." id="${getUniqueId()}" data-property="questions:0:answers:${index}:answers:${idx}:text" data-type="text" .value="${this._getPropertyValue(`questions:0:answers:${index}:answers:${idx}:text`)}" required label-hidden label="${this.localize('practice:correctAnswer')}" @change="${this._change}"></d2l-input-text>
```
grant-cleary, line 655: Placeholder should be a localized term.

components/practice/llm-controller.js
```diff
@@ -0,0 +1,469 @@
...
+					<div class="d2l-error-text">
+						<div class="d2l-question-summary-icon">
+							<d2l-icon icon="tier2:unapproved"></d2l-icon>
+						</div>
+						<div class="d2l-question-summary-text">
+							<div class="d2l-question-summary-text1"><span>${html`${LlmController.errorText}`}</span></div>
+							<div class="d2l-question-summary-text1"><span>${html`${LlmController.errorSuggestion}`}</span></div>
+						</div>
+						<div class="d2l-question-summary-edit-button">
+							<d2l-button-subtle id="d2l-llm-generate" ?disabled=${LlmController.loading}  @click="${LlmController.getLlmSuggestionOnClick}" text="Try Again"></d2l-button-subtle>
```
grant-cleary, line 58: Unlocalized text.

components/practice/llm-controller.js
```diff
@@ -0,0 +1,469 @@
...
+			LlmController.clearFields(templateIndex, this.shadowRoot);
+			const context = requestInstance(this.editor.getElement(), 'context');
+			const component = context.component;
+			const dialog = component.shadowRoot.querySelector('d2l-practice-dialog');
+			dialog._updateActivity();
+
+			try {
+				await element.getLlmSuggestions();
+			} catch (e) {
+				console.log('Error with question generation', e);
```
grant-cleary, line 229: No console statements.

components/practice/llm-controller.js
```diff
@@ -0,0 +1,469 @@
...
+				LlmController.error = true;
+				element.errorText = errorLocalization || 'Error: Unable to generate question text';
```
grant-cleary, line 231: Unlocalized text.

components/practice/llm-controller.js
```diff
@@ -0,0 +1,469 @@
...
+		LlmController.meaningfulResult = true;
+		let isFirstChar = true;
+
+		const url = `${LlmController.genAIEndpoint}streamQuestions/${questionType}`;
+
+		const tokenJson = window.localStorage['D2L.Fetch.Tokens'];
+		if (!tokenJson) {
+			console.log('No D2L token. Try relogging in');
+			LlmController.errorText = 'Your access has expired.';
+			LlmController.errorSuggestion = html`Please <d2l-link href="${origin}/d2l/login?logout=1" target="_blank">sign in</d2l-link> again on a new tab and click Try Again.`;
```
grant-cleary, line 292: Unlocalized text.

components/practice/llm-controller.js
```diff
@@ -0,0 +1,469 @@
...
+	static sourceTextLimit = 8000;
+
+	//The genAIEnpoint is controlled by config/flag, so if it's there, the config
+
+	static templateIndex = -1;
+
+	// Cancels the question generation/streaming if in progress.
+	// cancelLocalize: Localized string for "Cancelling"
+	static cancelStream(cancelLocalize) {
+		LlmController.loadingMessage = cancelLocalize || 'Loading...';
```
grant-cleary, line 185: Unlocalized text.

components/practice/llm-controller.js
```diff
@@ -0,0 +1,469 @@
...
+			LlmController.errorText = 'Your access has expired.';
+			LlmController.errorSuggestion = html`Please <d2l-link href="${origin}/d2l/login?logout=1" target="_blank">sign in</d2l-link> again on a new tab and click Try Again.`;
+			LlmController.error = true;
+			return;
+		}
+		const { access_token: accessToken } = JSON.parse(tokenJson)['*:*:*'];
+		if (!accessToken) {
+			console.error('No access token. Try relogging in');
+			LlmController.errorText = 'Your access has expired.';
+			LlmController.errorSuggestion = html`Please <d2l-link href="${origin}/d2l/login?logout=1" target="_blank">sign in</d2l-link> again on a new tab and click Try Again.`;
```
grant-cleary, line 300: I'm going to stop repeating it, but console statements definitely need to go, and all text should be localized.

components/practice/true-false.js
```diff
@@ -257,7 +328,72 @@ class TrueFalsePractice extends RequesterMixin(Localizer(LitElement)) {
...
+			textToUse = this.editorText; // all text in editor
+		}
+		const questionType = 'trueFalse';
+
+		LlmController.loading = true;
+		await this.lc.getLlmStream(questionType, { content: LlmController.getDisplayText(textToUse) }, processContent);
+		LlmController.loading = false;
+	}
+
+	// eslint-disable-next-line sort-class-members/sort-class-members
```
grant-cleary, line 390: Should be removed and the code should be altered to pass linting.

components/practice/multiple-select.js
```diff
@@ -273,6 +368,105 @@ class MultipleSelectPractice extends RequesterMixin(Localizer(LitElement)) {
...
+					const matches = content.match(/\d/g);
+					if (!matches) {
+						return;
+					}
+					try {
+						const correctAnsArr = matches.map((n) => parseInt(n) - 1);
+						this.activityData.questions[0].correctAnswers = correctAnsArr;
+						this._updateValue('questions:0:correctAnswers', correctAnsArr);
+					} catch (e) {
+						console.error(e);
```
JamesVaChao, line 433: @grant-cleary Would you like me to remove all consoles or only the ones that are non failure cases?

ianmclean2011: Remove all console statements.

grant-cleary: All, please. We don't want anything writing to the console.

components/practice/fill-blanks.js
```diff
@@ -254,6 +258,69 @@ class FillBlanksPractice extends RequesterMixin(Localizer(LitElement)) {
 				border-spacing: 0;
 				padding: 0 5px;
 			}
+
```
JakeProuse, line 261: Could we bring the shared css out into a file that we can import into each component instead of having them copy and pasted?

components/practice/fill-blanks.js
```diff
@@ -300,12 +372,15 @@ class FillBlanksPractice extends RequesterMixin(Localizer(LitElement)) {
 		const template = templates[this._templateIndex];
 
 		return html`
+
```
JakeProuse, line 375: Nit: Can remove the extra spaces here

components/practice/fill-blanks.js
mdgbayly: I'm still surprised this is even needed. As far as I recall we don't render anything in the question components themselves. All the gen AI status messages are rendered in the dialog that renders the questions, so I don't see why we would need these styles.

components/plugins/practice.js
JamesVaChao: Changed, in follow up cleanup. Btw there are some usages you might wanna change as well: 
image

components/practice/fill-blanks.js
JamesVaChao: Agreed, is this fine to be done post merge @JakeProuse? Also @erickang21 how did you come to this conclusion https://github.com/BrightspaceUI/htmleditor/pull/1592/files/7cccea9d8d7c821cbffdc3a16119ad2346e1c681#r1294710158 What issue were you running into?

components/plugins/practice.js
grant-cleary: ^ FYI @jschlumkoski

components/practice/fill-blanks.js
JakeProuse: Yeah it's not a deal breaker if you're trying to get this merged, but should definitely get tackled at some point

lang/ar.js
```diff
@@ -391,6 +391,45 @@ export default {
 	"practice:fillBlanks:regularExpressionDescription": "للحصول على دليل شامل لإنشاء التعبيرات العادية، يُرجى زيارة",
 	"practice:fillBlanks:regularExpressionDescriptionLink": "تعليمات D2L Brightspace لإنشاء مكتبات الأسئلة، الأحرف المستخدمة في التعبيرات العادية.",
 	"practice:fillBlanks:regularExpressionInfo": "معلومات التعبير العادي",
+	// Content not meaningful banner
+	"practice:llm:contentNotMeaningfulBanner:error": "",
```
grant-cleary, line 395: Where these generated from `npm run langs:sync`? I haven't used that workflow in a little while (haven't had to add lang terms to the editor), but I recall that the workflow used to just fill in the English terms here. I'm not really concerned because we will get proper translations in before this releases, but I'm curious whether that workflow has changed or if you did this manually.

components/plugins/practice.js
```diff
@@ -1012,7 +1012,7 @@ class PracticeDialog extends RequesterMixin(Localizer(LitElement)) {
 				return html`<d2l-sequencing-practice id="d2l-sequencing-practice" @activityUpdate="${this._activityUpdated}" @activitySave="${this._saveActivity}" @togglePreview="${this._togglePreview}" @handleClose="${this._checkForChangesBeforeClose}" .activityData="${this.activityData}" ?display-feedback="${this._modifiers[0].value}" .editorText="${editorText}" .titleText="${titleText}" .summary="${this.summary}"></d2l-sequencing-practice>`;
 
 			case 3:
-				return html`<d2l-true-false-practice id="d2l-true-false-practice" @activityUpdate="${this._activityUpdated}" @activitySave="${this._saveActivity}" @togglePreview="${this._togglePreview}" @handleClose="${this._checkForChangesBeforeClose}" .activityData="${this.activityData}" ?display-feedback="${this._modifiers[0].value}" ?has-hint="${this._modifiers[1].value}" ?use-enumeration="${this._modifiers[2].value}" ?display-short-description="${this._modifiers[3].value}" .editorText="${editorText}" .titleText="${titleText}" .summary="${this.summary}"></d2l-true-false-practice>`;
```
JamesVaChao: This should've always been like this according to my merge. https://github.com/BrightspaceUI/htmleditor/pull/1611/files . I think you made a mistake while resolving branch conflicts. Good opportunity to learn https://github.com/BrightspaceUI/htmleditor/commit/1c42c9f1d85e745a63e3bce36750a990723d080d#r124899316

grant-cleary (approved):

PR #1592, opened 2023-08-08
Generative ai cleaning 1

JamesVaChao:
Cleanup PR for [US154701](https://rally1.rallydev.com/#/?detail=/userstory/706310404987&fdp=true): Switch HTML Editor to BSI Version

commands.js -> command not used
formatting.js -> changed spacing back to original, shouldn't show as change in final merge
practice.js -> 
- removed a bunch of helpers out of practice.js into llmcontroller
- removed global outerEditor,selectedText,seen to llmcontroller 
- reverted formatting
- removed mention of outerEditor
- a lot of functions needed sorting because eslint-disable-sorting was left in the file
- move passing of variables to each practice like ._gpt and .loadingMessage into llmController
- any weird spacing, was in the original repo, I added them back in to not show as a change

For each practice type in /practices -> Removed the userSelectedText and loadingMessageProperty as they were all duplicate code

llm-controller ->
Changed naming to _functionName to indicate they they're only used with llmController itself. 

valence-controller -> reverted back to original
toolbar-full -> added back spacing so doesn't show as change

For each language file -> Updated en.js to put all llm changes together. Put placeholder for all languages other than en.js


components/practice/dropdown-blanks.js
mdgbayly, line 1: We seem to have added a bunch of styes to each of these question components like `.d2l-question-summary`. From what I can tell they are not being used. They also exist in the the `plugins/practice.js`. Maybe they are left over from an earlier implementation approach?

JamesVaChao: Yes, I can look into styles that are not being used.

JamesVaChao: @erickang21 these are being used in llm-controller but they are defined in each question type? Did you just copy and paste all of them? Could they be in practice?

components/practice/multiple-choice.js
```diff
@@ -662,7 +659,7 @@ class MultipleChoicePractice extends RequesterMixin(Localizer(LitElement)) {
                                 </div>
                             ` : null}
 							</div>
-							<d2l-button-icon class="d2l-remove-answer" icon="tier1:delete" text="${this.localize('practice:questions:_deleteAnswerHandler')}" data-idx="${index}" @click="${this._deleteAnswerHandler}" ?disabled="${this._getPropertyValue(property.property).length <= this._minAnswersAmount || disabled}"></d2l-button-icon>
```
JamesVaChao, line 665: bug fixed

components/commands.js
```diff
@@ -23,7 +23,6 @@ export const cmds = {
 	llmSuggest: 'llm-suggest',
 	hr: 'insertHorizontalRule',
 	image: 'd2l-image',
-	generateImage: 'd2l-generate-image',
```
JamesVaChao, line 26: not used

erickang21 (commented):
Should we be removing the instances of static variables in this case?

JamesVaChao:
Yes eventually the goal is to move away from static variables as we can have more than one editor. Right now we have a lot of things as static, so I think the priority with this cleanup is consistency. We can have a further discussion with Grant and his team to determine if this is a must for this merge or if having a plan to transistion is enough.

components/practice/dropdown-blanks.js
mdgbayly: I thought they were already in practice as well?

JamesVaChao: Oh i meant to say "could they solely exist in practice, or did they have to be defined in each question type?"

components/plugins/practice.js
```diff
@@ -1222,27 +1012,28 @@ class PracticeDialog extends RequesterMixin(Localizer(LitElement)) {
...
+				return html`<d2l-true-false-practice id="d2l-true-false-practice" @updateValue="${this._updateValue}" @activityUpdate="${this._activityUpdated}" @activitySave="${this._saveActivity}" @togglePreview="${this._togglePreview}" @handleClose="${this._checkForChangesBeforeClose}" .activityData="${this.activityData}" ?display-feedback="${this._modifiers[0].value}" ?has-hint="${this._modifiers[1].value}" ?use-enumeration="${this._modifiers[2].value}" ?display-short-description="${this._modifiers[3].value}" .editorText="${editorText}" .titleText="${titleText}" .summary="${this.summary}"></d2l-true-false-practice>`;
```
mdgbayly, line 1021: I think this `@updateValue` event handler was added for the LLM code integration.
It doesn't exist on the other question types, and I don't see an `updateValue` event being dispatched anywhere.
Do we know what this was added for?

components/plugins/practice.js
```diff
@@ -1432,6 +1277,33 @@ class PracticeDialog extends RequesterMixin(Localizer(LitElement)) {
 			this._mobileView = true;
 		}
 	}
+	_processBlanks() {
```
mdgbayly, line 1280: Was trying to understand what is calling this `_processBlanks()` method and some of the related methods that it calls.
Might need to run the code to understand. I don't see them being called in this class, but there are similar methods in some of the other question type classes.
I'm wondering if this is a result of the way in which the question specific `getLlmSuggestions` is being called, possibly with the `this` reference pointing to the `PracticeDialog` instance rather than the question component instance, hence you copied some of these methods as a workaround. I vaguely recall seeing some `this` weirdness when I was debugging through some of the code a while back.
If that's the case then there may be alternative workarounds for that which don't involve copying these methods to the dialog class.

components/plugins/practice.js
```diff
@@ -377,13 +282,11 @@ export function addPracticePlugin() {
...
 
-		editor.addCommand(cmds.practice, () => {
-			launchPracticeEditor();
-		});
+		editor.addCommand(cmds.practice, () => launchPracticeEditor());
 
 		editor.addCommand(cmds.llmSuggest, () => {
 			LlmController.autoSuggest = true;
-			selectedText = '';
+			LlmController.selectedText = '';
```
mdgbayly, line 289: Is this command related to the autosuggest stuff that we discussed removing to remove the extra toolbar? Are we planning on removing this as part of this clean up PR or was that going to come later?
I think removing the autosuggest would also remove some additional code from each question file which would be good to reduce to overall merge diff too. But fine if it comes after this initial clean up.

JamesVaChao: Yes, I think this was an initial pass. Ian will clean up the button and auto suggest and test further.

components/plugins/practice.js
JamesVaChao: @erickang21 Is this unused? They're also in the individual questions. Make a note to look into this and potentially remove them.

components/plugins/practice.js
JamesVaChao: @erickang21 this can be removed along with the _UpdateValues in the practice correct? Did you need it for anything in particular or a reason why it had to be in practice?

components/practice/dropdown-blanks.js
erickang21: i will have to do some testing to see if we can get rid of it. because some of the html that comes from other files (think of the llmcontroller,s different banners) is styled in those sections.

components/plugins/practice.js
erickang21: i don't think this is necessary anymore. it was most likely to work on some code when we were first transitioning to centralizing things in llmcontroller

components/plugins/practice.js
erickang21: I'll need to take a look as well. I thought process blanks was part of the clearFields function, but maybe that isn't needed

components/plugins/practice.js
erickang21: update: i think `this.userSelectedText` is still being used in getLlmSuggestions()

mdgbayly:
@erickang21 
I think this line can probably be removed too?
https://github.com/BrightspaceUI/htmleditor/pull/1592/files#diff-edaba111609db2f3d183e60773a08ab7389faee40fca5e253575b9dcf9ad53a7L889
```
if (mediaQueryList.addEventListener) mediaQueryList.addEventListener('d2l-alert-button-press', () => LlmController.getLlmSuggestions(this._templateIndex, this.localize('practice:llm:topBanner:error')));
```
It seems to come from this PR
https://github.com/D2L-Labs/htmleditor-mirror/pull/33/files#diff-edaba111609db2f3d183e60773a08ab7389faee40fca5e253575b9dcf9ad53a7R831

But not sure it was ever used. Not clear what it means to add an event listener to the mediaQueryList in that way. Maybe it was copied from similar usages in the original practices code, but I'm not clear their listeners are doing anything either.
But it doesn't look like we even have any `d2l-alert` components anymore. That code is commented out.

mdgbayly:
@erickang21 

It looks like we added the toggle for the Short description field?

https://github.com/D2L-Labs/htmleditor-mirror/commit/e33e52794733f67bd88682fd97e73f65f0abaf63#diff-756c22ae6d9424ef0b604c18478ab11f25c80d9ad80df7ef529084eff40bd92c

Just wondering what was the reason for that and if it is something that needs to be reverted as part of this cleanup?

mdgbayly:
@erickang21 

Just digging a little bit more into how the the original practices code works, and how we're transforming our LLM output into something that renders as practices.

I'm slightly confused by how we generally seem to both call `_updateValue` and update the `activityData` directly. As it seems like these are both doing the same thing.

An example from the True/False question.

```
this.activityData.questions[0].text = contentFormatted;
this._updateValue('questions:0:text', contentFormatted);
```

https://github.com/BrightspaceUI/htmleditor/pull/1592/files#diff-b11d12e1ea5372df85cd8e1e40db59baff799f0629ef5839f240a7362e2c3fbbR378

Were you running into issues with getting things to re-render correctly, or am I misunderstanding how some of this is working. 

It just seems like this adds a lot of duplicated code. My perception is that the `_updateValue` approach is tied to the field and property abstractions related to how the practices use templates/schemas to try to generalize some of the code. But if we are explicitly updating properties that we know exist then we could probably just update the `activityData` directly, and then call `this.updateActivity()` to trigger a rerender?

Your comment on one of the other messages about `clearFields` was making me wonder too. I looked at the original implementation of that and it looks like you were originally just using `_updateValue` but subsequently that was modified to use both approaches too? But it doesn't seem entirely consistent in different places.

https://github.com/D2L-Labs/htmleditor-mirror/commit/f8c7de38fd72c6a4140851416f340509f6e4031d#diff-874eca19d6194a10d6f2f7c0162b5f0d8c6a7c358073ca1e7c808eb997652ca2L757

mdgbayly:
@erickang21 

Apologies :-) so many questions.

I guess a follow up question to my previous question about how we're updating the activity data is do you recall why we needed to have the `clearFields` function on each practice type in the first place?

It seems to be used in two places:
* When changing the question type
* When cancelling from the current editor

But in both those cases it seems like without the Gen AI integrations, the practices editor was handling things appropriately without requiring that clean up.

Again, apologies for all these searching questions, that are probably requiring you to pluck memories from the distant past. I can't remember what I did last week so not expecting you to remember just like that either. But these questions are to help me understand whether there are things we can do to cut down the size of the merge, but also just generally to help me understand how we might re-organize some of this code so there is less coupling between the Gen AI stuff and the original practices.

I'm sure there were good reasons why the code was added at the time.

components/practice/dropdown-blanks.js
erickang21: This CANNOT be removed - it provides styling that comes from LlmController.htmlType

components/plugins/practice.js
erickang21: Seems like they are used in the individual practice types - dropdown blanks and fill in the blanks. Not used in practice so I removed it.

mdgbayly:
Good to just merge this into `generative-ai-merge`. Eric can do follow up PRs for other things discussed here by creating a new PR off that branch.

erickang21:
> @erickang21 I think this line can probably be removed too? https://github.com/BrightspaceUI/htmleditor/pull/1592/files#diff-edaba111609db2f3d183e60773a08ab7389faee40fca5e253575b9dcf9ad53a7L889
> 
> ```
> if (mediaQueryList.addEventListener) mediaQueryList.addEventListener('d2l-alert-button-press', () => LlmController.getLlmSuggestions(this._templateIndex, this.localize('practice:llm:topBanner:error')));
> ```
> 
> It seems to come from this PR [D2L-Labs/htmleditor-mirror@d734618#diff-edaba111609db2f3d183e60773a08ab7389faee40fca5e253575b9dcf9ad53a7R818](https://github.com/D2L-Labs/htmleditor-mirror/commit/d734618c029202e34f14d5bd8a78f406e5c4e8ff#diff-edaba111609db2f3d183e60773a08ab7389faee40fca5e253575b9dcf9ad53a7R818)
> 
> But not sure it was ever used. Not clear what it means to add an event listener to the mediaQueryList in that way. Maybe it was copied from similar usages in the original practices code, but I'm not clear their listeners are doing anything either. But it doesn't look like we even have any `d2l-alert` components anymore. That code is commented out.



> @erickang21 I think this line can probably be removed too? https://github.com/BrightspaceUI/htmleditor/pull/1592/files#diff-edaba111609db2f3d183e60773a08ab7389faee40fca5e253575b9dcf9ad53a7L889
> 
> ```
> if (mediaQueryList.addEventListener) mediaQueryList.addEventListener('d2l-alert-button-press', () => LlmController.getLlmSuggestions(this._templateIndex, this.localize('practice:llm:topBanner:error')));
> ```
> 
> It seems to come from this PR [D2L-Labs/htmleditor-mirror@d734618#diff-edaba111609db2f3d183e60773a08ab7389faee40fca5e253575b9dcf9ad53a7R818](https://github.com/D2L-Labs/htmleditor-mirror/commit/d734618c029202e34f14d5bd8a78f406e5c4e8ff#diff-edaba111609db2f3d183e60773a08ab7389faee40fca5e253575b9dcf9ad53a7R818)
> 
> But not sure it was ever used. Not clear what it means to add an event listener to the mediaQueryList in that way. Maybe it was copied from similar usages in the original practices code, but I'm not clear their listeners are doing anything either. But it doesn't look like we even have any `d2l-alert` components anymore. That code is commented out.

seems like this was added for older work I did. removed it now.

erickang21:
> D2L-Labs/htmleditor-mirror@e33e527#diff-756c22ae6d9424ef0b604c18478ab11f25c80d9ad80df7ef529084eff40bd92c

No, this was intended. One of the things we did was make the short description section toggle-able.

erickang21:
> this.activityData.questions[0].text

some of this code is indeed duplicated. i'll remove parts where it's not necessary.

components/plugins/practice.js
erickang21: > Was trying to understand what is calling this `_processBlanks()` method and some of the related methods that it calls.
> 
> Might need to run the code to understand. I don't see them being called in this class, but there are similar methods in some of the other question type classes. I'm wondering if this is a result of the way in which the question specific `getLlmSuggestions` is being called, possibly with the `this` reference pointing to the `PracticeDialog` instance rather than the question component instance, hence some of these methods were copied as a workaround.
> 
> I vaguely recall seeing some `this` weirdness when I was debugging through some of the code a while back. If that's the case then there may be alternative workarounds for that which don't involve copying these methods to the dialog class.

process blanks needs to be kept, based on my testing. additionally, updating `this.questionText` is necessary because of processBlanks as well.

mdgbayly:
> > > [D2L-Labs/htmleditor-mirror@e33e527](https://github.com/D2L-Labs/htmleditor-mirror/commit/e33e527)#diff-756c22ae6d9424ef0b604c18478ab11f25c80d9ad80df7ef529084eff40bd92c
> > 
> > 
> > No, this was intended. One of the things we did was make the short description section toggle-able.
> 
> But that has implications for the product? Was that discussed with the Creator+ Practices PM/Designers? It's also a material product change so might have other implications e.g. docs/release notes etc.

In case it wasn't clear what I mean here, I don't think we can be making changes to the Practices as part of adding gen AI to them unless we discuss that with the Practices team/Surekha etc.

components/practice/dropdown-blanks.js
mdgbayly: Still not sure I understand though - the reference to the styles come from the LlmConroller.htmlType, but the styles shouldn't need to be duplicated?

PR #1599, opened 2023-08-15
Move Content Styler to new plugin framework

grant-cleary:
As part of the overall plugin refactor work, we also want to move the Creator+ plugins over to the new framework prior to shifting them out of this repo. This PR aims to start with the Content Styler plugin, since it's relatively simple compared to the other plugins.


components/plugins/contentstyler.js
```diff
@@ -3,11 +3,49 @@ import { requestInstance } from '@brightspace-ui/core/mixins/provider-mixin.js';
...
+	isAvailable(inlineLimited, context, pluginAvailabilityProps) {
+		return !inlineLimited
+			&& context.contentStyler
+			&& pluginAvailabilityProps.contentStyler
+			&& pluginAvailabilityProps.fullPage;
+	}
+
+	registerPlugin(context, helpers) {
+		helpers.eventRegistry.registerBeforeSetContentEvent(content => {
+			if (helpers.internalAPIWrapper.testFullPageStyleSheets(url => !safelistCSS.test(url), attrs => attrs.get('data-override') !== 'override')) return;
```
grant-cleary, line 24: Interestingly, I don't believe the way this was done before was actually working. In testing, TinyMCE won't ever return the full document as a parameter in its events, so this safelist check was always returning false.

components/plugins/contentstyler.js
```diff
@@ -3,11 +3,49 @@ import { requestInstance } from '@brightspace-ui/core/mixins/provider-mixin.js';
...
+
+	isAvailable(inlineLimited, context, pluginAvailabilityProps) {
+		return !inlineLimited
+			&& context.contentStyler
+			&& pluginAvailabilityProps.contentStyler
+			&& pluginAvailabilityProps.fullPage;
+	}
+
+	registerPlugin(context, helpers) {
+		helpers.eventRegistry.registerBeforeSetContentEvent(content => {
```
grant-cleary, line 23: I believe this was written as a `BeforeGetContent` handler because of how the fullpage stuff worked before. I'm not convinced that's still true, so I think this can now be switched to a `BeforeSetContent` handler like most similar plugins. We'll want to extensively test this though.

JakeProuse (approved):

components/plugins/contentstyler.js
grant-cleary: ... Oh, you're right. I don't know why I had it in my head that this was still a `Get` handler. Now that I think about it, I do recall when that change was made. 🤦

components/plugins/contentstyler.js
grant-cleary: I also remember conversations about this a while ago, but I don't recall what the outcomes were, or whether we were seeing the same thing I'm seeing now. 🤔

PR #1600, opened 2023-08-15
cleaned up a lot of code inside generativeai-cleanup based on comments

erickang21:


components/practice/true-false.js
```diff
@@ -375,7 +375,7 @@ class TrueFalsePractice extends RequesterMixin(Localizer(LitElement)) {
 			}
 			const contentFormatted = content.trim().replace('"""', '');
 			if (fieldToFill === fields.question) {
-				this.activityData.questions[0].text = contentFormatted;
+				//this.activityData.questions[0].text = contentFormatted;
```
JamesVaChao, line 378: don't comment this, just delete it. Also do a full testing of it. Make sure you save the question and try to edit it to verify that its working/saving correctly.

erickang21: did a full test, including inserting it, and i can confirm it works.

components/plugins/practice.js
```diff
@@ -449,7 +449,7 @@ class PracticeDialog extends RequesterMixin(Localizer(LitElement)) {
 		}
 
 		.d2l-config {
-			display: flex;
+
```
ianmclean2011, line 452: Extra space here can be removed.

ianmclean2011: Also, this change appears to impact pre-existing styling, so just want to make sure it isn't breaking anything Practices related?

components/practice/dropdown-blanks.js
```diff
@@ -672,24 +672,20 @@ class DropdownBlanksPractice extends RequesterMixin(Localizer(LitElement)) {
...
-				this.activityData.questions[0].text = contentFormatted;
 				this._updateValue('questions:0:text', contentFormatted);
 				this.questionText = contentFormatted;
 
 			} else if (fieldToFill === fields.answer) {
 				const answerIndex = this.activityData.questions[0].answers.map(function(answer) { return answer.id; }).indexOf(answerId);
 				if (answerIndex > -1) {
-					this.activityData.questions[0].answers[answerIndex].answers[0].text = contentFormatted;
 					this._updateValue(`questions:0:answers:${answerIndex}:answers:0:text`, contentFormatted);
-					this._updateActivity();
```
ianmclean2011, line 684: Curious ... is there a reason this call to `_updateActivity()` is being removed, but other calls in other areas of code aren't being removed?

components/plugins/practice.js
erickang21: seems like this should be kept based on my testing. i'll fix this

components/practice/dropdown-blanks.js
erickang21: on second thought, this isn't necessary for cleanup so i will add it back

components/plugins/practice.js
```diff
@@ -1012,7 +1012,7 @@ class PracticeDialog extends RequesterMixin(Localizer(LitElement)) {
...
 
 			case 3:
-				return html`<d2l-true-false-practice id="d2l-true-false-practice" @updateValue="${this._updateValue}" @activityUpdate="${this._activityUpdated}" @activitySave="${this._saveActivity}" @togglePreview="${this._togglePreview}" @handleClose="${this._checkForChangesBeforeClose}" .activityData="${this.activityData}" ?display-feedback="${this._modifiers[0].value}" ?has-hint="${this._modifiers[1].value}" ?use-enumeration="${this._modifiers[2].value}" .editorText="${editorText}" .titleText="${titleText}" .summary="${this.summary}"></d2l-true-false-practice>`;
+				return html`<d2l-true-false-practice id="d2l-true-false-practice" @activityUpdate="${this._activityUpdated}" @activitySave="${this._saveActivity}" @togglePreview="${this._togglePreview}" @handleClose="${this._checkForChangesBeforeClose}" .activityData="${this.activityData}" ?display-feedback="${this._modifiers[0].value}" ?has-hint="${this._modifiers[1].value}" ?use-enumeration="${this._modifiers[2].value}" ?display-short-description="${this._modifiers[3].value}" .editorText="${editorText}" .titleText="${titleText}" .summary="${this.summary}"></d2l-true-false-practice>`;
```
ianmclean2011, line 1015: I believe @JamesVaChao just removed the display-short-description part.

ianmclean2011 (approved):
Looks ok to me.

components/practice/dropdown-blanks.js
JamesVaChao: So are we removing this or adding it back? and jsut to confirmed you tested it a few times to make sure things are updating properly?

JamesVaChao: you can either keep it (and test it) or remove these change and not have to test that part

PR #1603, opened 2023-08-16
US156138 - Move Practices to new plugin framework

jschlumkoski:
@grant-cleary This is me taking an early stab at swapping the Practices plugin to the new way of doing things. A lot of trial and error here, but I believe I have things mostly working! 😅 

I did hit a couple of roadblocks that I think I'll need some assistance with. There's some functionality I wasn't able to replicate because of plugins not having access to the editor document. I know there's a concept of "privileged plugins" now, but I'm not sure if ours should be added to that list or not, and I didn't want to assume so. 

The things we can't currently replicate in Practices:

1. The old Practices plugin used the `SetContent` listener to check for any Practices being included in the page, and made sure the CSS and JS dependencies were added to the page (in case they were removed in the source code editor). I know I can register an event for this still (`helpers.eventRegistry.registerSetContentEvent`), but I can't access the document body to check for Practices or not. However, after switching to `addFullPageStyleSheets` and `addFullPageScripts`, I've noticed that if you remove the JS and CSS dependency references in the source code editor, it will restore our JS one but not our CSS one. (This will also affect Insert Element / Layouts)
2. The old Practices plugin had contextual toolbar buttons for moving a Practice up or down the page inside of it's current container by basically swapping the Practice node with either it's previous or next sibling. However, I can't reliably get the parent container with `findComposedAncestor`, because if the Practice being moved is in the body, it comes back null. The same issue prevents me from calling `insertBefore()` on the node's parent if it's the body. (This will also affect Insert Element / Layouts)

Mostly looking for direction on how to handle these last pieces!


grant-cleary:
> 1. The old Practices plugin used the `SetContent` listener to check for any Practices being included in the page, and made sure the CSS and JS dependencies were added to the page (in case they were removed in the source code editor). I know I can register an event for this still (`helpers.eventRegistry.registerSetContentEvent`), but I can't access the document body to check for Practices or not. However, after switching to `addFullPageStyleSheets` and `addFullPageScripts`, I've noticed that if you remove the JS and CSS dependency references in the source code editor, it will restore our JS one but not our CSS one. (This will also affect Insert Element / Layouts)

Interesting. I think I'll need to poke at this somewhat, since I would have expected this behaviour to remain unchanged. When I was playing with Content Styler, I noticed that the stylesheets seemed to be persistent. I'll pull down your branch either today or tomorrow and will try to see what's going on.

> 2. The old Practices plugin had contextual toolbar buttons for moving a Practice up or down the page inside of it's current container by basically swapping the Practice node with either it's previous or next sibling. However, I can't reliably get the parent container with `findComposedAncestor`, because if the Practice being moved is in the body, it comes back null. The same issue prevents me from calling `insertBefore()` on the node's parent if it's the body. (This will also affect Insert Element / Layouts)

Ah, interesting. I think this is just a gap I need to close. There's a check I do to make sure plugins can only interact with contents within the document body that's blocking these actions. Is this swapping all that needs to be done? i.e. Are you just fetching the parent so you can pass it into `insertBefore`?

grant-cleary:
Though for (2), maybe it makes more sense to return the body, but just restrict the operations that can be done on it to avoid, say, messing with its attributes or styles.

components/plugins/practice.js
```diff
@@ -38,14 +38,171 @@ import { inputLabelStyles } from '@brightspace-ui/core/components/inputs/input-l
...
+	}
+
+	isAvailable() {
+		return true;
+	}
+
+	registerPlugin(context, helpers) {
+		const dialogFactory = dialog => {
+			const editorDocument = new DOMParser().parseFromString(helpers.internalAPIWrapper.getContent(), 'text/html');
+			const component = context.component;
```
grant-cleary, line 63: Hmm, this isn't supposed to be available on the context object, but you also shouldn't need it anymore. I'll need to take a look to see if this is somehow bleeding in.

components/plugins/practice.js
```diff
@@ -38,14 +38,171 @@ import { inputLabelStyles } from '@brightspace-ui/core/components/inputs/input-l
...
+
+			if (layout) {
+				dialog._forbidden = true;
+				dialog.addEventListener('d2l-practice-forbidden-dialog-close', event => {
+					if (!event || !event.detail || event.detail.action !== 'close') {
+						return;
+					}
+
+					component._dialogOpen = false;
+					component.focus();
```
grant-cleary, line 89: This should be handled for you now when you register the dialog factory. Once the promise resolves, the framework handles closing the dialog and re-applying focus to the core editor. :)

components/plugins/practice.js
```diff
@@ -38,14 +38,171 @@ import { inputLabelStyles } from '@brightspace-ui/core/components/inputs/input-l
...
-const name = 'd2l-practice-dialog';
+const name = 'd2l-htmleditor-practice-dialog';
+
+export class PracticesPlugin {
+	get name() {
+		return name;
+	}
+
+	isAvailable() {
+		return true;
```
grant-cleary, line 57: I imagine you'll want to leave this as is for testing, but when this merges, you'll want this to reflect all of the conditions that used to be here: https://github.com/BrightspaceUI/htmleditor/blob/main/components/toolbar/toolbar-full.js#L70

You can find an example of how that's done here: https://github.com/BrightspaceUI/htmleditor/blob/main/components/plugins/contentstyler.js#L15

Properties that were once on the component are now on the `pluginAvailabilityProps` object.

grant-cleary: Though now I'm kind of wondering if it makes more sense to expose an `isAvailableInlineLimited()` function instead, because I feel like most plugins are going to want that case to be false. 🤔 

When the dust is settled on this, I'll definitely be asking you and @JakeProuse for thoughts on what parts of this process didn't make enough sense, or what needs to be more explicitly documented.

PR #1605, opened 2023-08-16
Replace insertBefore node API with insertAdjacentElement to keep private methods private, allow body to surface but with restricted properties

grant-cleary:
There were some gaps in the `EditorNode` APIs that are blockers for Practices, Insert Element, and Layouts. This PR attempts to address the ones I currently know about:

1. `childNodes` wasn't exposed on the API. **Solution:** Expose it.
2. `insertBefore` wasn't properly functional because it requires reaching into the private DOM node of the node objects. **Solution:** Replace with `insertAdjacentElement`, which is more restrictive because it doesn't allow moving nodes across documents, but should still cover the relevant use cases (in combination with `remove`).
3. We need to be able fetch the document body, as it's possible for layouts, practices, etc. to be direct children. **Solution:** Relax the body escape check to allow the body to be returned, but restrict what can be done with it to prevent modifying the class, attributes, etc.


components/selection.js
```diff
@@ -50,14 +52,21 @@ class EditorNode {
 	}
 
 	get attributes() {
+		if (this.#isBody()) return undefined;
 		return this.#domNode.attributes;
 	}
 
+	get childNodes() {
+		return [...this.#domNode.childNodes].map(node => new EditorNode(node));
```
grant-cleary, line 60: This will behave slightly differently from how it behaves in the browser, since it will no longer be a live list. Nothing that currently relies on this needs the list to be live, however.

components/selection.js
```diff
@@ -128,8 +141,9 @@ class EditorNode {
 		return this.#domNode.hasAttribute(attribute);
 	}
 
-	insertBefore(newNode, refNode) {
-		this.#domNode.insertBefore(newNode.$domNode, refNode.$domNode);
+	insertAdjacentElement(position, node) {
+		if (this.#isBody()) return;
+		return this.#domNode.insertAdjacentHTML(position, node.outerHTML);
```
grant-cleary, line 146: Ignore the `$` typo above 😅. This wouldn't have worked anyways since `#domNode` is private. What we can do instead is use a little bit of trickery to get the same behaviour with a similar API.

JakeProuse (approved):

PR #1613, opened 2023-08-17
True or false updating properly

JamesVaChao:
Removed spacing changes
Removed unused bools
Reverted some code changes back to original
Consistency in some variable/functions


components/practice/true-false.js
```diff
@@ -384,16 +384,25 @@ class TrueFalsePractice extends RequesterMixin(Localizer(LitElement)) {
 
 		LlmController.loading = true;
 		await this.lc.getLlmStream(questionType, { content: LlmController.getDisplayText(textToUse) }, processContent);
+		console.log('SETTING LOADIGN TO FALSE');
```
ianmclean2011, line 387: Remove

components/practice/true-false.js
```diff
@@ -384,16 +384,25 @@ class TrueFalsePractice extends RequesterMixin(Localizer(LitElement)) {
...
 
 	// eslint-disable-next-line sort-class-members/sort-class-members
 	clearFields() {
+		LlmController.loading = false;
 		this._updateValue('questions:0:text', '');
 		this._updateValue('questions:0:correctAnswers', []);
 	}
 
 	_buildField(property, index, disabled) {
+		console.log('DISABLED? ', disabled);
```
ianmclean2011, line 405: Remove

components/practice/dropdown-blanks.js
```diff
@@ -743,6 +742,7 @@ class DropdownBlanksPractice extends RequesterMixin(Localizer(LitElement)) {
 		// Change scoring to "Correct Inputs" by default
 		this.activityData.scoreCalculation = 1;
 		this._updateValue('scoreCalculation', 1);
+		this._updateActivity();
```
mdgbayly, line 745: Doesn't `_updateValue` already call `_updateActivity`?

components/plugins/practice.js
```diff
@@ -461,6 +460,10 @@ class PracticeDialog extends RequesterMixin(Localizer(LitElement)) {
 			width: 100%;
 		}
 
+		.d2l-config > d2l-button {
```
ianmclean2011, line 463: Does this interfere with anything Practices?

ianmclean2011 (approved):

components/practice/multiple-choice.js
```diff
@@ -480,6 +480,7 @@ class MultipleChoicePractice extends RequesterMixin(Localizer(LitElement)) {
 		}
 
 		this._updateValue('questions:0:randomize', true);
+		this._updateActivity();
```
mdgbayly, line 483: ditto: I thought `_updateValue` calls `_updateActivity`  - I might be missing something around how `updateActivity` is working.

components/practice/true-false.js
mdgbayly, line 387: nit: no console statements?

components/plugins/practice.js
JamesVaChao: This was what was originally in practice.js, our main-merge shows we changed it to subtle, but we should have both styles kept

components/practice/true-false.js
JamesVaChao: agreed, will comb them over

components/practice/multiple-choice.js
JamesVaChao: Yes, so the only file the truely needed the updateActivity at the end. was true/false. The other ones, i added just for consistency and in case anything was added/not updated properly as a fail safe. I can remove them as they're basically redundant.

components/practice/true-false.js
JamesVaChao: will comb through for console.logs

PR #1615, opened 2023-08-18
Jprouse/us156139 convert layouts plugin to use new api

JakeProuse:
Rally Story: https://rally1.rallydev.com/#/?detail=/userstory/719095595769&fdp=true


components/plugins/layouts.js
```diff
@@ -37,20 +38,177 @@ export function setLayoutsNotContentEditable(htmlDoc) {
...
+		return pluginName;
+	}
+
+	isAvailable() {
+		return true;
+	}
+
+	registerPlugin(context, helpers) {
+		const component = context.component;
+		const localize = component.localize.bind(component);
```
grant-cleary, line 52: Hmm, this reminds me that localization is still a bit of a tetchy question. The component isn't available on the context object anymore, so this won't work... But I do need (for now) to pass the localizer along, which I don't currently do. Eventually we're going to have a solution that doesn't require using the editor's localizer, so your lang terms will be able to live in your repo as well, but we're not there quite yet.

components/plugins/layouts.js
```diff
@@ -37,20 +38,177 @@ export function setLayoutsNotContentEditable(htmlDoc) {
...
+		const layouts = layoutsHtml(localize, context.useNewPluginFramework, null, helpers);
+
+		const dialogFactory = dialog => {
+			const editorDocument = new DOMParser().parseFromString(helpers.internalAPIWrapper.getContent(), 'text/html');
+
+			dialog.lang = editorDocument.querySelector('html').getAttribute('lang') || 'en-us';
```
grant-cleary, line 58: Interesting. I think I should also pass this in via the helpers.

components/plugins/layouts.js
```diff
@@ -37,20 +38,177 @@ export function setLayoutsNotContentEditable(htmlDoc) {
...
+
+			component._dialogOpen = true;
+			dialog.opened = true;
```
grant-cleary, line 61: Note that you don't need these anymore. The automatic dialog stuff will now handle both opening and closing the dialog when relevant, including setting the right properties on the editor.

components/plugins/layouts.js
```diff
@@ -37,20 +38,177 @@ export function setLayoutsNotContentEditable(htmlDoc) {
...
+			if (layout) {
+				dialog._forbidden = true;
+
+				dialog.addEventListener('d2l-layout-dialog-close', event => {
+					if (!event || !event.detail || event.detail.action !== 'close') {
+						return;
+					}
+
+					component._dialogOpen = false;
+					component.focus();
```
grant-cleary, line 76: Same as above; this is now handled for you.

components/plugins/creator-plus.js
```diff
@@ -18,6 +20,28 @@ const foundDependency = (editorDocument, path, tag) => {
...
+};
+
+export const addDependencies = (helpers, scripts) => {
+	for (const script of scripts) {
+		helpers.internalAPIWrapper.addFullPageScripts(
+			helpers.internalAPIWrapper.getContent(),
+			[{
+				url: script.src,
+				attrs: new Map([['defer', '']])
+			}]
```
grant-cleary, line 40: FYI, this already takes in an array of these objects, so I think instead of the outer loop you could do something like this:
```
export const addDependencies = (helpers, scripts) => {
	helpers.internalAPIWrapper.addFullPageScripts(
		helpers.internalAPIWrapper.getContent(),
		scripts.map(script => ({ url: script.src, attrs: new Map([['defer', '']]) }))
	);
};
```
Caveat: I haven't checked the syntax to make sure this works properly.

components/selection.js
```diff
@@ -168,6 +178,10 @@ export class EditorNode {
 		return window.getComputedStyle(this.#domNode, null);
 	}
 
+	getDomNode() {
+		return this.#domNode;
```
grant-cleary, line 182: This we definitely don't want, since exposing the DOM node defeats the purpose of the wrapper. Where does this get used?

grant-cleary: Oh... That was me getting the API mixed up with another TinyMCE one. 🤦 

Is there ever a case where you need this to work for non-wrapped DOM nodes? I have an idea of how I can make this work with `EditorNode`s.

JakeProuse: No I think in the new plugin I'd only be doing it with wrapped nodes 👍 If you can make it work with EditorNodes that'd be poifect

grant-cleary: Excellent, then I think this should do the trick: https://github.com/BrightspaceUI/htmleditor/pull/1638

components/selection.js
```diff
@@ -220,7 +220,7 @@ export class EditorNode {
 
 	querySelectorAll(selector) {
 		const descendantNodes = this.#domNode.querySelectorAll(selector);
-		return descendantNodes.map(node => new EditorNode(node, this.#selection));
+		return [...descendantNodes].map(node => new EditorNode(node, this.#selection));
```
JakeProuse, line 223: @grant-cleary Added a spread operator here to make the `NodeList` an `Array` since they don't have the `map` method. Noticed the [findElements method](https://github.com/BrightspaceUI/htmleditor/blob/b7057906f4a5ec74b9e100de9bb2424610e3bbfd/components/api-wrapper.js#L92) has this already so just added it here.

components/plugins/layouts.js
```diff
@@ -37,20 +39,244 @@ export function setLayoutsNotContentEditable(htmlDoc) {
...
+							if (
+								range.startOffset === 0 &&
+								range.commonAncestorContainer &&
+								range.commonAncestorContainer.classList.contains(layoutColumnClass)
+							) {
+								e.preventDefault();
+								e.stopPropagation();
+							}
+						} else if (e.keyCode === 9) {
+							e.preventDefault();
```
JakeProuse, line 248: @grant-cleary When I activate the listener with tab I get an error on `e.preventDefault();`. Looks like this may be an issue with how the [event is deconstructed](https://github.com/BrightspaceUI/htmleditor/blob/b7057906f4a5ec74b9e100de9bb2424610e3bbfd/components/selection.js#L131) ([similarish stack overflow](https://stackoverflow.com/a/49616808)). 

Changing it to `const eventProps = { ...e }` works but then we get a bunch of extra stuff, wondered what thoughts you had!
image

components/context-menu/formatting/formatting.js
```diff
@@ -29,11 +30,11 @@ export class FormattingContextMenuPlugin {
 	doRender(range, node, useNewPluginFramework) {
 		if (range.collapsed) return false;
 		if (node.nodeType === Node.ELEMENT_NODE && disallowedNodes.has(node.tagName)) return false;
+		if (node.nodeType === Node.ELEMENT_NODE && node.classList.contains(layoutClass)) return false;
```
JakeProuse, line 33: @grant-cleary If you're curious why I have this twice, for some reason when I arrow key over the layout it brings up the Formatting context menu. It doesn't when I add this condition before inNonEditable. I think I'm a little lost in the flow of the isNonEditable function, but is there a better way to check for this? As it is with these two lines you can highlight text inside a layout and it'll bring up the formatting menu which is what I need.

grant-cleary: So this line you've added right here will short-circuit the `non-editable` check entirely and just block rendering if the class on the selected node contains `layoutClass`. `isNonEditable` is used with `findComposedAncestor` to walk up the DOM and try to make sure the element you're editing isn't in a non-editable ancestor. So returning `false` in there doesn't early exit; it just prompts `findComposedAncestor` to try the next level up.

What I'm confused about is why this behaviour is different now. I would be curious if there's a difference in behaviour between `node.findComposedAncestor` and the original `findComposedAncestor` that's causing this.

components/context-menu/formatting/formatting.js
```diff
@@ -5,6 +5,7 @@ import { html, LitElement } from 'lit';
 import { cmds } from '../../commands.js';
 import { ContextMenuToolbarMixin } from '../context-menu-toolbar-mixin.js';
 import { findComposedAncestor } from '@brightspace-ui/core/helpers/dom.js';
+import { layoutClass } from '../../d2l-layouts.js';
```
grant-cleary, line 8: As a sidenote, remember that you won't be able to do this once your plugins are in a new repo, because the editor won't be able to import any of your stuff without creating a circular reference.

JakeProuse: Hmm, the d2l-layouts file may end up living in the editor since `layoutsHtml` is used in the context menu. Unless we can also setup some context menu loading like plugins, so we could export those from our repo 🤔

grant-cleary: So with context menu plugins now registered via the registry, that can indeed live in your repo 😄. There shouldn't be anything the editor actually has to import to support your plugins adding context menu items.

components/context-menu/formatting/formatting.js
JakeProuse: Yeah it threw me for a loop since the function looks very similar to as it was before. In 20.23.9 on my [sitelord site](https://slims.dev.brightspace.com/login?url=https://layoutplugintst.devlms.desire2learn.com) the formatting context menu doesn't pop up, but if I switch the BSI to 20.23.10 then it starts showing up.

components/context-menu/formatting/formatting.js
JakeProuse: Oh sweet, so we can just carry over a copy and register it that way without an issue?

components/context-menu/formatting/formatting.js
JakeProuse: Ah okay so the cause for the changes is just on 20.23.10, that's good news :D

grant-cleary: So just to clarify, in 20.23.10, are these two additional lines necessary? If so, I would still be curious what changed.

JakeProuse: ln: 37 is in both, so I think it's needed. Ln; 33 is new and needed to stop the formatting context menu from popping up in 20.23.10. Ln:33 isn't needed in 20.23.9 though so something's up

grant-cleary: I would be curious what the output of Ln 46 is in both versions. I'm going to guess that the output of `findComposedAncestor` isn't the same. I would be curious what it's spitting out in both cases.

JakeProuse: I'm trying to test this out locally but I'm finding that the behaviour is different than on the sitelord site. Taking the HTMLEditor versions from BSI 20.23.9 and 20.23.10 the formatting context menu shows up on both even though it didn't on the sitelord.

grant-cleary: Very odd... Well if this works for now, I have no problem with this little hack in the interim, though I would like to investigate more to make sure there isn't a broader problem with the new API.

JakeProuse: Our test dev will be doing some thorough QA on Elements/Practices/Layouts once any of them are merged, so hopefully we can suss out any issues ahead of time!

components/toolbar/toolbar-full.js
```diff
@@ -65,7 +65,7 @@ export function getPlugins(component, useNewPluginFramework) {
 		!useNewPluginFramework ? addImagePlugin() : '',
 		(useNewPluginFramework || component.noInsertCode || (component._context && component._context.noInsertCode)) ? '' : addInsertCodePlugin(),
 		!useNewPluginFramework ? addISFPlugin() : '',
-		(component.fullPage || requestInstance(component, 'd2l-insert-element-dialog-open')) && component.layouts ? addLayoutsPlugin() : '',
+		(!useNewPluginFramework && (component.fullPage || requestInstance(component, 'd2l-insert-element-dialog-open'))) && component.layouts ? addLayoutsPlugin() : '',
```
grant-cleary, line 68: As a sidenote, I'm curious whether we can revisit this `d2l-insert-element-dialog-open` check here. We can make it work with the new framework, but I'm also wondering whether this `(component.fullPage || requestInstance(...))` check is even necessary. If `layouts` can work outside of `fullPage` mode (e.g. in the case where the insert-element dialog is open), is there a reason we need to have this restriction? 🤔

components/plugins/layouts.js
```diff
@@ -37,20 +39,260 @@ export function setLayoutsNotContentEditable(htmlDoc) {
 	}
 }
 
+export class LayoutsPlugin {
+	get name() {
+		return pluginName;
+	}
+
+	isAvailable() {
+		return true;
```
grant-cleary, line 48: This will need to reflect the logic I reference here: https://github.com/BrightspaceUI/htmleditor/pull/1615/files#r1311886086, which (for context) is why I'm curious if we need that first check. I can make it work if we do, so it's not a huge deal.

components/toolbar/toolbar-full.js
JakeProuse: I think we initially set up layouts so that it would only show up in the full editor, and not the limited/inline ones. Is that still what fullPage means when it's true?

JakeProuse: Would checking for !inlineLimited be the same as checking for fullPage?

grant-cleary: No, `fullPage` means we're in a context that loads an entire HTML document... i.e. Think of creating a new Content file. Some of the other C+ plugins have that as a pre-requisite because they need to load scripts into the document, which we can't do if we're only working on a document fragment. But I don't think layouts falls into that case, right? Because the insert-element dialog doesn't set that up.

JakeProuse: In here I do a similar thing to insert-element where I add stylesheets/scripts [here](https://github.com/BrightspaceUI/htmleditor/pull/1615/files#diff-7075764a8e2c88e837f68db06dc81b6ab7ea0950f00556e18b48104696a7603eR215), but does that require being a fullPage plugin? I also query the body with [findElements](https://github.com/BrightspaceUI/htmleditor/pull/1615/files#diff-7075764a8e2c88e837f68db06dc81b6ab7ea0950f00556e18b48104696a7603eR222), but since it's going through the API does it also not require being fullPage?

grant-cleary: Actually yeah, that does require `fullPage`... `addFullPageStylesheets/Scripts` are no-ops if `fullPage` isn't enabled. So how is insert-element using layouts in its internal editor then? 😕

test/editor-node.test.js
```diff
@@ -386,7 +389,15 @@ describe('d2l-htmleditor node wrapper', () => {
...
 
-			expect(handlerSpy.args[0][0]).to.deep.equal(expectedEvent);
+			// Iterate over our non-function values and assert their values.
+			for (const [key, value] of Object.entries(expectedEvent)) {
+				if (typeof value === 'function') continue;
+
+				expect(handlerSpy.args[0][0][key]).to.equal(value);
+			}
+
+			expect(preventDefaultSpy.notCalled).to.be.true;
```
JakeProuse, line 399: @grant-cleary I wasn't sure what to have here, the methods don't get called when setting the event in the handler so I couldn't check for calledOnce or anything. Any input on this?

components/toolbar/toolbar-full.js
grant-cleary: Hmm, there's probably some sort-of-sketchy stuff we could figure out, but TinyMCE doesn't make it easy out of the box because it doesn't allow styles in body elements. We can't add it to the head in the non-fullPage case because that's the Brightspace UI and the editor can't/shouldn't be messing with it. Though I'm curious how this works with the flag off. The inability to style non-fullPage editors isn't new with this flag. 🤔

test/editor-node.test.js
grant-cleary: Ah, good point. There isn't a way we can actually force a call into `stopPropagation` without directly invoking it on the event... Which then defeats the purpose of checking the spies. Maybe it makes more sense to just verify that keys called `stopPropagation` and `preventDefault` exist with type `function`. We can't verify that they're just pass-throughs as intended, but I think that might be the best we can do.

components/toolbar/toolbar-full.js
JakeProuse: Currently with the flag off the [SetContent hook](https://github.com/BrightspaceUI/htmleditor/blob/main/components/plugins/layouts.js#L190C5-L190C27) in layouts will run when inserting a layout into the Insert Element dialog. This adds the editor.css [here](https://github.com/BrightspaceUI/htmleditor/blob/main/components/plugins/creator-plus.js#L65) to the head which styles the layout.

test/editor-node.test.js
```diff
@@ -386,7 +391,15 @@ describe('d2l-htmleditor node wrapper', () => {
 			const expectedEvent = Object.assign({}, mockEvent);
 			delete expectedEvent.someOtherMethod;
 
-			expect(handlerSpy.args[0][0]).to.deep.equal(expectedEvent);
+			const args = handlerSpy.args[0][0];
+			for (const [key, value] of Object.entries(args)) {
+				if (typeof value === 'function') {
+					// Check that only the methods we allow are exposed.
+					expect(expectedEventMethods.includes(key)).to.be.true;
```
JakeProuse, line 398: @grant-cleary thoughts on this setup? I wanted to make it less manual to update so I added an array of the methods we want passed in. If we wanted to make it even less manual we could use a constant to replace the array and import it to [here](https://github.com/BrightspaceUI/htmleditor/pull/1615/files#diff-ad170e923c414d200f618a966ccd4877efb020f91a5751898fcd3d13b6da868eR132) to loop over, but I wanted to check if there'd be any import issues although I see things are already imported anyway.

components/toolbar/toolbar-full.js
grant-cleary: Oh, so it only applies within the editor iframe. That's kind of iffy because it's then expecting Insert Element to apply the appropriate styles to the overall document. I'll have to think more about this use case, because it's definitely an edge case.

test/editor-node.test.js
grant-cleary: I think this setup is pretty good. The only thing I might suggest is that we add a quick check to make sure the number of keys is the same size for the mock event and the spy. That would guard against someone making a change to `onkeydown` that would add a new method for whatever reason. Importing a common constant is probably overkill, since if we add that length check, then if the test starts failing, it should be fairly obvious why.

components/toolbar/toolbar-full.js
JakeProuse: Yeah seeing as nothing you drop in the dialog requires the css except layouts its very much an edge case.

test/editor-node.test.js
JakeProuse: That sounds good, I'm never sure how much is overkill on testing 😆

grant-cleary: Haha yeah, that's always my struggle.

components/toolbar/toolbar-full.js
JakeProuse: That would work, the InsertElement SetContent hook will run once you've submitted the dialog and add the editor.css to the fullPage editor anyway so the iframe editor's head content can def just be temporary. Helps having all of our plugins using the same css so they have parity that way. Does InsertStuff inject [this one](https://github.com/BrightspaceUI/htmleditor/blob/7ce0c0d0ffe3f082583510a27b9bc6c49d1497bd/components/plugins/creator-plus.js#L2)?

components/plugins/layouts.js
```diff
@@ -37,20 +39,262 @@ export function setLayoutsNotContentEditable(htmlDoc) {
...
+export class LayoutsPlugin {
+	get name() {
+		return pluginName;
+	}
+
+	isAvailable(inlineLimited, context, pluginAvailabilityProps) {
+		return !inlineLimited &&
+			context.layouts &&
+			pluginAvailabilityProps.fullPage;
+	}
```
JakeProuse, line 51: @grant-cleary Is there a good way to use requestInstance here? I noticed that you're doing it for [isfContextIds](https://github.com/BrightspaceUI/htmleditor/blob/7ce0c0d0ffe3f082583510a27b9bc6c49d1497bd/components/api-wrapper.js#L13)

grant-cleary: Not as of yet. I'm wondering how sure we are this is needed, circling back on what we discussed earlier. Is it actually necessary to include the `fullPage || insert-element` check, or could we just omit it? If we did omit it, there's a chance someone could set the `layouts` attribute on the editor and get weird behaviour out of it, but is that such a big deal? 🤔

If we want the behaviour to just have parity, I think the timing works out that the `pluginRegistry` code could call `requestInstance` and pass down the result into the `pluginAvailabilityProps` object.

components/toolbar/toolbar-full.js
grant-cleary: InsertElement, you mean? I believe so, by virtue of this: https://github.com/BrightspaceUI/htmleditor/blob/407cbad5bd0ca8a37a875b9a1a874a484b65e50b/components/plugins/lcs.js#L108

JakeProuse: Oh haha sorry you said Insert Stuff in the previous comment and I got confused 😄

grant-cleary: Whoops. Corrected that. 😆

components/api-wrapper.js
```diff
@@ -126,6 +126,11 @@ export class InternalAPIWrapper {
 		this.#component.focus();
 	}
 
+	focusToolbar() {
+		const contextMenu = this.#component.shadowRoot.querySelector('d2l-htmleditor-toolbar-full');
```
grant-cleary, line 130: Not for this PR, but note to self that I want to change this later, because the selector for the toolbar changes depending on type. I'll figure out a way to centralize this logic a bit better.

components/context-menu/formatting/formatting.js
```diff
@@ -29,11 +29,11 @@ export class FormattingContextMenuPlugin {
 	doRender(range, node, useNewPluginFramework) {
 		if (range.collapsed) return false;
 		if (node.nodeType === Node.ELEMENT_NODE && disallowedNodes.has(node.tagName)) return false;
+		if (node.nodeType === Node.ELEMENT_NODE && node.classList.contains('d2l-cplus-layout')) return false;
```
grant-cleary, line 32: Did we ever find out why this became necessary? Not that I think it's a huge issue that it's here, but our of curiosity. 🤔

JakeProuse: No I'm still not sure, I couldn't tell what the difference was locally when testing 😕

grant-cleary (approved):

PR #1618, opened 2023-08-18
spacing, eslint disabled, optional chaining, unused button, console

JamesVaChao:


components/practice/multiple-choice.js
```diff
@@ -443,7 +443,7 @@ class MultipleChoicePractice extends RequesterMixin(Localizer(LitElement)) {
 					this.activityData.questions[0].correctAnswers = parsedCorrectAnswer;
 					this._updateValue('questions:0:correctAnswers', parsedCorrectAnswer);
 				} catch (e) {
-					console.error('Error parsing correct answer ', e);
+					return;
```
ianmclean2011, line 446: I'm fine with this (and all other examples of it) for the moment, but I would like us to do something more meaningful with this in future. Could you add this to the list of things to do in the doc?

components/practice/dropdown-blanks.js
```diff
@@ -394,6 +393,128 @@ class DropdownBlanksPractice extends RequesterMixin(Localizer(LitElement)) {
 		}
 	}
 
+	clearFields() {
```
ianmclean2011, line 396: Just confirming that you didn't actually change any of the code, but just moved things around?

components/plugins/practice.js
```diff
@@ -327,15 +327,6 @@ export function addPracticePlugin() {
 			}
 		});
 
-		editor.ui.registry.addButton('selection-llm', {
```
ianmclean2011, line 330: Was this something we just missed?

ianmclean2011 (approved):
Just a few comments to answer, but the changes look good.

components/plugins/practice.js
grant-cleary: This would have been necessary if this button had ever been included in the Practices context menu. I don't see evidence that it was added there though (at least in the merge PR).

JamesVaChao: yes, we missed it the first pass, grant caught it

components/practice/dropdown-blanks.js
JamesVaChao: yes, no changes, just refactored due to the eslint-disable-class-sort

components/practice/multiple-choice.js
JamesVaChao: yes sure, we'd probably have some sort of telemetry in the future

PR #1631, opened 2023-08-23
Add support for extensible toolbar items

grant-cleary:
This is a very limited move towards defining our toolbar as data instead of having the toolbar make decisions about which plugins to load. For now, it's restricted to what I'm calling "extensible" toolbar items, which is just stuff tacked onto the end of the "Insert" menu, provided by external plugins not defined within the HTML editor repo.


components/toolbar/menu-item.js
```diff
@@ -87,8 +89,10 @@ class MenuItem extends MenuItemSelectableMixin(ToolbarItemMixin(LitElement)) {
 	}
 
 	render() {
+		const icon = this.external ? this.icon : (icons[this.icon] || '');
```
grant-cleary, line 92: I'm a little on the fence about this one. There's no guarantee this will be an icon if it's being set externally, but we obviously don't want to force folks to place their icons in the editor repo if they're being used solely by an external plugin. I'm curious if there's any kind of validation that can be done to make sure this is an SVG like we would expect (perhaps something more robust than a regex).

grant-cleary: Actually, this is interesting: https://lit.dev/docs/api/templates/#svg

grant-cleary: This would require having the plugin provide the innards of the SVG, and we would be responsible for constructing the SVG out of it. It's kind of weird.

htmleditor-mixin.js
```diff
@@ -269,6 +275,12 @@ export const HtmlEditorMixin = superclass => class extends Localizer(LabelledMix
 			 * @type {boolean}
 			 */
 			sourceEditable: { type: Boolean, attribute: 'source-editable' },
+			/**
+			 * Whether to render the editor in `full`, `inline`, or `inline-limited` mode
+			 * @type {'full'|'inline'|'inline-limited'}
+			 * @default "full"
+			 */
+			type: { type: String },
```
grant-cleary, line 283: Potentially controversial, but if the editor is constructing the toolbar structure, then it's not really true anymore that the `type` is a property intrinsic only to the toolbar.

htmleditor-sample-external-plugin.js
```diff
@@ -0,0 +1,40 @@
...
+		<path fill="#494c4e" d="M9 0a9 9 0 1 0 9 9 9 9 0 0 0-9-9zm4.49 6.924l-5.02 5.51a.983.983 0 0 1-1.442 0l-2.48-2.482a.983.983 0 0 1 .008-1.417 1.027 1.027 0 0 1 1.4.02L7.712 10.3l4.3-4.73a1.018 1.018 0 0 1 1.4-.075 1.006 1.006 0 0 1 .078 1.43z"/>
+	</svg>`;
+
+class SamplePlugin {
+
+	get name() {
+		return 'sample-plugin';
+	}
+
+	get toolbarItems() {
```
grant-cleary, line 16: I wanted to go with the idea of exposing something for this on the UI registry... But the problem is that we sort of need these to be tied to the plugin. If the plugin were to just provide its name in the call, you could end up in a weird situation where PluginA could register toolbar items for PluginB.

One alternative I've been musing about is having the editor know about _commands_ for the in-repo plugins and do the sorting that way. Then this loading could still be plugin-agnostic and done through the registry.

components/toolbar/toolbar-builder.js
```diff
@@ -0,0 +1,23 @@
...
+	externalPlugins.forEach(plugin => {
+		if (!plugin.toolbarItems) return;
+
+		plugin.toolbarItems.forEach(item => {
+			if (!item.cmd) return;
+
+			extensibleToolbarItems.push({
+				cmd: item.cmd,
+				icon: item.icon,
+				text: item.text
```
grant-cleary, line 12: Note from offline conversation with Dave: Add support for type/extensionPoint/items properties to support current extension points (just `insert-more-menu`), menu item types (`menu-item-default`, `menu-item-toggle`, `menu-item-menu`), and submenu trees respectively.

components/toolbar/toolbar-builder.js
```diff
@@ -0,0 +1,23 @@
...
+			});
+		});
+	});
+
+	return extensibleToolbarItems;
+}
+
+export function buildToolbar(inline, externalPlugins) {
+	if (inline) return;
```
grant-cleary, line 21: This is probably too broad. While no current extensible plugin items load in the inline editor's `Insert More` menu, that's more coincidence because we only use full editors in places where that's relevant. We likely want this only to be restricted in the `inline-limited` case.

components/toolbar/toolbar-builder.js
```diff
@@ -0,0 +1,98 @@
...
+	const segmentItems = [];
+	items.forEach(item => {
+		if (!validateItem(item, external, submenu)) return;
+
+		switch (item.type) {
+			case menuItemTypes.DEFAULT:
+
+				segmentItems.push({
+					cmd: item.cmd,
+					extensionPoint: submenu ? undefined : item.extensionPoint,
```
grant-cleary, line 25: If we're in a submenu, it should be inferred that the extension point is the same as the parent.

components/toolbar/toolbar-mixin.js
```diff
@@ -346,6 +348,36 @@ export const ToolbarMixin = superclass => class extends Localizer(RtlMixin(super
 		`;
 	}
 
+	_renderExtensibleToolbarItems() {
+		if (!this.toolbarStructure) return '';
+		return this._renderToolbarItems(JSON.parse(this.toolbarStructure), true);
```
grant-cleary, line 353: For now we're kind of just ignoring the extension point because there's only one of them. When this gets extended to the rest of the toolbar, that's when internal/external extension points will come into play. As discussed offline, we do still want it in the explicit API though, so that we don't get any external plugin providers sending us nothing and forcing us to support a default configuration.

PR #1633, opened 2023-08-24
US156918 Implement foundation components for new generation workflow

mdgbayly:
### Summary of Changes

This is a work in progress suggestion for how we can refactor the existing practice generation implementation to make it easier to integrate Nick's workflow redesign, and represents some initial steps to decouple the GenAI implementation from the practices code.

There is probably more changes here than I was hoping for, as our original thought was to limit any major refactoring until after the initial EAP. However, as I worked through the ideas, the implementation I'm proposing here felt like the most natural way to set things up to support Nick's changes, while minimizing changes to the core workflow of actually integrating with the API and streaming the response back as changes to the practices.

This is based on the mocked up workflow in NIck's demo (private Slack link https://d2l.slack.com/archives/C05LACR6TPB/p1692820885047819)

Nothing is set in stone here so would value any feedback/alternative ideas.

Some of the major workflow changes include:
* Only using the full question text from the HTML editor as the source.
* No longer supporting features such as using the current selection, or headers in the source text to determine what should be used.
* Using `d2l-backdrop` to limit interactions during generation, in preference to disabling individual question fields.

I've also removed the GenAI right hand panel preview pane that was shown during generation, but we can put something similar back using the new implementation states.

Some of primary implementation goals were to:
* Encapsulate all Gen AI UI in our own web components to make it easier to include all the workflow changes and new feedback mechanisms in components that don't touch practices code.
* Decouple some of the more contentious static usages, such as exposing the tinymce editor. This is mostly avoided now because we only need the full raw text of the editor based on the workflow changes.
* We now get the question text by implementing a request provider in the practices dialog that allows us to expose the html editor's content using a closure over the current editor which avoids leaking it into the global namespace. Recognize that the implementation of this provider will likely change with the new plugin implementation as direct reference to the editor will likely not be possible, but I'm assuming the plugin API will provide similar ability to get at the editor contents.
* In general, minimize the number of references to LlmController from practices code
* Remove all Gen AI specfic styling and UI rendering from practices code.

While at the same time not changing the following core functions which are complex and we don't want to break right now prior to EAP:
* Reuse the existing `llm-controller.js/getLlmStream` implementation for now
* Reuse the per practice question `getLlmSuggestions` implementation for now

Our continued dependency on these per question `getLlmSuggestions` implementations does mean we need to continue exposing the practice question `shadowRoot` as a `LlmController` which is also not ideal but we have plans to remove that dependency in further refactoring that we'll do after our initial release.

The general structure of the components is as follows:

```
<d2l-practices-dialog>
      <d2l-question-generation-panel>
           <d2l-question-generation-editor> // This encapsulates the top panel from Nick's designs
           <d2l-question-generation-feedback>  // This encapsulates all the feedback controls from Nick's designs
```

I've mostly only worked on getting the basics of the `<d2l-question-generation-editor>` working for the following workflows
* Generate question
* Cancel generation
* Regenerate question
* Change question type

I've not paid any attention to styling yet.
I've also not implemented the dialog to edit the source text yet.

I have though removed a lot of references to LllmController HTML rendering/styes and workflow variables from the practices code including the dialog to edit the source text.

These are things that we will need to reimplement within the context of the redesign, but I think we can lift a lot of the existing code that I've removed to fast track the new implementation. in the context of the new components. e.g. error handling state, tracking min/max source text lengths, tracking counts in the source text dialog etc.

### Rally Story
[US156918](https://rally1.rallydev.com/#/706276043991d/custom/17898093484?detail=%2Fuserstory%2F720469302025) Implement foundation components for new generation workflow


JamesVaChao (approved):

components/practice/ai/generate-question-feedback.js
```diff
@@ -0,0 +1,39 @@
...
+		super();
+		this.api = new FeedbackController(this);
+	}
+
+	connectedCallback() {
+		super.connectedCallback();
+	}
+
+	render() {
+		return html`
```
JakeProuse, line 30: Nit: could return `nothing` here from lit

components/practice/ai/generate-question-editor.js
```diff
@@ -0,0 +1,185 @@
...
+		if (changedProperties.has('templateIndex')) {
+			this.api.templateIndex = this.templateIndex;
+		}
+	}
+
+	get _errorTemplate() {
+		if (this.api.errorState === State.Error) {
+			return html`
+				<div>
+					An error occurred
```
grant-cleary, line 84: I suspect this is already on your radar, but just a reminder that we'll need to localize all of these bits of rendered English text.

ianmclean2011: Yep, we'll do a thorough check/localization push 😄

JakeProuse (approved):

PR #1640, opened 2023-08-25
Fix append methods and add a method to wrap bare DOM nodes

grant-cleary:
Aligning `append` and `appendChild` APIs to behave like their browser alternatives, as well as adding a new API to wrap an existing DOM node in an `EditorNode` so it can be manipulated by other code in this area.


components/api-wrapper.js
```diff
@@ -72,6 +72,10 @@ export class InternalAPIWrapper {
 		return paragraphCount;
 	}
 
+	createEditorNode(domNode) {
+		return new EditorNode(domNode, this.#editor.selection);
```
grant-cleary, line 76: This could get weird if you tried to wrap a document or something, but ultimately I don't think there's much reason to block anything zany like that, as the underlying APIs should prevent shenanigans.

JakeProuse (approved):

PR #1646, opened 2023-08-28
US156938 launch edit source text popup

mdgbayly:
### Summary of Changes

Clicking "Generate" in the AI box launches a reimplementation of the Source Text Dialog from the original POC code.
The dialog is initialized with the plain text content representation of the current contents of the editor.

I was surprised that just calling `editor.getContent({format: 'text'})` doesn't provide a true plain text representation of the HTML from the editor, which I guess is why code had been added to the `LlmController` to do some conversions. Maybe the tinymce converter is just buggy? I saw some very old posts online describing similar issues but surprised it's not better now. But didn't want to get held up by that, so stuck with the POC code.

Clicking Generate Question in the dialog will kick off the generation process.

For now, for simplicity, I've used the native min/max character limits supported by the d2l textbox compont to set upper/lower limits on character counts. We have a follow up story to re-implement the character counting. The validation uses standard brightspaceui validation support tooltips. I've also made the field required so you cannot leave it empty.

There are various ways of launching dialogs using the d2l dialog components. Some options include using `opened` props and events, but in previous usages similar to this one I've found it clearer to use the promise based support of the dialogs to provide a more programatic functional calling mechanism. I just find it clearer than having disconnected props/events to open/close the dialog. It does mean you need to implement the promise based open mechanism in the dialog component but this is encapsulated away from users of the component.

### Rally Story
[US156](https://rally1.rallydev.com/#/706276043991d/custom/17898093484?detail=%2Fuserstory%2F719624512119) Launch edit prompt popup from PQ edit screen


grant-cleary:
TinyMCE's plain text problem is something that actually bit us recently. I think it's potentially a change in TinyMCE 6, but it looks like `format: text` now basically just behaves like calling `textContent` on the editor body. 😞

grant-cleary (approved):

components/practice/ai/generate-question-editor.js
```diff
@@ -159,7 +161,10 @@ class GenerateQuestionEditor extends RequesterMixin(Localizer(LitElement)) {
 		this.api.cancel();
 	}
 
-	_editSourceText() {
+	async _editSourceText() {
+
+		// We could have got this sooner e.g. when connected or first updated, but deferring to first use as it's
+		// quite possible the user won't use AI hence we wouldn't need to get the text
 		if (!this.api.originalSourceText) {
```
mdgbayly, line 168: One thing to note, is that we'll probably have to make this null check a bit more sophisticated, as I plan to re-use this method for the Edit Source Text button, and if the user hadn't entered any text in source HTML editor, this null check may cause their current text to be overwritten again, but I'm planning on working on that story next, so will fix up then.

ianmclean2011 (approved):

PR #1651, opened 2023-08-29
US156403 Passing state for editor

JamesVaChao:


components/practice/ai/generate-question-panel.js
```diff
@@ -60,10 +77,20 @@ class GenerateQuestionPanel extends RequesterMixin(Localizer(LitElement)) {
...
 		return nothing;
 	}
+
+	_updateEditorState(e) {
+		const { state } = e.detail;
+		if (state) {
+			this.editorState = state;
+		}
+		e.stopPropagation();
+		this.requestUpdate();
```
mdgbayly, line 91: nit: I think the reason we have to call `requestUpdate` here is because `this.editorState` is just a regular JS property rather than a lit property. If you added it to the properties declaration for the component, you could probably avoid having to call this manually.
For such "internal" properties, Lit supports a `state` flag on the property, which indicates its only used for managing internal state and cannot be set as an attribute on the element.
https://lit.dev/docs/components/properties/#internal-reactive-state

But not super important - how you have it is totally fine.

components/practice/ai/generate-question-panel.js
```diff
@@ -30,12 +32,25 @@ class GenerateQuestionPanel extends RequesterMixin(Localizer(LitElement)) {
...
 			}
+
+			#generate-question-panel {
+				background: linear-gradient(to right, #ceeafe, #f2f0ff);
+				border-radius: 0.4rem;
+				min-height: 2rem;
+				padding: 1rem;
+			}
+
+			#generate-question-panel.generating {
```
mdgbayly, line 43: Does this still work? The editor code is adding the `generating` class to the internal div in the editor component.
So I'm wondering if the backdrop works correctly now?

Were you thinking that we might want to apply the backdrop to the feedback controls too? Which is something we might need to consider. But possibly during generation we'd always hide the feedback controls? Was that the reason for moving the styles?

My primary reason for defining them internally was again with a future eye on maybe we'd get rid of the `generate-question-panel` component, and only have the `generate-question-editor`.

But if we do decide we want the backdrop to apply to the feedback controls too, then yes it might make sense to split up the styles so that the backdrop styling is handled here, but I think we could still leave the main blue gradient styling in the editor?

mdgbayly (approved):

grant-cleary (approved):

PR #1652, opened 2023-08-29
Rearrange test folders/structure for better organization

grant-cleary:
This PR looks nastier than it is because of how many visual-diff goldens have been moved around. The change has two components:
1. Move visual diff tests into their own subdirectory so they're separated from other tests.
2. Shift some common demo/test helpers into a `test-helpers` subdirectory so they're not just hanging around in the root.

These changes are intended to make it a bit easier to navigate the editor's tests in preparation for adding a boatload of new tests and a couple new helpers in PRs to come.


klhuff (approved):

grant-cleary:
Bah, there's that stupid bug that crops up now and again where some of the goldens end up just a blank page. Going to merge them to force visual-diff to regenerate them.

grant-cleary:
There we go. Don't ask me why kicking visual-diff again resolves that issue, but it seems to do so reliably. I wonder if it's the retries that break it? 🤔

ca2hill (approved):

PR #1654, opened 2023-08-30
US156507 AI Questions: Implement basic rating UI controls

mdgbayly:
### Summary of Changes

Builds on James's PR for US156403 which shares the editor state with the feedback component.

I added a shared stylesheet for the AI box styling which can be shared between the editor and feedback components.
This might be overkill. The other approach is to import the styles from one of the components into the other as described here.
https://daylight.d2l.dev/developing/best-practices/styles/#do-not%3A-separate-css-into-a-separate-file-for-no-reason
So I may revert this in a follow up PR.

Also tweaked where we are sending the event that the editor state is updated, because I wanted to incorporating the editor's "Generating" state to hide the feedback panel, so now the event is sent on any editor state change.

The feedback panel currently uses smiley emojis as we don't have the thumbs icons yet and they are not part of Daylight typography.

Also implemented a simpfiled version of the Undo Rating UI which doesn't include a Tell Us More button, as that will come in a later story.

Just included a placeholder UI for the Compare state (better, worse, same) as that will be handled in a follow up story.

Still need to hook up this UI to actually send a telemetry event, which we've now agreed will be done in the context of these stories rather than as distinct stories. So I'll work on that next.

When we get to error handling, I think we'll want to rationalize the editor error state to be just another regular editor state. I had initially kept them separate expecting us to want to show an error whilst still showing other editor state. But based on latest designs from Nick, we'll show Errors on their own. The updates to the feedback state included in this PR in response to the changes in the editor state assume that Error, will be one of the editor states, and will hide the feedback box.

### Rally Story
[US156407](https://rally1.rallydev.com/#/706276043991d/custom/17898093484?detail=%2Fuserstory%2F719624512191) Implement thumbs up/down controls in AI fb box


components/practice/ai/feedback-controller.js
```diff
@@ -1,6 +1,81 @@
...
+	get state() {
+		return this._state;
+	}
+
+	set state(value) {
+		this._prevState = this._state;
+		this._state = value;
+	}
+
+	compare() {
```
mdgbayly, line 32: `compare`/`more` are not used yet, but included as I was trying to get my head around all the different states.

Considered building this into some sort of simple state machine in code, but seemed overkill for this, because most all states only really have a single next state. Could still do this if we feel this code is already too complex.

components/practice/ai/generate-question-feedback.js
```diff
@@ -30,13 +38,59 @@ class GenerateQuestionFeedback extends RequesterMixin(Localizer(LitElement)) {
...
+	}
+
+	get _moreTemplate() {
+		if (this.api.state !== FeedbackState.More) {
+			return nothing;
+		}
+		return html`
+			<d2l-icon icon="tier1:wizard"></d2l-icon>
+			${this.localize('practice:llm:rate:thankyou')}
+			<d2l-button-subtle @click=${{ handleEvent: () => this.api.undoRating(), once: true }} text="${this.localize('practice:llm:rate:undo')}" ></d2l-button-subtle>
```
mdgbayly, line 77: Using these more elaborate event declarations which also allow you to specify the event will only be handled once.
To avoid double clicks - sending double telemetry.
Also helps to be able to pass a prop to the `api.rate(rating)` method for the rating button handlers because our lint rules prevent using a lambda function directly in a declarative event handler.

JamesVaChao (approved):

grant-cleary (approved):

PR #1658, opened 2023-08-30
Add tests to cover new node/selection/range wrappers

grant-cleary:
This PR adds tests that cover the behaviour of our new wrappers. Not too much to say here in terms of interesting behaviour. 😆


components/selection.js
```diff
@@ -146,10 +146,7 @@ export class EditorNode {
 
 	append(...nodes) {
 		for (const node of nodes) {
-			if (node instanceof EditorNode) {
-				this.appendChild(node);
-				node.remove();
-			}
+			if (node instanceof EditorNode) this.appendChild(node);
```
grant-cleary, line 149: When I was writing the tests I realized that the call to `node.remove()` here was redundant because `this.appendChild(node)` already does that. There's no harm in the second call, because in the browser, calling `remove()` on a node that's no longer in the DOM is just a no-op... But there's also no reason for us to keep the duplication.

test/editor-node.test.js
```diff
@@ -0,0 +1,846 @@
...
+			sandbox.stub(closestElement, 'nodeType').value(Node.DOCUMENT_NODE);
+			const node = new EditorNode(mockDomNode, mockSelection);
+
+			const closest = node.closest(testSelector);
+			sandbox.assert.calledOnceWithExactly(mockDomNode.closest, testSelector);
+			expect(closest).to.be.null;
+		});
+	});
+
+	describe('findComposedAncestor', () => {
```
grant-cleary, line 490: This test is kind of blech because it's not possible to stub out ES6 imports... So basically, the fact that we import `findComposedAncestor` from core means we have to kind of bend the test to make sure that call to core returns what we want for testing. This does mean that some core changes to `findComposedAncestor` could break this test, but it should be pretty obvious if that happens.

klhuff (approved):

ca2hill (approved):

PR #1660, opened 2023-08-30
DE44059 - Implement fullscreen exit via esc

PaulKrznaric:
Users can now use the `esc` key to exit fullscreen view for the HTML Editor. 

Full Screen Esc Jiff


components/event-registry.js
```diff
@@ -61,7 +61,7 @@ export class EventRegistry {
 			};
 		};
 
-		this.#editor.on('keydown', e => handler(trimEventDetails(e)));
+		this.#editor.on('keydown', e => handler(trimEventDetails(e), this.#editor.hasFocus()));
```
grant-cleary, line 64: I _think_ I would prefer to see a separate `internalEditorHasFocus()` API in the `APIWrapper` class that exposes the latter instead. It seems a bit odd to return it from the event handler. Then the workflow would be to call `helpers.internalAPIWrapper.internalEditorHasFocus()` within the event handler to replace this second param. I want to think about this a bit more though, so maybe don't make any changes quite yet.

components/plugins/fullscreen.js
```diff
@@ -15,30 +18,45 @@ export class FullscreenPlugin {
...
+			await helpers.internalAPIWrapper.toggleFullscreen(isFullscreen);
 			if (applyFocus) await helpers.internalAPIWrapper.setToolbarFocus('d2l-fullscreen');
 		};
 
-		const handleShortcutKey = e => {
-			if (e.keyCode !== 70) return;
-			if (!(e.metaKey && e.shiftKey) && !(e.ctrlKey && e.shiftKey)) return;
+		const handleShortcutKey = (e) => {
+			const isEditorFocused = helpers.internalAPIWrapper.internalEditorHasFocus();
+			const isValidEscEvent = (e.keyCode === escKey && isEditorFocused === true && isFullscreen === true);
```
grant-cleary, line 47:
```suggestion
			const isValidEscEvent = (e.keyCode === escKey && isEditorFocused && isFullscreen);
```

grant-cleary, line 47:
```suggestion
			const isValidEscEvent = (e.keyCode === escKey && isEditorFocused && isFullscreen);
```

components/plugins/fullscreen.js
```diff
@@ -15,30 +18,45 @@ export class FullscreenPlugin {
...
+			const isValidFullscreenShortcut = e.keyCode === fKey && (e.metaKey && e.shiftKey) || (e.ctrlKey && e.shiftKey);
+			if (!isValidEscEvent && !isValidFullscreenShortcut) return;
+
 			toggleFullscreen(false);
+			if (e.keyCode !== escKey || isFullscreen === true) handleFullscreenDismiss();
```
grant-cleary, line 52:
```suggestion
			if (e.keyCode !== escKey || isFullscreen) handleFullscreenDismiss();
```

grant-cleary (approved):

components/plugins/fullscreen.js
```diff
@@ -15,30 +18,45 @@ export class FullscreenPlugin {
...
 			if (applyFocus) await helpers.internalAPIWrapper.setToolbarFocus('d2l-fullscreen');
 		};
 
-		const handleShortcutKey = e => {
-			if (e.keyCode !== 70) return;
-			if (!(e.metaKey && e.shiftKey) && !(e.ctrlKey && e.shiftKey)) return;
+		const handleShortcutKey = (e) => {
+			const isEditorFocused = helpers.internalAPIWrapper.internalEditorHasFocus();
+			const isValidEscEvent = (e.keyCode === escKey && isEditorFocused && isFullscreen);
+			const isValidFullscreenShortcut = e.keyCode === fKey && (e.metaKey && e.shiftKey) || (e.ctrlKey && e.shiftKey);
```
klhuff, line 48: Should there be an extra pair of brackets around the pair of operators around the OR?

klhuff (approved):

components/plugins/fullscreen.js
grant-cleary: Good catch, yes. `&&` has a higher operator precedence than `||`.

PR #1661, opened 2023-08-30
Add test framework for testing plugin structure

grant-cleary:
On top of adding tests for our internal APIs, we also want to add a general testing framework that can be used to validate plugins, both internally and externally defined. The goal here is to make it general enough to be imported by anyone looking to create an external plugin, while still being usable for testing internal editor plugins.


grant-cleary:
FYI @JakeProuse, since I know you were interested in this.

test-helpers/plugin-helpers.js
```diff
@@ -0,0 +1,292 @@
...
+		return {
+			success: false,
+			message: 'Top-level toolbar items must specify an extension point'
+		};
+	}
+
+	if (external && itemExtensionPoint !== extensionPoint.INSERT_MORE_MENU) {
+		return {
+			success: false,
+			message: `External plugins can only register toolbar items for the '${extensionPoint.INSERT_MORE_MENU}' extension point`
```
JakeProuse, line 155: @grant-cleary veering away from calling it `insert stuff`? 😆

package.json
```diff
@@ -16,6 +16,7 @@
     "/components",
     "/generated",
     "/lang",
+	"/test-helpers/plugin-helpers.js",
```
JakeProuse, line 19: I think this just needs linted

JakeProuse (approved):

ca2hill (approved):

package.json
grant-cleary: Oh yeah, I keep forgetting that package.json uses spaces instead of tabs. I have GitHub's UI set to display them the same way so I don't notice.

test-helpers/plugin-helpers.js
grant-cleary: We have too many things labeled `Insert Stuff` already. We need to escape! 😅

PR #1665, opened 2023-08-31
US157083 Change behaviour of cancel button to cancel immediately

JamesVaChao:
Removed cancelStream to transition to managing things with controllers. Cancel will go back to previous state properly

Story:
[US157083](https://rally1.rallydev.com/#/?detail=/userstory/720787475709&fdp=true): Change behaviour of cancel button to cancel immediately


components/practice/llm-controller.js
```diff
@@ -128,7 +128,7 @@ export class LlmController {
 					<div class="d2l-question-summary-loading"><d2l-loading-spinner></d2l-loading-spinner></div>
 					<div class="d2l-question-summary-loading-text"><span>${LlmController.loadingMessage}</span></div>
 					<div class="d2l-question-summary-cancel-button">
-						<d2l-button-subtle ?disabled=${LlmController.cancelled} @click="${LlmController.cancelStream}" text="${localizations.cancel}"></d2l-button-subtle>
+						<d2l-button-subtle ?disabled=${LlmController.cancelled} text="${localizations.cancel}"></d2l-button-subtle>
```
ianmclean2011, line 131: I think you'll need to put back the click functionality to point it at the cancel function (maybe that's coming in another PR?).

JamesVaChao: I think we are transitioning from using html - templates from llm-controller. I just removed all references go cancelStream for eslint.

components/practice/ai/editor-controller.js
```diff
@@ -50,7 +50,9 @@ export class EditorController {
 	}
 
 	cancel() {
-		LlmController.cancelStream();
+		LlmController.generated = false;
+		LlmController.cancelled = true;
+		this._determineNextState();
```
mdgbayly, line 55: Does this cause the UI go switch immediately back to the "Generate/Regenerate" state?

Is there a danger here that by not waiting for the LlmController stream to finish that we could allow the user to click Generate again, whilst the LlmController is still finishing processing the previous request?

mdgbayly (approved):
Just my concern expressed in the comment about whether this can lead to race conditions with the LLM controller not being finished if user clicks Generate again. If we think that's not an issue then good to merge.

components/practice/ai/editor-controller.js
JamesVaChao: So I tried to do this but I think each stream is immediately discarded so I've never ran into this issue while spaming the cancel/regenerate.

JamesVaChao: We could potentially fix this once each run has a generation-id and we can cancel specific generation-ids to make sure two requests don't overlap

grant-cleary (approved):

components/practice/llm-controller.js
ianmclean2011: Ohhhhhhhhhhhh ... right ... the llm controller stuff still throws me off sometimes 😄

PR #1667, opened 2023-09-01
US156407 AI Questions TA178870 Implement basic thumbsup/down/undo rating telemetry

mdgbayly:
### Summary of Changes

Added telemetry events for ThumbsUp, ThumbsDown and UndoRating.

Prior to sending a rating telemetry event it will send a QuestionGenerated telemetry event containing the associated source text and generated question, if one has not been sent since the last feedback event.

I chose not to use the existing telemetry "plugin" being used for the practices as that was opinionated for the way practices is using telemetry and I wanted to use telemetry with different fields.

I also took advantage of the default telemetry client setup by BSI on `window.D2L.Telemetry`

The basic idea is that when the user clicks thumbs down/up/undo we will send a telemetry event to indicate the action was taken. If we have not already sent a `QuestionGenerated` event for the current question generation, we will send one first.

The format of the events mostly follows the proposal I created [ here ](https://desire2learn.atlassian.net/wiki/spaces/HAI/pages/3756786929/WIP+User+Feedback+Telemetry) with some tweaks to use existing StandardEvent fields rather than custom properties where possible to make it easier to query this data in BSI/Athena.

I did have to modify the LLM stream processing in each question component to pass back the questionType and the full contents of the streamed generated question so that this can be included in the relevant events. To capture the full contents of the generated question I put back some code that was previously concatenating the chunks during streaming.

The implementation tracks how many times the user generated a question, and sends that with the events. It also sends a `SessionId` which is just a UUID initialized when the create practice page loads which can be used to tie together different events from the session.

I'll document the full details of the events, as implemented, next week.

I chose to implement all the telemetry handling in a new class managed by the top level `<generate-question-panel>` component. Then the editor and feedback panels send regular browser events which are handled at this top level component to co-ordinate the sending of telemetry. This seemed simpler than trying to send telemetry messages from the editor/feedback components as that would have required sending data up then down the tree. e.g. to get `GenerationId` info when sending feedback events.

Note: In order to use the standard `Activity` field for capturing the SessionId, I had to extend the `d2l-telemetry-browser-client` to support that field as it wasn't currently supported by the client. The changes were already merged and deployed so it should be automatically be picked up by BSI builds, but in case you see issues this is the [PR](https://github.com/Brightspace/d2l-telemetry-browser-client/pull/49).

I also added a new `SourceId` `generative-ai` to the `d2l-telemetry-service`. This has also been merged and deployed to Quad sites, so these events should be being sent without error but just in case this is the associated [PR](https://github.com/Brightspace/d2l-telemetry-service/pull/147).

### Rally Story
[US156407](https://rally1.rallydev.com/#/706276043991d/custom/17898093484?detail=%2Fuserstory%2F719624512191) Implement thumbs up/down controls in AI fb box


components/practice/ai/editor-controller.js
```diff
@@ -76,6 +81,11 @@ export class EditorController {
 	}
 
 	async generate() {
+		const element = LlmController._getTemplateElement(this.templateIndex, LlmController.shadowRoot);
```
mdgbayly, line 84: Decided to return early here if we can't find the element.
Previously I had it continuing with state management even if the element was not found which didn't make much sense.

components/practice/ai/generate-question-editor.js
```diff
@@ -202,19 +202,9 @@ class GenerateQuestionEditor extends RequesterMixin(Localizer(LitElement)) {
 		}
 	}
 
-	async _generate() {
-		await this.api.generate();
-		const event = new CustomEvent('d2l-update-editor-state', {
```
mdgbayly, line 207: I thought I'd removed this because it is now sent in the controller's `set state()` method, but maybe I forgot or it got merged back in.

ianmclean2011 (approved):
Looks good!

components/practice/ai/generate-question-feedback.js
```diff
@@ -43,9 +60,18 @@ class GenerateQuestionFeedback extends RequesterMixin(Localizer(LitElement)) {
 		}
 		return html`
 			<div id="feedback-panel" class="ai-panel d2l-body-compact">
-				${this._rateTemplate}
-				${this._compareTemplate}
-				${this._moreTemplate}
+				<div id="sparkle-container">
+					<d2l-icon-custom size="tier2">
+						<svg width="24" height="25" viewBox="0 0 24 25" fill="none" xmlns="http://www.w3.org/2000/svg">
```
ianmclean2011, line 65: @grant-cleary Just wanted to call out that these are temporary SVGs. We're still working through the finalized designs for them, and looking at getting them properly added to the standard Daylight set.

JamesVaChao (approved):

grant-cleary (approved):

grant-cleary:
@JamesVaChao, just a reminder that your merge commit messages (or the commit messages within the changeset) should follow our semantic release rules to ensure the editor's release notes stay clean and unrelated changes aren't inappropriately bundled together: https://github.com/BrightspaceUI/htmleditor#versioning--releasing

One caveat is that you should never trigger a major release.

PR #1668, opened 2023-09-01
US156409 Implement better worse same controls with associated telemetry

mdgbayly:
### Summary of Changes

Implement better/worse/same controls for a regenerated question with associated telemetry.

Integrated Nick's temp svgs.
Went with subtle buttons with labels with the svg as a slotted element, but doesn't really match Nick's design, but I'm not sure we can. It might be better to spend the time adding these svgs to the standard Daylight icon set rather than us trying to hack around the limitations.

Tried to implement the hi-fidelity designs from Figma but it is challenging given the svgs, and the fact that Nick's design's don't really take into account that we are using standard Daylight buttons.

### Rally Story
[US1564079](https://rally1.rallydev.com/#/706276043991d/custom/17898093484?detail=%2Fuserstory%2F719624512207) Implement better/worse controls in AI fb box


ianmclean2011 (approved):
Looks good. Yes, we'll need to get the SVGs into the standard daylight set, since these will be used elsewhere eventually (I imagine).

PR #1670, opened 2023-09-01
Add additional internal APIs for C+ plugins

grant-cleary:
This adds a couple additional APIs to the editor to support some work migrating over the C+ plugins to the new plugin framework.
1. Allow stylesheets to be injected temporarily into TinyMCE's content document (will not apply on save).
2. Expose information on whether the `insert-element` dialog is open, so `layouts` can use this to determine whether it's permitted.


components/plugin-registry.js
```diff
@@ -71,6 +71,7 @@ export class PluginRegistry {
 			contentStyler: this.#component.contentStyler,
 			fullPage: this.#component.fullPage,
 			insertElement: this.#component.insertElement,
+			insertElementDialogOpen: requestInstance(this.#component, 'd2l-insert-element-dialog-open'),
```
grant-cleary, line 74: I don't _love_ this because it means the editor now knows about this magic thing it might need to request. I think ideally we would find a better way to do this (or jettison the check entirely), but I think that requires some more thought. In the meantime, this should do the job.

components/api-wrapper.js
```diff
@@ -33,6 +33,27 @@ export class InternalAPIWrapper {
...
+			if (absoluteStylesheetUrl === absoluteUrl) {
+				return;
+			}
+		}
+
+		const elem = editorDoc.createElement('link');
+		elem.rel = 'stylesheet';
+		elem.href = url;
+
+		editorDoc.head.appendChild(elem);
```
grant-cleary, line 54: I didn't see a reason to apply attributes here, since we're not using this to inject any content into the document being edited.

JakeProuse (approved):

PR #1671, opened 2023-09-02
Add shared data cache so plugins can communicate with embedded editors

grant-cleary:
Quick prototype of what a plugin-agnostic cross-editor data cache could look like. The idea here is to use providers/requesters to allow plugins from one editor to talk to plugins in an embedded editor. This would allow Insert Element, for example, to signal to its embedded editor that Layouts should be allowed to function despite not being in `fullPage` mode.


htmleditor-mixin.js
```diff
@@ -528,6 +528,13 @@ export const HtmlEditorMixin = superclass => class extends Localizer(LabelledMix
 		if (!this._contextMenuPlugins.has(plugin.name)) this._contextMenuPlugins.set(plugin.name, new plugin());
 	}
 
+	_addToSharedDataCache(key, value) {
+		if (typeof key !== 'string') return;
+		if (typeof value !== 'string' && typeof value !== 'number' && typeof value !== 'boolean') return;
+
+		this.provideInstance(this._createSharedDataCacheKey(key), value);
```
grant-cleary, line 535: I could use a map for this locally, but the implementation of `provideInstance/requestInstance` already builds a map on the editor component, so it feels kind of redundant.

grant-cleary:
Eh, I tried to do this with GitHub's built-in code editor, but it didn't work so well because it didn't do linting. That might be a configuration error, but given that I was just trying to cram something in quickly so I don't forget it next week, it makes more sense for me to just leave it until Tuesday when I can make changes properly in my desktop IDE. 😆

components/plugin-registry.js
```diff
@@ -23,7 +23,6 @@ import { UiRegistry } from './ui-registry.js';
 import { WordCountPlugin } from './plugins/wordcount.js';
 
 const privilegedDefaultPlugins = [
-	'd2l-fullpage',
```
grant-cleary, line 26: This isn't a plugin anymore; it was just left in here by accident.

JakeProuse (approved):

PR #1680, opened 2023-09-06
Remove TinyMCE version/image enhancement flag

grant-cleary:
The editor is going to keep references to the `useNewTinyMCEVersion` context object, since this is going to be reused in the future when we wish to make TinyMCE updates. However, the image enhancement stuff that was attached to this flag can now be made permanent and removed from the `useNewTinyMCEVersion` check.


htmleditor-mixin.js
```diff
@@ -755,7 +755,7 @@ export const HtmlEditorMixin = superclass => class extends Localizer(LabelledMix
 			a11ychecker_issue_url_callback: ruleId => a11yCheckerRuleUrlMap[ruleId],
 			allow_html_in_named_anchor: true,
 			allow_script_urls: true,
-			automatic_uploads: !this.pasteLocalImages && (!this._context || !this._context.useNewTinyMCEVersion),
+			automatic_uploads: !this.pasteLocalImages,
```
klhuff, line 758: We don't still want the context check here? If useNewTinyMCEVersion is just "true" now, that makes this something || false.

grant-cleary: The context check only existed because `this._context.useNewTinyMCEVersion` would throw if it didn't, and we still support versions of Safari that don't support optional chaining. If we're not checking the property, we don't actually care if the context object exists.

klhuff (approved):

ca2hill (approved):

PR #1682, opened 2023-09-06
Centralize toolbar fetching when applying focus and clean up tests

grant-cleary:
Just a couple tweaks to the toolbar focus API to allow it to work for all editor types. Also just tweaking some tests to account for new return types.


components/api-wrapper.js
```diff
@@ -127,8 +127,8 @@ export class InternalAPIWrapper {
 	}
 
 	focusToolbar() {
-		const contextMenu = this.#component.shadowRoot.querySelector('d2l-htmleditor-toolbar-full');
-		if (contextMenu) contextMenu.focus();
+		const toolbar = this.#component.getToolbar();
```
JakeProuse, line 130: Thanks for catching my variable name mixup!

JakeProuse (approved):

components/api-wrapper.js
grant-cleary: I also missed it in the review. I only noticed when I went to paste in the method signature when making this change. 😆

PR #1683, opened 2023-09-06
Jprouse/us156137 convert insert element plugin to use new api

JakeProuse:
Rally Story: https://rally1.rallydev.com/#/?detail=/userstory/719095594091&fdp=true


components/plugins/insert-element.js
```diff
@@ -33,11 +34,113 @@ import { RtlMixin } from '@brightspace-ui/core/mixins/rtl-mixin.js';
...
+		});
+
+		helpers.commandRegistry.registerCommand(cmds.insertElementDelete, () => {
+			const selection = helpers.internalAPIWrapper.getSelection();
+			const node = selection ? selection.getNode() : null;
+
+			if (!node) return;
+
+			node.remove();
+			helpers.internalAPIWrapper.insertContent('');
```
grant-cleary, line 127: I assume this is here to trigger the `BeforeSet/SetContent` handlers?

JakeProuse: Yes that's correct!

JakeProuse: Are there other ways of doing this rather than an empty insert?

grant-cleary: Chatted offline, but in case anyone else looks at this... No, there isn't really a better way. If this became a more common pattern, TinyMCE has APIs for firing events that we could expose, but it doesn't seem that much more intuitive, since you would still need to know you need to force the events to fire to accomplish what you were after.

PR #1686, opened 2023-09-06
US156397 AI Questions: Style feedback panel

mdgbayly:
### Summary of changes

Apply designs for Feedback panel based on Nick's Figma designs.

I also switched to using the correct Daylight button controls with placeholder icons for the thumbs/equals. I think these are possibly even acceptable for EAP if gaudi don't manage to add the real SVGs in time.

Refactored the rendering of the 3 different feedback states to use a common template parameterized with the correct buttons/lang terms. This template shares some of the same css style ideas as the main AI box with the generation controls, but there were enough differences in the design requirements that it didn't make sense to move them into a shared stylesheet.

### Rally Story ###
https://rally1.rallydev.com/#/706276043991d/custom/17898093484?detail=%2Fuserstory%2F719624512111


components/practice/ai/ai-styles.js
```diff
@@ -7,3 +8,13 @@ export const aiStyles = css`
 		padding: 1.2rem;
 	}
 `;
+
+export function sparkleIcon() {
```
mdgbayly, line 12: I had thought to use this function in the `generate-question-editor` too. But we probably won't need it once we have a real sparkle icon. Considered just switching to a regular `d2l-icon` with a placeholder icon for now too which would also make this obsolete but decided that might be a step too far in case we don't get it...so not sweating it for now.

ianmclean2011 (approved):
Looks good!

JamesVaChao (approved):

grant-cleary (approved):

PR #1687, opened 2023-09-07
US156406 AI Questions: Style source text dialog

mdgbayly:
### Summary of Changes

Seeing changes from previous branch merged into this one too.
Just the last commit applies to this PR. Will sort out tomorrow.

Updated source text dialog based on Nick's Figma design.

Also tried to set text area to have unlimited rows to try and avoid issues with double scroll bars.
But I'm now seeing more frequent weirdness where the text area shows an unnecessary vertical scroll bar. Will follow up with additional changes if necessary.

### Rally story

https://rally1.rallydev.com/#/706276043991d/custom/17898093484?detail=%2Fuserstory%2F719624512183


components/practice/ai/generate-question-source.js
```diff
@@ -52,23 +65,28 @@ class GenerateQuestionSource extends RtlMixin(Localizer(LitElement)) {
...
-						<div id="summary-text-bullets">
-							<div><d2l-icon icon="tier1:check-circle"></d2l-icon>${this.localize('practice:llm:editSource:line1')}</div>
-							<div><d2l-icon icon="tier1:check-circle"></d2l-icon>${this.localize('practice:llm:editSource:line2')}</div>
-							<div><d2l-icon icon="tier1:check-circle"></d2l-icon>${this.localize('practice:llm:editSource:line3')}</div>
-							<div><d2l-icon icon="tier1:check-circle"></d2l-icon>${this.localize('practice:llm:editSource:line4', { min: formatNumber(sourceTextLimitMin), max: formatNumber(sourceTextLimitMax) })}</div>
+					<div class="dialog-content">
+						<div id="summary-text" class="ai-panel, d2l-body-standard">
+							${sparkleIcon()}
+							<div>
+								<div>${this.localize('practice:llm:editSource:title')} <a href="https://d2l.com" target="_blank">${this.localize('practice:llm:editSource:ai-link')}</a></div>
```
ianmclean2011, line 72: Should this be treated as one string using the [localization link stuff](https://github.com/BrightspaceUI/core/tree/3ae22dd59a827683687dcacd5f821602ae1353ef/mixins/localize#links)?

JamesVaChao (approved):

components/practice/ai/generate-question-source.js
mdgbayly: I don't think so. It's just a link as an additional sentence. So it's not connected to the previous sentence.

mdgbayly: But maybe we should be using the `d2l-link` component. I didn't realize we had one. But I don't think it's super important.

ianmclean2011 (approved):

grant-cleary (approved):

PR #1689, opened 2023-09-07
Us157007 Hide ai box for unsupported questions

JamesVaChao:


components/practice/ai/generate-question-panel.js
```diff
@@ -54,6 +54,8 @@ class GenerateQuestionPanel extends RequesterMixin(Localizer(LitElement)) {
 	}
 
 	render() {
+		this.api.determineHidden();
```
mdgbayly, line 57: It's not a huge deal, but it's not ideal to be checking and changing state like this in the `render` method as the render method can be called a lot.

A better pattern for reacting to changes in the `templateIndex` is to use the `willUpdate` lifecyle method like we use in the `generate-question-editor`, and then update the controller state (for example by calling this `determineHidden()` method from there).

components/practice/ai/panel-controller.js
```diff
@@ -15,12 +16,21 @@ export class PanelController {
 		return false;
 	}
 
+	determineHidden() {
+		this.host.hidden = true;
+
+		const context = requestInstance(this.host, 'context');
+		if (context && context.generativeAIEndpoint) {
+			if (VALID_TEMPLATE_ID.has(this.host.__templateIndex)) {
```
mdgbayly, line 24: You should be able to just reference `this.host.templateIndex` here, as it's fine for the controller to reference properties of the host directly, rather than accessing the internal property implementation.

mdgbayly (commented):
As discussed we should just use the standard `supporting` slot for the icon and accept that it has to be right aligned for now.

Plus just a couple of other more nitpicky comments on how we're hiding the panel.

components/practice/ai/panel-controller.js
JamesVaChao: done

components/practice/ai/generate-question-panel.js
JamesVaChao: done

components/practice/ai/generate-question-panel.js
```diff
@@ -74,6 +74,10 @@ class GenerateQuestionPanel extends RequesterMixin(Localizer(LitElement)) {
 		super.updated(changedProperties);
 	}
 
+	willUpdate() {
+		this.api.determineHidden();
```
mdgbayly, line 78: This could probably just check to see if the `templateIndex` property has changed to make it more explicit as that is what we are interested in here. I think that should also cover the case where the editor is not configured, given it's also checking the endpoint being defined and the `templateIndex` should also be updated on first render. But no big deal given the only property on this component is the `templateIndex`.

mdgbayly (approved):
LGTM

grant-cleary (approved):

PR #1691, opened 2023-09-07
US156404 Discard button resets preview pane

JamesVaChao:
Discard button now resets preview pane properly

Related story:
[US156404](https://rally1.rallydev.com/#/?detail=/userstory/719624512167&fdp=true): Implement Discard button in AI box


components/plugins/practice.js
```diff
@@ -114,6 +114,12 @@ export function addPracticePlugin() {
 				dialog.removeEventListener('d2l-dialog-save', handleSave);
 			}, { once: true });
 
+			dialog.addEventListener('d2l-genai-reset-preview-pane', (e) => {
```
mdgbayly, line 117: nit: this could be added as a declarative event handler within the dialog component on the `d2l-generate-question-panel`, that way it isn't accessing internal dialog state and keeps the event closer to the component/target.

components/practice/ai/editor-controller.js
```diff
@@ -78,7 +78,15 @@ export class EditorController {
 		this._sourceTextInitialized = false;
 		this.initSourceText();
 
-		this.host.requestUpdate();
+		const event = new CustomEvent('d2l-genai-reset-preview-pane', {
```
mdgbayly, line 81: nit: this could be implemented as a more generic event like `d2l-genai-discard` so that user's of the component can use it more abstractly.

mdgbayly (approved):
Looks good - just a couple of small nits, but no need to fix unless you want to.

ianmclean2011 (approved):

grant-cleary (approved):

PR #1692, opened 2023-09-07
Make sure scripts aren't persistent within the HTML document

grant-cleary:
Script references behave a bit differently from stylesheets and aren't included in the editor's iframe. That means we need to react a bit differently when trying to decide whether to remove obsolete script references. Effectively, if we don't find a script reference in the HTML document we're processing, we _must_ assume it's obsolete and needs to be removed. This does put the onus on plugin authors to ensure that their scripts are always being re-added if manual deletion occurs and isn't an intended workflow.


grant-cleary:
@JakeProuse, I have some minor concerns that this could impact the C+ plugins. Would you mind taking a quick look-over and letting me know? I _think_ the C+ plugins all try to re-add their scripts on `SetContent` so it shouldn't be a concern, but I can't really be certain I haven't missed something there.

JakeProuse:
> @JakeProuse, I have some minor concerns that this could impact the C+ plugins. Would you mind taking a quick look-over and letting me know? I _think_ the C+ plugins all try to re-add their scripts on `SetContent` so it shouldn't be a concern, but I can't really be certain I haven't missed something there.

The content styler plugin uses BeforeSetContent, but the others use SetContent and reset their scripts each time if they're not there. Would Content Styler be affected by this change?

grant-cleary:
> The content styler plugin uses BeforeSetContent, but the others use SetContent and reset their scripts each time if they're not there. Would Content Styler be affected by this change?

I tested Content Styler (since it's the only C+ plugin that I can get working in my demo) and it worked okay. While it does use `BeforeSetContent`, it does re-add its scripts every time. The major concern would be if there were any plugins that set their scripts once (like on plugin initialization or something) and then didn't re-add them either on `SetContent` or `BeforeSetContent`. I think this would break if any of the existing plugins were still doing that on `GetContent`, but I believe we removed the last usage of that a while ago.

PaulKrznaric (approved):

JakeProuse:
Okay Content Styler should be good then, we'll catch anything that comes up in QA anyhow. Realizing now looking at it, Practices still uses a GetContent hook, but it seems to be doing adding the same scripts/style that the SetContent hook does + doing some iframe attribute setting. I'm guessing it'd be okay seeing as this change just affects scripts, not iframe stuff, but @jschlumkoski can double check my work here 😄

components/plugins/fullpage.js
```diff
@@ -247,16 +247,17 @@ export class FullPageState {
...
 			// If a ref has been removed from the HTML markup (say, via the source code editor)
 			// and is still referenced in TinyMCE's iframe, wipe it from our ref map and remove
-			// it from the editor iframe.
-			if (removeObsolete && existingEditorRef) {
+			// it from the editor iframe. Script refs are never included in TinyMCE's iframe, so
+			// we just remove those references regardless.
+			if (removeObsolete && (existingEditorRef || type === refTypes.SCRIPT)) {
 				refMap.delete(url);
-				if (updateInDoc) existingEditorRef.remove();
+				if (updateInEditor) existingEditorRef.remove();
```
klhuff, line 260: Was this just a typo before? It's hard for me to tell what changing the bool being checked here does.

grant-cleary: I have a commit en route already, but this was... Sort of a typo? Before this change, `updateInDoc` was functionally the same check as `updateEditorDoc`, at least as far as any of this code mattered. However, `updateEditorDoc` is the check that's supposed to be here. So I fixed a typo with another typo.

components/plugins/fullpage.js
```diff
@@ -247,16 +249,17 @@ export class FullPageState {
...
 			// If a ref has been removed from the HTML markup (say, via the source code editor)
 			// and is still referenced in TinyMCE's iframe, wipe it from our ref map and remove
-			// it from the editor iframe.
-			if (removeObsolete && existingEditorRef) {
+			// it from the editor iframe. Script refs are never included in TinyMCE's iframe, so
+			// we just remove those references regardless.
+			if (removeObsolete && (existingEditorRef || type === refTypes.SCRIPT)) {
 				refMap.delete(url);
-				if (updateInDoc) existingEditorRef.remove();
+				if (updateEditorDoc) existingEditorRef.remove();
```
grant-cleary, line 262: `updateEditorDoc` is never true when the ref is a script. For stylesheets, `updateEditorDoc` is always true when `removeObsolete` is true as well. `updateInDoc` happened to function properly before because for stylesheets specifically, the checks were equivalent.

klhuff (approved):

joshhoey (approved):

components/plugins/fullpage.js
grant-cleary: With the refactor work, it's going to be possible to clean this up somewhat to make it a bit easier to deal with, since we'll no longer have the same requirement for persisting scripts/stylesheets (it's expected that plugins will handle that themselves going forward). However, I don't really want to tackle that right now.

PR #1693, opened 2023-09-07
Clean up get/set content APIs to properly maintain fullpage formatting (if applicable) and allow math to properly convert/render

grant-cleary:
There were some inconsistencies in the get/set content APIs that could allow plugins to pretty badly mangle the HTML being passed around if `fullPage` was enabled. This was also the case prior to the refactor, but now that we have this logic centralized, we can actually stop it.

So there are two main changes here:
1. Add functionality to the get/set handlers to make sure we're unwrapping the document element and head prior to passing it off to the plugin event handlers, and then rewrapping it once we're done.
2. Move the `setContent` API to be an exclusively privileged function, since it allows rewriting the entire document. This does require making the Source Code and Attributes plugins privileged.
3. Fix an issue in the bookmarking logic that was causing bookmarks to fail to save, which prevented existing equations from being edited.


components/event-registry.js
```diff
@@ -7,17 +9,42 @@ export class EventRegistry {
 		this.#editor = editor;
 	}
 
+	#extractBodyContents(html) {
```
PaulKrznaric, line 12: What do the `#`s actually represent on properties?

PaulKrznaric (approved):

joshhoey (approved):

components/event-registry.js
grant-cleary: It means that these are _actual_ private properties. You'll get an exception thrown in modern browser versions if you try to call these from outside of the class.

PR #1699, opened 2023-09-08
fix: show alert if question generation fails

SheranWiratungaD2L:
Rally story: https://rally1.rallydev.com/#/706276043991ud/custom/17898093484?detail=%2Fuserstory%2F720280279363

Testing:
- when first attempt to generate question fails
  - alert is shown, and AI box is hidden
  - clicking "Try again" opens edit text dialog
  - closing alert shows initial screen
  - edits to source text are preserved
- when attempt to regenerate question fails
  - same as when first attempt fails
  - feedback box is hidden


lang/en.js
```diff
@@ -450,7 +450,7 @@ export default {
 	"practice:llm:editSource:line4": "Source text must be in English between {min}-{max} characters.",
 	"practice:llm:editSource:label": "Edit Source Text",
 
-	"practice:llm:generated:error": "An error has occurred",
+	"practice:llm:generated:error": "The question failed to generate, please try again.",
```
ianmclean2011, line 453: Do we have any specific error messages in the pipeline? Not something that has to change right now, but just wondering if this error key should be more specific than `error`.

JamesVaChao (approved):

ianmclean2011 (approved):
Looks good!

lang/en.js
SheranWiratungaD2L: That should be covered in [US156919](https://rally1.rallydev.com/#/?detail=/userstory/720469730259&fdp=true): Review/refine error handling, I think. Right now we don't distinguish between different errors.

mdgbayly (approved):
LGTM

grant-cleary (approved):

PR #1700, opened 2023-09-08
Fix word count to not always use selection

grant-cleary:
We were always fetching content as though it were on a selection object, so let's fix that.


components/plugins/wordcount.js
```diff
@@ -37,7 +37,7 @@ export class WordCountPlugin {
 
 		const dialogFactory = dialog => {
 			return new Promise(resolve => {
-				const text = helpers.internalAPIWrapper.getContent(true, true);
+				const text = helpers.internalAPIWrapper.getContent(false, true, true);
```
grant-cleary, line 40: I don't love the boolean params, so I may split this out at some later date into a method on the selection object itself, along with the generic one here. I could also replace the second and third params with an opts object param, but I'm a bit less sure about that one, since it does make the API less straightforward.

PaulKrznaric (approved):

PR #1706, opened 2023-09-08
US156408 AI Questions: Implement Feedback dialogs and associated telemetry

mdgbayly:
### Summary of Changes

Added a new **Tell Us More** bottom which is displayed after a user gives a rating.
Clicking it launches a Feedback dialog configured with appropriate title and checkboxes based on the type of rating given.
Cancelling from the feedback dialog leaves state unchanged.
Submitting the feedback puts feedback back to initial state until user regenerates a new version of the question.

Also included a fallback for `crypto.RandomUUID()` in case some Retro browsers don't support it. This is only used for sending UUIDs so we can correlate telemetry events, so not critical.

Included `LMSVersion` in telemetry events which is pulled from the page's `html` element.

Also fixed a bug I noticed with Discard where the call to host.requestUpdate had been removed. Assuming this was a mistake. It was prevent the UI from resetting state correctly following a Discard.

### Rally Story

https://rally1.rallydev.com/#/706276043991d/dashboard?detail=%2Fuserstory%2F719624512199


components/practice/ai/telemetry.js
```diff
@@ -21,7 +21,7 @@ export async function sendTelemetryEvent(sourceId, action, context, activity, ob
 		// Create the base of the event body.
 		const eventBody = new d2lTelemetryBrowserClient.EventBody()
 			.setAction(action)
-			.setContext(context.orgUnitId)
+			.setContext(context.orgUnitId, 'unknown')
```
mdgbayly, line 24: The TelemetryEvent schema is based on standard events, and the `Type` attribute of the `Context` property is required. It doesn't prevent events being sent but it does generate logged warnings in the Telemetry Service which only recently started validating against the schema.

I don't see an obvious way to get the OrgUnitType in the html editor currently without us doing the work to pass it in as part of the context, so just setting to `unknown` for now so we pass schema validation. This should not affect BI usage of this data.

JamesVaChao (approved):

grant-cleary (approved):

PR #1708, opened 2023-09-11
Us157252 Imposing consumption limitations

JamesVaChao:
When consumption limit is reached the backend sends a 429 error.
pr parses it out and sets the error type to be a consumptionLimit error
Displays the consumptionLimit message and stops user from further generating.
Note that on switching practice type, we do not persist this error as it makes resetting state more complicated. The user will find out that they can't circumvent this consumption limit error.

Testing:
[screen-capture.webm](https://github.com/BrightspaceUI/htmleditor/assets/32684226/83468c87-e111-49ca-b38f-eac26bda8923)

Related story:
[US157252](https://rally1.rallydev.com/#/706276043991d/detail/userstory/721015255149): Imposing Consumption limitations (Rate Limiting) in UI


components/practice/llm-controller.js
```diff
@@ -306,6 +307,11 @@ export class LlmController {
 						LlmController.errorSuggestion = LlmController.localizations.contactAdmin;
 						LlmController.error = true;
 					// If user provided source text that isn't relevant for the question type
+					} else if (response.status === 429 && chunkValue.includes('USAGE_LIMIT_EXCEEDED')) {
+						LlmController.errorText = LlmController.localizations.consumptionLimit;
+						LlmController.errorSuggestion = LlmController.localizations.contactAdmin;
+						LlmController.error = true;
+						LlmController.errorType = ErrorTypes.ConsumptionLimit;
```
JamesVaChao, line 314: Will probably need to be redone once we remove the statics and control most of this not in the llm controller

grant-cleary (approved):

PR #1714, opened 2023-09-12
fix: making sure the feedback box is hidden for all question generati…

SheranWiratungaD2L:
…on failure cases


components/practice/ai/editor-controller.js
```diff
@@ -128,7 +128,7 @@ export class EditorController {
 
 		LlmController.clearFields(this.templateIndex, LlmController.shadowRoot);
 
-		let generationResult = {};
+		let generationResult = undefined;
```
SheranWiratungaD2L, line 131: Changed this so `_determineNextState` function doesn't think generation was successful and change the state.

mdgbayly (approved):

ianmclean2011 (approved):
Looks good!

grant-cleary (approved):

PR #1718, opened 2023-09-13
Jprouse/us156138 convert practices plugin to use new api

JakeProuse:
Rally Story: https://rally1.rallydev.com/#/?detail=/userstory/719095594941&fdp=true


components/plugins/practices.js
```diff
@@ -41,21 +41,190 @@ import { inputLabelStyles } from '@brightspace-ui/core/components/inputs/input-l
...
+			}
+
+			dialog.addEventListener('d2l-genai-reset-preview-pane', (e) => {
+				if (e.detail.resetStartEditing) {
+					dialog._startedEditing = false;
+				}
+			});
+
+			helpers.internalAPIWrapper.addToSharedDataCache('d2l-practice-source-text', () => {
+				return helpers.internalAPIWrapper.getContent(false, false, true);
```
JakeProuse, line 138: @grant-cleary Is this how I'd go about replicating [this call to getContent](https://github.com/BrightspaceUI/htmleditor/blob/682255770192a5e29a0596facacc08737deeca20/components/plugins/practices.js#L295)? I wasn't sure what the defaults were

components/plugins/practices.js
```diff
@@ -41,21 +41,190 @@ import { inputLabelStyles } from '@brightspace-ui/core/components/inputs/input-l
...
+
+	registerPlugin(context, helpers) {
+		const addPracticesDependencies = () => {
+			helpers.internalAPIWrapper.addFullPageStyleSheets(
+				helpers.internalAPIWrapper.getContent(),
+				[ { url: `${cdnDomain}/assets/practice.css`, attrs: new Map([['crossOrigin', 'anonymous']]) } ]
+			);
+
+			helpers.internalAPIWrapper.addFullPageScripts(
+				helpers.internalAPIWrapper.getContent(),
```
JakeProuse, line 79: @grant-cleary I call getContent a couple times here, is there a way I can just do it with one call like below to reduce the fetching or is there any performance impact to this?
```
const content = helpers.internalAPIWrapper.getContent()
```

components/plugins/practices.js
grant-cleary: That's right. I think I want to change this API to not have so many boolean params, but I'll wait for your work to be done to avoid you having to rewrite any of this.

components/plugins/practices.js
grant-cleary: Yeah, since this is just expecting a string, you could call it outside of the `addFullPage*` methods and just pass in the result in both cases.

JakeProuse: Looks like I run into an issue when re-using the content variable and it just adds one of them. In this case do I have to run getContent on each?
```
const content = helpers.internalAPIWrapper.getContent();

helpers.internalAPIWrapper.addFullPageStyleSheets(
	content,
	[ { url: `${cdnDomain}/assets/practice.css`, attrs: new Map([['crossOrigin', 'anonymous']]) } ]
);

helpers.internalAPIWrapper.addFullPageScripts(
	content,
	[ { url: `${cdnDomain}/helpers/events.js`, attrs: new Map([['crossOrigin', 'anonymous']]) } ]
);
```

components/plugins/practices.js
```diff
@@ -41,25 +42,192 @@ import { inputLabelStyles } from '@brightspace-ui/core/components/inputs/input-l
...
+
+			if (!node) return;
+
+			node.remove();
+		});
+
+		helpers.eventRegistry.registerGetContentEvent(() => {
+			const practices = helpers.internalAPIWrapper.findElements(`.${practiceClass}`);
+			if (practices.length > 0) {
+				addPracticesDependencies();
```
JakeProuse, line 196: I'm noticing this part of GetContent make a call stack exceeded error when dropping  a practice, should I be adding scripts/styles a different way here that won't kick off a looping setContent? This is where we used to set [the content](https://github.com/BrightspaceUI/htmleditor/blob/f13890b37f84164a55bd707b71c7a7bc90278f4b/components/plugins/practices.js#L404) that comes from the event

components/plugins/practices.js
grant-cleary: Oh, hmm, because it's mutating the content. It would be nice if we had a way to avoid these duplicate calls, but for now everything's designed to assume you're working within a handler that provides content within its event. Maybe just go with the duplication for now and I'll think more on this pattern.

components/plugins/practices.js
grant-cleary: Ah, so yeah, that's because you're calling `getContent()` as part of `addPracticesDependencies`, so this creates an infinite loop. How do we add dependencies for the other plugins? Because I don't think we run into the same problems there, no?

JakeProuse: Our other plugins just use SetContent, I don't know if we had a `GetContent` hook to make sure our deps were there on page load in case they were missing somehow? I guess SetContent would be running in that case anyway so maybe we don't need it. I'll have to ask Justin about it, but I'll remove the script/style addition from the hook for now.

grant-cleary: Yeah, nothing should be able to get rid of styles/scripts on `GetContent` (except other plugins), so as long as they're added in a `SetContent` handler, I think you should still be good here? I'm wondering if the use of `GetContent` here was just mimicking the older pattern we had originally used with Content Styler when script/stylesheet additions were effectively manual.

grant-cleary:
Doesn't the AI stuff tie into this somewhere? I'm not sure where that happens, but I suspect some of it might break because it's using the old editor object. 🤔

grant-cleary:
It's weird... I can see direct editor references being used, say, here: https://github.com/BrightspaceUI/htmleditor/blob/1d770bf506387a0a34a15ea649879671c8b03424/components/practice/llm-controller.js#L217, but I can't find anything that sets that static reference. 🤔

grant-cleary (approved):

grant-cleary:
That method looks unused, so I'm assuming it's dead code. I've asked in Slack, but I know Martin's PRs were intended to avoid setting that entirely, so I imagine we're good here.

PR #1719, opened 2023-09-13
fix: Including backend version in AI telemetry events

SheranWiratungaD2L:
The AI endpoint will send a `x-d2l-version` header containing the backend version. Including the value from this header in telemetry events.

Testing
- tested all supported question types: generated question and clicked feedback button, and confirmed (in debugger) that `BackendVersion` was set in telemetry event
- when `x-d2l-version` is not present, `BackendVersion` is an empty string


mdgbayly (approved):
LGTM

grant-cleary (approved):

PR #1720, opened 2023-09-13
US157488: AI Box should not be available for users who do not have the IMS Role Mapping

JamesVaChao:
Added a role check for instructor and admin. Should prevent people without those roles from seeing gen ai box and content.

Related Story:
[US157488](https://rally1.rallydev.com/#/?detail=/userstory/721484208551&fdp=true): AI Box should not be available for users who do not have the IMS Role Mapping


components/practice/ai/panel-controller.js
```diff
@@ -16,21 +17,58 @@ export class PanelController {
 		return false;
 	}
 
+	async checkEnrollments() {
+		const context = requestInstance(this.host, 'context');
+
+		const accessToken = await window.D2L.LP.Web.Authentication.OAuth2.GetToken('*:*:*');
+
+		const url = `${origin}/d2l/api/lp/1.35/enrollments/myenrollments/${context.orgUnitId}/access`;
```
mdgbayly, line 25: nit: I guess this origin is not specified so it's replaced with empty resulting in an absolute path reference (starts with a forward slash), which is probably sufficient given the way I think D2L Apis work. I see other parts of the htmleditor/practices code where it looks like they are making the same assumption, so should be fine.

mdgbayly (approved):
LGTM

grant-cleary (approved):

ianmclean2011 (approved):

components/practice/ai/panel-controller.js
JamesVaChao: I'm actually not sure where its defined but it's defined somewhere: 
image

grant-cleary: `origin` is a global property.

PR #1731, opened 2023-09-14
US157801 Fix source text max length by adding counter

mdgbayly:
### Summary of changes

Add character counter to source text dialog and disable generate button if length exceeded.

### Rally story

[US157801](https://rally1.rallydev.com/#/706276043991d/detail/userstory/722481578117): Fix max length validation in source text dialog


JamesVaChao (approved):

components/practice/ai/generate-question-source.js
```diff
@@ -136,10 +165,22 @@ class GenerateQuestionSource extends RtlMixin(Localizer(LitElement)) {
...
+	_opened() {
+		this.shadowRoot.querySelector('#dialog-content').scrollIntoView();
+	}
+
 	_resetSourceText() {
 		this.dispatchEvent(new CustomEvent('d2l-reset-source-text', {	bubbles:true, composed: true }));
 	}
 
+	// runs when user edits text
+	_updateTextLengthCount() {
```
mdgbayly, line 177: Function could be simplfied - didn't notice after I removed all the color set!ting.

components/practice/ai/generate-question-source.js
```diff
@@ -71,7 +92,7 @@ class GenerateQuestionSource extends RtlMixin(Localizer(LitElement)) {
 			<div class="dialog-div">
 				<d2l-dialog-fullscreen id="set-source-text-dialog" @d2l-dialog-before-close="${this._checkValid}" title-text="${this.localize('practice:llm:generateQuestionLabel')}">
 					<div id="dialog-content">
-						<div id="summary-text" class="ai-panel d2l-body-standard">
+						<div id="summary-text" class="ai-panel d2l-body-standard" autofocus>
```
mdgbayly, line 95: This is to prevent focus being set to the textarea which I think the dialog implementation is doing by default.

This `autofocus` seems dubious given that divs are not technically focusable, but it seems to work.

grant-cleary (approved):

PR #1734, opened 2023-09-15
US157834 Use subtle button color for thumbs icons

mdgbayly:
### Summary of changes

Not suggesting we should try and get this into the release but I guess it can be done if we really want it.

Override color of thumbs up/down icon buttons to be consistent with subtle button used elsewhere in the feedback panel.

There might be other implications of doing this (particular for consistency, but other things like theming, other settings, becoming inconsisent if subtle buttons are ever changed etc etc), but given this only applies to our component's usage of the buttons, it seems safe to do.

### Rally story

[US157834](https://rally1.rallydev.com/#/706276043991d/detail/userstory/722541165263): Consider color of thumbs up/down button icons


ianmclean2011 (approved):
Looks good!

grant-cleary (approved):

PR #1737, opened 2023-09-15
Improve vdiff reliability

grant-cleary:
Just making some small changes here to try to improve the reliability of a couple of the vdiff tests.


test/visual-diff/htmleditor.vdiff.js
```diff
@@ -28,6 +28,7 @@ describe('htmleditor', () => {
 		const editor = await elem._getEditor();
 		editor.execCommand('d2l-fullscreen');
 		await elem.updateComplete;
+		await focusElem(elem);
```
grant-cleary, line 31: The fullscreen functionality applies focus to the fullscreen toolbar button when fullscreen is enabled programatically. This makes for an unreliable visual-diff test here, though, as by the time we take our screenshot, focus may or may not be on the toolbar button. The solution here is to explicitly focus the editor, which will always push focus into the iframe, and won't leave us with an uncertain focus state by the time we take our comparison.

package.json
```diff
@@ -36,7 +36,8 @@
     "test:headless": "d2l-test-runner --chromium --timeout 10000",
     "test:headless:watch": "d2l-test-runner --chromium --watch",
     "test:vdiff": "d2l-test-runner vdiff",
-    "test:vdiff:golden": "d2l-test-runner vdiff --golden"
+    "test:vdiff:golden": "d2l-test-runner vdiff --golden",
+    "test:vdiff:report": "d2l-test-runner vdiff report"
```
grant-cleary, line 40: This is just useful locally.

test/visual-diff/components.vdiff.js
```diff
@@ -20,8 +20,8 @@ describe('components', () => {
 	runVisualDiffTestsWithTinyMCEVersioning(() => {
 
 		describe('button', () => {
-			const createButton = disabled => {
-				return fixture(html`
+			const createButton = async disabled => {
```
grant-cleary, line 23: None of these async/await changes actually do anything, because these were returning promises anyways. I just wanted to make it explicit here because that's what the other tests do already.

test/visual-diff/htmleditor.vdiff.js
```diff
@@ -12,10 +12,10 @@ describe('htmleditor', () => {
 		elem.shadowRoot.querySelector('iframe').contentDocument.body.style.caretColor = 'transparent';
 	};
 
-	const toggleChomping = elem => {
+	const toggleChomping = async elem => {
```
grant-cleary, line 15: Same as the other async/await comment. This wasn't broken; just inconsistent.

PaulKrznaric (approved):

PR #1742, opened 2023-09-18
Run unit tests in all browsers, not just Chrome

grant-cleary:
Now that we've fixed the "everything explodes on teardown" issue we were experiencing before, let's do what we can to finish cleaning up the tests so they can run in all major browsers.


htmleditor.js
```diff
@@ -332,6 +332,11 @@ class HtmlEditor extends HtmlEditorMixin(SkeletonMixin(LitElement)) {
 		}
 	}
 
+	blur() {
+		const focusedElem = getComposedActiveElement();
+		if (isComposedAncestor(this, focusedElem)) focusedElem.blur();
+	}
```
grant-cleary, line 338: This turned out not to be strictly necessary, because the tests can't make use of it anyways. However, I still want this in here, because it makes sure that if the editor is programatically blurred and TinyMCE's iframe happens to be focused, it will still behave as expected and fire a `blur` event.

PaulKrznaric (approved):

PR #1743, opened 2023-09-18
Don't wait for the click event; we don't need the promise it returns

grant-cleary:
Just a quick little optimization here.


PaulKrznaric (approved):
Should consider two PRs for these kind of long changes in the future.

grant-cleary:
> Should consider two PRs for these kind of long changes in the future.

Good point. It was irresponsible of me to try to condense this all into one.

PR #1752, opened 2023-09-18
fix: US157940 AI Questions - fix bug introduced by plugin refactor for getting source text from editor

mdgbayly:
### Summary of changes

The refactoring of the practices plugin changed the way we were getting the source text from the editor to use the new plugin  API, but it wasn't working as implemented. This PR switches back to using a regular request provider implemented by the practices dialog, but that still uses the new plugin API.

### Rally Story

[US157940](https://rally1.rallydev.com/#/706276043991d/detail/userstory/723422494351): Fix bug introduced by Practices plugin refactoring


grant-cleary:
I'm re-running the visual diff tests. They're failing because of a bit of flake I'm still sorting out from switching to the new framework. Not relevant to this PR.

grant-cleary (approved):

PR #1756, opened 2023-09-19
Don't destroy source code dialog on close

grant-cleary:
There seems to be an issue with CodeMirror within Firefox in particular, where destroying and recreating the source code dialog each time it's closed/opened causes the CodeMirror theme to be applied inconsistently. While I want to make sure I address that underlying issue, I don't want to make a change that significant within the existing refactor. Thus, for now, I'll just carve out an exemption to the way editor dialogs work to make sure the source code dialog is only ever created once.


PaulKrznaric (approved):

components/command-registry.js
```diff
@@ -116,7 +120,10 @@ export class CommandRegistry {
 			requestAnimationFrame(() => {
 				dialog.opened = false;
 				this.#component._dialogOpen = false;
-				dialog.remove();
+				// Because of an issue with CodeMirror in Firefox,
+				// removing the source code dialog from the DOM
+				// breaks the editing experience.
+				if (command !== cmds.sourceCode) dialog.remove();
```
grant-cleary, line 126: Note that this is the patterns dialog used to use. Destroying them on close is new with the refactor, so this is kind of like a very small, very targeted revert.

PR #1757, opened 2023-09-20
fix: US157085 Send telemetry event when a question is accepted

mdgbayly:
### Summary of Changes

@JakeProuse Just a heads up that this PR adds a call to our GenAI component from the Practices `saveActivity` handler.

- Call a method on the generative AI component when the practices editor Inserts a question so we can send an appropriate Gen AI telemetry event. 
- Ordinarily I'm not a fan of calling methods on components like this and prefer to make connections with events/props. But in this case the component doesn't get re-rendered after Insert is clicked so a prop wouldn't work and adding some kind of event based request provider/plugin mechanism seemed like overkill.
- There is a possibilty we may want to do something like that later if it's decided we need to add more specific telemetry, such as whether the question was edited following generation.
- Another option could have been to extend the existing creatorplus telemetry events, by indicating whether a question was AI generated, but that wouldn't be as consistent with other prod telemetry events we sent for the Gen AI feature which embed additional data.

From the Gen AI context other changes in this PR include:

- Choosing to discard generation status completely when the user changes the question type. 
- Previously I had chosen to not discard the generationId/count when changing question type, as I felt it could be interesting to know if users try different question types in the context of a generation workflow. 
- But this was always contentious and would have made it more complex to determine whether to send an "Accepted" telemetry event on Insert because changing question type does discard the current generated question.

### Rally Story

[US157085](https://rally1.rallydev.com/#/706276043991d/detail/userstory/720788850241): Additional product telemetry


grant-cleary (approved):

ianmclean2011 (approved):
Looks good!

PR #1771, opened 2023-09-21
Temporarily avoid problem with stylesheets being prematurely removed from the fullpage reference map

grant-cleary:
The way we opt to remove obsolete references here is a bit odd, built to suit a particular use case that no longer exists in the editor. In adapting this for script tags, we exposed a bug in which stylesheets can be prematurely removed from the reference map. Ideally, we would fix that particular problem... However, the complexity of this code and the variety of use cases it needs to support makes changing it a bit dangerous, so for now I'm opting for a simpler, albeit less ideal solution.


components/plugins/fullpage.js
```diff
@@ -251,19 +251,19 @@ export class FullPageState {
...
+			const refExistsInDoc = !!getExistingRef(doc, url);
 
 			// If a ref has been removed from the HTML markup (say, via the source code editor)
 			// and is still referenced in TinyMCE's iframe, wipe it from our ref map and remove
 			// it from the editor iframe. Script refs are never included in TinyMCE's iframe, so
 			// we just remove those references regardless.
 			if (removeObsolete && (existingEditorRef || type === refTypes.SCRIPT)) {
 				refMap.delete(url);
-				if (updateEditorDoc) existingEditorRef.remove();
+				if (!refExistsInDoc && existingEditorRef) existingEditorRef.remove();
```
grant-cleary, line 262: This is sort of conflating two things, and it's also inconsistent, because it leaves our ref map potentially out of sync with what's in the editor. However, the ref map will be regenerated every time the source code editor is saved, so in practice this doesn't turn out to be a problem. As much as this isn't how I would _prefer_ to solve this problem, it's the safer choice for now.

components/plugins/fullpage.js
```diff
@@ -251,19 +251,19 @@ export class FullPageState {
 		for (const [url, attrs] of refMap) {
 			const existingEditorRef = type === refTypes.STYLESHEET && getExistingRef(this._editor.contentDocument, url);
 			const updateInEditor = updateEditorDoc && !existingEditorRef;
-			const updateInDoc = !getExistingRef(doc, url);
+			const refExistsInDoc = !!getExistingRef(doc, url);
```
grant-cleary, line 254: I think inverting and renaming this check just makes it much easier to tell what it's actually doing.

PaulKrznaric, line 254: I understand why you've done this but it's still weird to look at.

test-helpers/htmleditor-test-context.js
```diff
@@ -2,5 +2,6 @@ console.warn('Using htmleditor test context, this is intended for demo pages and
 
 document.getElementsByTagName('html')[0].dataset.heContext = JSON.stringify({
 	useNewTinyMCEVersion: !(window.location.search.indexOf('oldVersion=true') !== -1),
-	useNewPluginFramework: true
+	useNewPluginFramework: true,
+	removeObsoleteScripts: true
```
PaulKrznaric, line 6: You didn't use this anywhere in this PR, is this supposed to be here?

PaulKrznaric (approved):

components/plugins/fullpage.js
grant-cleary: Which part is weird?

test-helpers/htmleditor-test-context.js
grant-cleary: This makes sure the "feature" for removing obsolete scripts is enabled on the editor's demo page. It's required to do any reasonable local testing on this.

components/plugins/fullpage.js
PaulKrznaric: `!!`

grant-cleary: Yeah, it's just type conversion, the JS way 😆: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_NOT#double_not_!!

PR #1776, opened 2023-09-22
feat: US156416 Initial idea for decoupling Practices and Gen AI generation

mdgbayly:
### Summary of Changes

This is a first attempt at decoupling the LCS managed Practices code and the Team HAI managed AI question generation code.

In the original implementation there was code in each of the Practice question components that explicity called the gen AI API and used internal functions of each question type to alter the question to match the generated question.

The idea of this PR is to remove that code from each question component. Instead, we use events fired from the composed `<d2l-generate-question-panel>` component to update the data in the Practices `activityData` structure and then use the standard practices render lifecycle to have the generated question data update the practice question UX. So the way the questions get rendered is using the same mechanism as if they were loaded from a file.

Other than practices continuing to embed an instance of our `<d2l-generate-question-panel>` component, the primary integration point is now a set of adapters in the `ai-adapters` folder which are used to adapt the generated question data into  the existing Practices `activityData` structure. These `ai-adapters` are conceptually owned by the Practices team and would live with the Practices code if/when we split out all this code into different repos.

As the AI generates the question, it sends an event with a detail object that contains the current state of the generated question We stream the question response from the API so the final question is built up incrementally over several events.

The format of the generated question object is based on the internal formats used by the existing Practices `activityData` but it is intended to be distinct. We want to use an event format that could also be used if this gen AI component was embedded in Brightspace quizzing too. But I didn't want to reinvent the wheel either so I just started with the Practices formats. But we don't want to share those formats. They need to be able to diverge if necessary. The `ai-adapters` are the glue where we will convert between the different formats.

With these changes in place, there is now no `LlmController` code in `plugins/practices.js` or the individual question files. Just the adapters and a reference to our top level `<d2l-generate-question-panel>`.

NOTE: As this is just a draft to try out the ideas, I have not removed code from the other question types yet.

Let me know what you think or if you have other ideas after seeing this that might be more elegant/better.

I acknowledge that having question type specific adapters is likely going to duplicate some code between the different question components and those adapters, such as what represents a new blank answer in an MC question. But I think with sufficient refactoring in Practices, some of that duplication could also be removed. Also I'm hand converting the formats at the moment, but there are possibly more elegant ways to do that, such as using something like a `deepcopy` lib for example, if the formats stay very similar. But wanted to keep things clear and simple for now.

I also acknowledge that sorting/sequencing questions use slightly different formats above the `activityData.question` sub object, but I'm pretty sure we can accomodate those with this approach too.

### Rally Story

[US156416](https://rally1.rallydev.com/#/?detail=/userstory/719624959645&fdp=true): Last main refactor of PQ generation code in HTML editor


components/practices/ai/questions/multiple-choice.js
```diff
@@ -0,0 +1,95 @@
...
+				{
+					'id': 1,
+					'text': '',
+				}
+			],
+			'correctAnswers': [],
+		};
+	}
+
+	async generate(api, data) {
```
mdgbayly, line 20: This is the bulk of the code lifted from the question components, tweaked to build up the generated question object and notify via events each time we have more data.

I think there are improvements we can make to this to reduce the number of events we fire, but we'll have more scope for that later once we are more decoupled.

components/practices/ai/questions/question-generator.js
```diff
@@ -0,0 +1,126 @@
...
+		]);
+	}
+
+	async generate(templateIndex, data) {
+		const questionGeneratorClass = this._questionTypes.get(templateIndex);
+		const questionGenerator = new questionGeneratorClass();
+		return questionGenerator.generate(this, data);
+	}
+
+	async getLlmStream(questionType, params, processContent) {
```
mdgbayly, line 22: This function is lifted mostly unchanged from the original `LlmController` and still has lots of the statics in it, but I see us replacing `LlmController completely` in future refactoring, and we can use class state in this module to replace all those statics.

components/practices/multiple-choice.js
```diff
@@ -225,14 +224,11 @@ class MultipleChoicePractice extends RequesterMixin(Localizer(LitElement)) {
...
 		this._noAnswers = false;
 		this._templateIndex = 5;
-
-		this.lc = new LlmController();
 	}
 
 	connectedCallback() {
 		super.connectedCallback();
 		this._context = this.requestInstance('context');
-		this.lc.orgUnitId = this._context.orgUnitId;
```
mdgbayly, line 235: We'll need to add code to grab the orgUnitId from the context in our own components.

components/practices/true-false.js
```diff
@@ -260,64 +256,6 @@ class TrueFalsePractice extends RequesterMixin(Localizer(LitElement)) {
 		}
 	}
 
-	clearFields() {
```
mdgbayly, line 263: We had added code to explicitly clear fields before generation and we also used this with our Discard generated question feature. In my testing just sending a default blank generated question is sufficient to clear the question prior to generation, and I plan on adding something similar to support our Discard question feature. Maybe a new event.

JakeProuse:
Even if this is a WIP I think that the adapter is a good middle ware between our logic!

JamesVaChao (approved):

mdgbayly:
Supersceded by a new version which incorporates all question types and does some other clean up.
https://github.com/BrightspaceUI/htmleditor/pull/1794

PR #1782, opened 2023-09-22
Add importable test framework for context menu plugins

grant-cleary:
Similar to the test framework we've added for plugin classes, we also want to add one for context menu plugins.


components/context-menu/formatting/formatting.js
```diff
@@ -13,7 +13,7 @@ const disallowedNodes = new Set(['IMG', 'FIGURE', 'AUDIO', 'VIDEO', 'IFRAME', 'O
 
 export class FormattingContextMenuPlugin {
 
-	static get name() {
+	static get pluginName() {
```
grant-cleary, line 16: The test exposed something funny. about this being a static property... Specifically, `name` is already an implicit property on any class. So if this wasn't defined within this plugin, `FormattingContextMenuPlugin.name` would still return `FormattingContextMenuPlugin`. This is not ideal and is definitely not intended, so I'm switching the property name to `pluginName` to force it to be explicitly set.

components/context-menu/link/link-edit.js
```diff
@@ -13,11 +13,11 @@ const elementName = 'd2l-htmleditor-link-edit-context-menu';
 
 export class LinkEditContextMenuPlugin {
 
-	static get name() {
+	static get pluginName() {
 		return 'link-edit';
 	}
 
-	static get tagName() {
```
grant-cleary, line 20: I don't know why this is static. The functionality requiring use of the tag name isn't needed for this particular context menu, which is the only reason this isn't broken in production.

JakeProuse (approved):

test/context-menu-plugins.test.js
```diff
@@ -0,0 +1,32 @@
...
+		{ name: 'insert-element', plugin: InsertElementContextMenuPlugin },
+		{ name: 'layouts', plugin: LayoutsContextMenuPlugin },
+		{ name: 'practices', plugin: PracticesContextMenuPlugin }
+	].forEach(info => {
+		describe(`${info.name} plugin structure`, () => {
+			it('should pass validation', () => {
+				const result = verifyContextMenuPluginStructure(info.plugin);
+				expect(
+					result.success,
+					'Context menu plugin verification failed. Encountered the following errors:\n\n'
```
PaulKrznaric, line 25: Doesn't this seem contradictory? Validation failed, but we're expecting a success/it to pass validation?

PaulKrznaric (approved):

test/context-menu-plugins.test.js
grant-cleary: The second parameter passed to `expect` is a message that's meant to be displayed on failure.

PR #1794, opened 2023-09-26
US156416 AI Questions: Refactor to decouple integration with Practices

mdgbayly:
### Summary of Changes

This is an updated/final version of the original draft PR fordecoupling the LCS managed Practices code and the Team HAI managed AI question generation code.

Sorry, it's quite a big PR but it made sense to make the changes for all the practice types as a single change. The changes follow a consistent pattern for each question type so once you understand how one question works the other changes should be fairly clear.

There is a new class `QuestionGenerator` that replaces the purpose of the original `LlmController`. This class is instantiated and called by the AI panel's UI controller when ever a new generation is requested.

For each question type the following changes were made:

- Move the code that handles the streaming response from backend LLM API from each of the practices components to a new utility class invoked by `QuestionGenerator` (one per question type)
- These utility classes were modfied to build up a question data structure that is modeled after the existing Practices `activityData` formats to make conversion to Practices data simpler.
- These components no longer mutate the internal practices data structures directly
- After each chunk of question response data is processed by the question specific classes, the `QuestionGenerator` dispatches an event with the current state of the `question` data.

This event is handled by the `PracticeDialog` and passed through an adapter layer which will be part of the Practices code base and owned by the practices team. There is one adapter implementation per question type. It takes the basic question data generated by the AI component, and merges it into the existing Practices `activityData.questions` data object, and adds additional Practices specific data such as input types and other default properties not populated byes the AI component.
The question then re-renders with the updated generated question data.

Other changes include:

- Removing all code relating to sequencing and sorting questions as we are not supporting those types right now
- Removing the majority of the now unused code from the `LlmController`. There are still some statics being used from this class that will be cleaned up in the future.

#### Some additional notes/details from original draft PR

Other than practices continuing to embed an instance of our `<d2l-generate-question-panel>` component, the primary integration point is now the set of adapters in the `ai-adapters` folder which are used to adapt the generated question data into  the existing Practices `activityData` structure. These `ai-adapters` are conceptually owned by the Practices team and would live with the Practices code if/when we split out all this code into different repos.

The format of the AI generated question object is based on the internal formats used by the existing Practices `activityData` but it is intended to be distinct. We want to use an event format that could also be used if this gen AI component was embedded in Brightspace quizzing too. But I didn't want to reinvent the wheel either so I just started with the Practices formats. But we don't want to share those formats. They need to be able to diverge if necessary. The `ai-adapters` are the glue where we will convert between the different formats.

With these changes in place, there is now no `LlmController` code in `plugins/practices.js` or the individual question files. Just the adapters and a reference to our top level `<d2l-generate-question-panel>`.

### Rally Story

[US156416](https://rally1.rallydev.com/#/?detail=/userstory/719624959645&fdp=true): Last main refactor of PQ generation code in HTML editor


grant-cleary (approved):
LGTM! Looks very clean. :)

ianmclean2011 (approved):
I like it 😄

JakeProuse (approved):

SheranWiratungaD2L (approved):

JamesVaChao (approved):

PR #1800, opened 2023-09-27
Restructure wrappers slightly for testing

grant-cleary:
Some restructuring was necessary to get the new wrappers into a state where they could be easily tested. This PR addresses that restructuring, and fixes a couple small typos in latent code.


components/plugin-registry.js
```diff
@@ -1,3 +1,4 @@
+import * as lmsAdapter from './lms-adapter.js';
```
grant-cleary, line 1: This ends up being a change in behaviour of sorts. While the functionality doesn't change, this does guarantee that the entire `lms-adapter` module is imported whenever the editor initializes. In practice, though, the only expensive bit of the `lms-adapter` module is ifrau, which is already dynamically loaded, so I don't expect this to incur much extra cost.

components/plugin-registry.js
```diff
@@ -226,7 +227,7 @@ function buildRegistries(editor, isPrivileged) {
 		eventRegistry: new EventRegistry(component, editor),
 		internalAPIWrapper: new InternalAPIWrapper(component, editor),
 		privilegedInternalAPIWrapper: new PrivilegedInternalAPIWrapper(component, editor),
-		uiRegistry: new UiRegistry(component, editor)
```
grant-cleary, line 229: The `UiRegistry` class didn't actually use this second param, so it did nothing.

components/ui-registry.js
```diff
@@ -1,11 +1,12 @@
-import { openDialog, openDialogWithParam, openLegacyDialog, openQuicklinksDialog } from './lms-adapter.js';
```
grant-cleary, line 1: ES6 imports are immutable and thus can't be stubbed by `sinon`. To get around this, I just pass the adapter module into the class and invoke its methods directly.

components/api-wrapper.js
```diff
@@ -156,6 +155,8 @@ export class InternalAPIWrapper {
 	}
 
 	getIsfContextId() {
+		if (this.#isfContextId) return this.#isfContextId;
+		this.#isfContextId = requestInstance(this.#component, 'isfContextId');
```
grant-cleary, line 159: Because ES6 imports are immutable, I can't stub out `requestInstance`. While I can (and do) stub out the things it's doing to the component in tests, running it in the constructor is annoying because it requires these stubs to be set up for every single test against this class. Better to confine it to this method to keep the side effects under control.

components/api-wrapper.js
```diff
@@ -64,7 +63,7 @@ export class InternalAPIWrapper {
 		const node = this.#editor.selection.getNode();
 		if (!node) return;
 
-		const anchor = this.#editor.dom.getParent(this.#editor.selection.getNode(), 'a');
+		const anchor = this.#editor.dom.getParent(node, 'a');
```
grant-cleary, line 66: Minor optimization. There's no need for us to fetch the node twice.

components/api-wrapper.js
```diff
@@ -145,7 +144,7 @@ export class InternalAPIWrapper {
 	}
 
 	getFromSharedDataCache(key) {
-		return this.#component.getFromSharedDataCache(key);
+		return this.#component._getFromSharedDataCache(key);
```
grant-cleary, line 147: Typo, though this method is currently unused so there's no actual impact.

components/ui-registry.js
```diff
@@ -3,9 +3,11 @@ import { getComposedActiveElement } from '@brightspace-ui/core/helpers/focus.js'
 
 export class UiRegistry {
 	#component;
+	#lmsAdapter;
 
-	constructor(component) {
+	constructor(component, lmsAdapter) {
 		this.#component = component;
+		this.#lmsAdapter = lmsAdapter;
```
grant-cleary, line 10: Tests will pass in a custom adapter module so it can be mocked out (can't stub ES6 modules, so the direct imports can't be mocked). Production code will pass in nothing, so the direct imports can be used.

grant-cleary:
I do want to get this set up on my local build so I can verify that it doesn't break anything, but it's reviewable in the state it's in now.

PaulKrznaric (approved):

grant-cleary:
Local testing looks good. Fairly confident I didn't horribly break any LMS dialogs.

PR #1806, opened 2023-10-02
feat: US156396 - GenAi - Added Affordance Component

JamesVaChao:
Added an affordance component which is used in Edit HTML screen.

image

Related PR:
https://github.com/BrightspaceHypermediaComponents/activities/pull/4248

To Test:
Enable and disable GenAi and see if affordance shown. Should be reliant on genAiEndpoint. Therefore, only be displayed with GenAi enabled.
Need to test on legacy activity vs new activity

Related Stories
[US156396](https://rally1.rallydev.com/#/?detail=/userstory/719624512103&fdp=true): Add affordance on content page edit screen (split into one per content exp)


components/practices/ai/affordance-controller.js
```diff
@@ -0,0 +1,59 @@
+export class AffordanceController {
+
+	constructor(host) {
+		this.host = host;
+		this.canAccess = false;
+	}
+
+	async checkEnrollments() { // 1 of 2 checkEnrollments to remove
```
ianmclean2011, line 8: I've forgotten what the plan is for this. Is it just to move it to an LMS-side check?

ianmclean2011 (approved):
Looks good to me :)

grant-cleary (approved):

components/practices/ai/affordance-controller.js
mdgbayly: Couldn't we just put this code in a shared module and use in both places?

PR #1808, opened 2023-10-02
Add final test fixtures for external wrappers and correct exception-throwing tests

grant-cleary:
Three things going on in this PR, really:
1. Adding the final test fixtures for the toolbar builder and plugin registry.
2. Correcting some existing tests checking for exception throwing.
3. Making sure to check dependencies in the plugin test helpers.


components/plugin-registry.js
```diff
@@ -118,6 +117,10 @@ export class PluginRegistry {
 		return plugin.dependencies && plugin.dependencies.size > 0;
 	}
 
+	#hasLmsContext() {
+		return this.#context.activeOrgEnrollment;
```
grant-cleary, line 121: This is the same behaviour that the import was formerly doing. I want to inline it in here because it's just unnecessary to jump through that hoop when we can just do the check in here instead.

PaulKrznaric (approved):

PaulKrznaric (approved):

klhuff (approved):

PR #1818, opened 2023-10-04
US158615: Add dynamic import for gen ai component to use BSI

ianmclean2011:
PR covers [US158615](https://rally1.rallydev.com/#/?detail=/userstory/727403904341&fdp=true): Get genai component in htmleditor from private repo

Dynamically loads the gen ai component from BSI.


components/plugins/practices.js
```diff
@@ -51,6 +50,8 @@ import { sparkleIcon } from '../practices/ai/ai-styles.js';
 import { ValenceController } from '../practices/valence-controller.js';
 import { VALID_TEMPLATE_ID } from '../practices/ai/generate-question-consts.js';
 
+import('/node_modules/@d2l/generate-question/components/generate-question-panel.js');
```
ianmclean2011, line 53: This returns a promise, which I'm not doing anything specific with. Is there a preference for handling this differently? Should I be conditionally loading this?

grant-cleary: As written, this will import that dependency as soon as this file is loaded, whether or not it's ever used. That's probably fine, given that this is only a temporary measure, but usual practice would be to load the dependency only when it's actually needed. i.e. await the import when we're in a situation where we're going to need to load/render the component, and only render it if the promise resolves.

grant-cleary: Actually, based on the test failures, I'm going to amend my statement to say you'll definitely want to move this such that it only imports when we're going to need to update the practices component. Visual diff and unit tests are both failing because the import doesn't resolve locally, and because of how the practices plugin is structured, that file always gets loaded.

ianmclean2011: Yep, was just thinking that 😄

JamesVaChao (approved):

grant-cleary (approved):

PR #1820, opened 2023-10-04
Limit release workflow concurrency to avoid conflicts

grant-cleary:
When two or more PRs are merged in quick succession, the semantic-release action can end up pulling stale refs and then failing. This is very annoying because it requires (my) manual intervention to clean up, followed by another PR from the original author to force a new release. To try to address this, I'm making a change to allow only one copy of the release workflow to run at once across the repo. Subsequent releases will be stuck queued until previous ones complete.

Similarly, we want to force the static site publishing job to wait on the release job to finish before running.


.github/workflows/release.yml
```diff
@@ -5,6 +5,7 @@ on:
       - main
       - '[0-9]+.x'
       - '[0-9]+.[0-9]+.x'
+concurrency: release
```
grant-cleary, line 8: This will apply across all branches. While that may not seem ideal, this particular type of semantic-release conflict can arise when two PRs on _different branches_ (e.g. `main` and `2.75.x`) are merged simultaneously. This is because the first release action to finish creates a release/tag in the repo, causing the ref used by the second action to become stale.

PaulKrznaric (approved):

PR #1824, opened 2023-10-06
Don't load plugins that require an LMS context if said context is not available

grant-cleary:
I think there's value in making it more explicit that certain plugins require an LMS context to function. This lessens the likelihood that a misconfiguration will cause the HTML editor to break by trying to load a plugin that calls internal LMS functions, and makes this differentiation more obvious to plugin developers.


components/plugins/contentstyler.js
```diff
@@ -12,6 +12,10 @@ export class ContentStylerPlugin {
 		return name;
 	}
 
+	get requiresLmsContext() {
+		return false;
```
grant-cleary, line 16: @JakeProuse, @jschlumkoski... This is something I'm considering doing that will impact the plugin implementations in your repo as well. I also wasn't 100% sure whether Insert Element and Layouts require an LMS context to work properly, so would appreciate your feedback on that. I know Content Styler does not, and Practices definitely does.

components/plugin-registry.js
```diff
@@ -215,13 +218,16 @@ function addTinyMCEPlugin(plugin, pluginName) {
...
+	const hasLmsContext = context.activeOrgEnrollment;
+
 	if (registryMap.has(editor.id)) {
 		const registries = registryMap.get(editor.id);
 		return {
 			commandRegistry: registries.commandRegistry,
 			eventRegistry: registries.eventRegistry,
 			internalAPIWrapper: isPrivileged ? registries.privilegedInternalAPIWrapper : registries.internalAPIWrapper,
-			uiRegistry: registries.uiRegistry
+			uiRegistry: hasLmsContext ? registries.lmsAdaptedUiRegister : registries.uiRegistry
```
grant-cleary, line 230: I'm not 100% sure about this differentiation. On the one hand, I think it makes sense to not provide tools plugins can use to shoot themselves in the foot if they accidentally misconfigure their plugin. On the other hand... It's JavaScript and this helpers object is kind of opaque anyways, so does this actually make a difference? 🤷

Certainly it's true that these LMS adapters will all explode if they're used without a valid LMS context, so I still think not passing them in makes more sense.

components/plugins/contentstyler.js
JakeProuse: I think our IE and Layouts plugins do need the context for opening a dialog (I seem to remember it being required, may be wrong here) @jschlumkoski ?

grant-cleary:
After chatting with @dbatiste offline, I _think_ this is probably the best solution for now. I don't think it's ideal for the editor to have to know things about the LMS, and ideally these are things that individual plugins would handle themselves... But the existence of FRAs forces the editor to have at least some LMS knowledge, and this overall change at least keeps that centralized.

I would expect this API to change in the future once FRAs are dead, but since we'll be dealing with external plugins by then, I expect we'll have to figure out a bit of a migration path to get there.

components/plugins/contentstyler.js
JakeProuse: In that case Layouts may be fine, Insert Element [uses the valence controller](https://github.com/BrightspaceUI/htmleditor/blob/main/components/plugins/insert-element.js#L36) so I think it'll still need it

JakeProuse: Layouts also fires events though, while it doesn't explicitly get cookies they are needed for the request

grant-cleary: Ah okay. Might be that it's safer to classify both of these plugins as requiring the LMS then. It mostly comes down to whether you ever anticipate (or want) your plugins to run in contexts with no backing LMS. There aren't many cases where that's currently a thing, but they exist.

JakeProuse (approved):

PR #1859, opened 2023-10-30
Add framework required for creating wrapped commands around all TinyMCE defaults

grant-cleary:
PR 1/N - This will be the first of likely many PRs looking to implement a proposed new toolbar design. Under this paradigm, we'll be moving toolbar construction for the entire editor to function the way external plugins do now; plugins will provide a standard interface for surfacing toolbar items to the toolbar builder, which will then construct a JSON representation of the toolbar that the toolbar component itself will render.

This does come with some sweeping changes elsewhere. For this to work sensibly, we need _all_ toolbar items that interact with TinyMCE to be surfaced from some kind of plugin. This means we need wrappers for the existing logic that ties into TinyMCE directly.


components/commands.js
```diff
@@ -95,3 +95,58 @@ export const cmds = {
...
+	tablePasteRowBefore: 'd2l-table-paste-row-before',
+	tableProps: 'd2l-table-properties',
+	tableSortAdvanced: 'd2l-advanced-table-sort',
+	tableSortColumnAsc: 'd2l-sort-table-by-column-ascending',
+	tableSortColumnDesc: 'd2l-sort-table-by-column-descending',
+	tableSplitCells: 'd2l-table-split-cells',
+	tableRowProps: 'd2l-table-row-properties',
+	underline: 'd2l-underline',
+	undo: 'd2l-undo',
+	unorderedList: 'd2l-insert-unordered-list',
```
PaulKrznaric, line 151: is this a trailing comma?

PaulKrznaric (approved):

components/commands.js
grant-cleary: Yeah, but JS objects don't care about trailing commas. 😆

grant-cleary: For my own picky standards I'll fix it with the next PR.

klhuff (approved):

PR #1861, opened 2023-10-31
Add new internal item params

grant-cleary:
PR 2/N - Now we get to the second piece of required work to prep for the toolbar refactor; we need to expand out our `item-params` definitions, making sure to add the various parameters we'll need to support internal toolbar item definitions. This also requires renaming the existing external extension point to be more descriptive, but nothing external to the editor repo is using this yet, so it's a safe change.


components/command-registry.js
```diff
@@ -21,6 +21,7 @@ const builtInCommands = {
 	ltr: 'mceDirectionLTR',
 	orderedList: 'insertOrderedList',
 	outdent: 'outdent',
+	preview: 'mcePreview',
```
grant-cleary, line 24: Why, GitHub... This was already in my previous merged PR.

grant-cleary:
Ick, something is weird with this PR. GitHub doesn't seem to realize what state this stuff should be in. I'm going to close and recreate it.

PR #1866, opened 2023-10-31
Add toolbar item code to the editor plugins (and create new ones as necessary)

grant-cleary:
PR 4/N - This PR adds the toolbar item code itself to the various plugins, and creates fill-in plugins to wrap existing TinyMCE functionality in the case where we were formerly relying on this behaviour directly.

Note that nothing consumes these yet, so this PR should still have no effect on the editor as things stand.


components/plugins/image.js
```diff
@@ -109,6 +116,35 @@ export class ImagePlugin {
...
+		return [
+			{
+				breakpoint: {
+					type: breakPointTypes.ABOVE,
+					value: breakPoints.LARGE
+				},
+				cmd: cmds.image,
+				extensionPoint: extensionPoints.INSERT_SECTION,
+				icon: icons.image,
+				sortOrder: 3,
```
PaulKrznaric, line 129: These have the same sort order?

PaulKrznaric (commented):
I'll continue this review around 2:30pmish

components/plugins/image.js
grant-cleary: Sort orders are specific to the `extensionPoint`.

components/plugins/equation.js
```diff
@@ -25,6 +32,85 @@ export class EquationPlugin {
...
+						type: menuItemTypes.MENU_ITEM_TOGGLE
+					},
+					{
+						cmd: cmds.equationChemistry,
+						icon: icons['equation-chemistry'],
+						title: this.#localizer('equation:chemistry'),
+						type: menuItemTypes.MENU_ITEM_TOGGLE
+					}
+				],
+				sortOrder: 4,
```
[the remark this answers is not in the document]
grant-cleary: Because we need to maintain parity with the existing toolbar. When the editor is at its full size and we haven't hit any of the responsive breakpoints, the equation editor toolbar item is the fourth in its section. However, when we hit the breakpoint that hides the equation editor in the Insert More menu, it moves into a different extension point and its rendering order changes.

In general, the way to look at these sort orders is to consider each "section" of the toolbar being defined (by the `extensionPoint`) and looking at where that item renders within that section when the specified breakpoint is satisfied.

components/plugins/image.js
klhuff: The sort orders depend on what extension point they're being added to. These aren't being added to the same one.

klhuff: eg. Between all of the files, there should be a linear sort order between things within the same extension point that they're being added to.

grant-cleary: That's right. I had hoped not to need a sort order at all (magic numbers suck), but unfortunately in the original design spec of the editor, the toolbar items don't necessarily maintain their relative order when they're shifted into menus by the responsive breakpoints. This means we can't just sort the plugins to maintain the correct ordering, but actually have to specify sorting on an item-by-item basis.

Also of note: There doesn't always _need_ to be a sort order. If an entire extension point is self-contained within a single plugin, then sort order is given implicitly by the order of toolbar items within that plugin.

klhuff (approved):

PaulKrznaric (approved):

PR #1868, opened 2023-11-01
feat: US158339 - Move to new release token

svanherk:
Once this is approved, I'll remove the `brightspace-bot` as a collaborator and remove access to the `D2L_GITHUB_TOKEN` before merging and releasing.

Also updating the README while I'm here, we've been trimming down the crazy release section and I noticed the vdiff stuff was pointing to the old commands.


.github/workflows/release.yml
```diff
@@ -31,7 +31,7 @@ jobs:
       - name: Semantic Release
         uses: BrightspaceUI/actions/semantic-release@main
         with:
-          GITHUB_TOKEN: ${{ secrets.D2L_GITHUB_TOKEN }}
+          GITHUB_TOKEN: ${{ secrets.D2L_RELEASE_TOKEN }}
```
svanherk, line 34: Once this merges and I've updated the setup to remove every `brightspace-bot` related, this line will need to be backported to any old release branches that need to release for a hotfix/certfix

grant-cleary: I'll make a note to take care of that. Thanks Stacey!

grant-cleary (approved):

.github/workflows/release.yml
svanherk: Thanks! It should only be an issue for the next few weeks, then any new release branch will have this change when it's created

PR #1870, opened 2023-11-02
Remove redundant code meant to support prototype version of registering external toolbar items

grant-cleary:
PR 5/N - This PR removes the vestiges of the previous external toolbar item registration method. Ultimately, no external plugins were using this code, and keeping it supported under the flag off state while also maintaining the new flag-on behaviour was going to add too much complexity to the code.

Note that I'm not purging the toolbar-builder file because some of the code in there will be reused and it would be a bit of a pain to have to start fresh.


components/toolbar/button-menu.js
```diff
@@ -125,7 +125,7 @@ class ButtonMenu extends DropdownOpenerMixin(ToolbarItemMixin(LitElement)) {
 				item.selected = (item.value === this._value);
 				if (item.selected) {
 					hasSelected = true;
-					if (item.textContent) this._valueText = item.textContent;
+					if (item.textContent.trim()) this._valueText = item.textContent;
```
grant-cleary, line 128: This is a small unrelated bug I found. If there's any meaningful whitespace within the element's root (as can happen when bootstrapping Lit elements), this check would have returned true even if no non-whitespace text was set anywhere.

test-helpers/htmleditor-test-context.js
```diff
@@ -3,5 +3,6 @@ console.warn('Using htmleditor test context, this is intended for demo pages and
 document.getElementsByTagName('html')[0].dataset.heContext = JSON.stringify({
 	useNewTinyMCEVersion: !(window.location.search.indexOf('oldVersion=true') !== -1),
 	useNewPluginFramework: true,
-	removeObsoleteScripts: true
+	removeObsoleteScripts: true,
+	toolbarRefactor: true
```
grant-cleary, line 7: This allows the toolbar refactor to take over in the demo pages. Note that this isn't hooked up in the LMS right now, so even once the final editor PRs are in, there won't be anything to see here in LMS builds.

components/toolbar/toolbar-full.js
```diff
@@ -101,7 +101,6 @@ class Toolbar extends ToolbarMixin(RequesterMixin(LitElement)) {
 			layouts: { type: Boolean, attribute: 'layouts' },
 			noInsertCode: { type: Boolean, attribute: 'no-insert-code' },
 			sourceEditable: { type: Boolean, attribute: 'source-editable' },
-			toolbarStructure: { type: String, attribute: 'toolbar-structure' },
```
grant-cleary, line 104: There's no point keeping this around because the new way of doing things is going to use a completely new component for rendering the toolbar.

klhuff (approved):

PaulKrznaric (approved):

PR #1873, opened 2023-11-02
jprouse/DE55634-layouts-context-menu-not-switching-layout-type

JakeProuse:
Rally Story: https://rally1.rallydev.com/#/?detail=/defect/730033994139&fdp=true


components/plugins/layouts.js
```diff
@@ -310,7 +310,7 @@ export function addLayoutsPlugin() {
 		const context = requestInstance(editor.getElement(), 'context');
 		const component = context.component;
 		const localize = context.component.localize.bind(context.component);
-		const layouts = layoutsHtml(localize, context.useNewPluginFramework, editor, null);
+		const layouts = layoutsHtml(localize, false, editor, null);
```
grant-cleary, line 313: I feel like this is likely to break stuff, since the context menu framework is still passing in `EditorNode` objects.

grant-cleary: Oh wait, this is the old plugin. Never mind. I misread.

grant-cleary (approved):

components/plugins/layouts.js
grant-cleary: Though I'm curious how this can be possible. This code shouldn't be reachable when the new plugin is available, since this function should never be called. In what case does that come up?

JakeProuse: Haha yeah sorry this is an oldie. I might just need to backport this, still waiting on a decision there 😄

JakeProuse: This is that 20.23.10 thing where we force the old C+ plugins to load regardless of the flag, but my old layouts plugin would keep trying to use the new API handler since true would get passed in here

JakeProuse: So I don't think I'll need to merge this to main really, it'd just be a backport if we decide to hotfix 20.23.10.

PR #1879, opened 2023-11-06
Add builders to construct toolbars from provided plugins

grant-cleary:
PR 6/8 - This PR adds the framework required to take plugins and their toolbar items and convert them to the JSON structures that will eventually be consumed by the new toolbar component. The only editor-specific information we consume here is the editor type, since that determines which plugins we load and in which order we render various extension points.


components/toolbar/toolbar-builder.js
```diff
@@ -1,89 +1,371 @@
...
+	if (!externalPlugins) return extensionPointMappings;
+
+	const externalExtensionPoint = extensionPoints.INSERT_MORE_MENU_EXTENSIBLE_SECTION;
 
-function constructExtensibleToolbarSection(externalPlugins) {
-	const extensibleToolbarItems = [];
 	externalPlugins.forEach(plugin => {
 		if (!plugin.toolbarItems) return;
-		extensibleToolbarItems.push(...constructToolbarSegment(plugin.toolbarItems, true));
+		plugin.toolbarItems.forEach(item => {
```
klhuff, line 38: I'm assuming we know these lists will always be relatively small, and the nested loops will be ok?

components/toolbar/toolbar-builder.js
```diff
@@ -1,89 +1,371 @@
...
+			extensionPointMappings[externalExtensionPoint].push(item);
+		});
+	});
+
+	return extensionPointMappings;
+}
+
+function filterExtensionPoints(extensionPointMappings, parentExtensionPoint) {
+	const containedExtensionPoints = [];
+	if (parentExtensionPoint === extensionPoints.INSERT_SECTION || parentExtensionPoint === extensionPoints.INLINE_SECTION) {
```
klhuff, line 50: Would it make sense to do this with some kind of static dictionary object?

klhuff (approved):

components/toolbar/toolbar-builder.js
```diff
@@ -1,89 +1,371 @@
...
+				extensionPointMappings[extensionPoint] = extensionPointMappings[extensionPoint].concat(pluginExtensionPointMappings[extensionPoint]);
+			} else {
+				extensionPointMappings[extensionPoint] = pluginExtensionPointMappings[extensionPoint];
+			}
+		}
+	});
+
+	if (!externalPlugins) return extensionPointMappings;
+
+	const externalExtensionPoint = extensionPoints.INSERT_MORE_MENU_EXTENSIBLE_SECTION;
```
PaulKrznaric, line 34: Does this function not always return a value?

components/toolbar/toolbar-builder.js
```diff
@@ -1,89 +1,371 @@
...
+			extensionPoints.INSERT_MORE_MENU_EXTENSIBLE_SECTION
+		);
+	} else if (parentExtensionPoint === extensionPoints.VIEW_SECTION) {
+		containedExtensionPoints.push(extensionPoints.VIEW_MENU);
+	}
+
+	const filteredExtensionPointMappings = {};
+	containedExtensionPoints.forEach(extensionPoint => {
+		const items = extensionPointMappings[extensionPoint];
+		if (items && items.length > 0) filteredExtensionPointMappings[extensionPoint] = items;
```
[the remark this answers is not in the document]
grant-cleary: `filteredExtensionPointMappings` is an object, so if there are multiple extension points, it will just produce an object that looks like this:
```
{
   extensionPoint1: [
      item1,
      item2
   ],
   extensionPoint2: [
      item3,
      item4
   ]
}
```

components/toolbar/toolbar-builder.js
grant-cleary: That's not really guaranteed, but any kind of insane nesting is also going to be visually problematic (i.e. the menus are going to start overlapping each other, overflowing off the page, etc.). For the plugins that exist right now, that is true, and I'm not _super_ concerned about this happening in external plugins. It's already possible for someone to create an external plugin that just eats all memory in a loop. 🤷

components/toolbar/toolbar-builder.js
grant-cleary: I don't think it would buy us much, but I can make that change.

PaulKrznaric (approved):

components/toolbar/toolbar-builder.js
grant-cleary: Sorry, I don't follow. What do you mean?

components/toolbar/toolbar-builder.js
```diff
@@ -1,89 +1,370 @@
...
+];
+
+const viewMenuExtensionPoints = [
+	extensionPoints.VIEW_MENU
+];
+
+const containedExtensionPoints = {};
+containedExtensionPoints[extensionPoints.INSERT_SECTION] = insertMoreMenuExtensionPoints;
+containedExtensionPoints[extensionPoints.INLINE_SECTION] = insertMoreMenuExtensionPoints;
+containedExtensionPoints[extensionPoints.VIEW_SECTION] = viewMenuExtensionPoints;
```
grant-cleary, line 62: Unfortunately we can't use the pretty object initializer syntax when the object keys are dynamic.

components/toolbar/toolbar-builder.js
```diff
@@ -1,89 +1,370 @@
...
+	};
+
+	return opener;
+}
+
+function addExtensionPointSeparators(items, extensionPoint) {
+	if (!items || items.length === 0) return items;
+
+	const menu = insertMoreMenuExtensionPoints.includes(extensionPoint)
+		|| viewMenuExtensionPoints.includes(extensionPoint);
```
grant-cleary, line 104: Once I broke out the arrays of menu extension points, I also figured it would be cleaner to just reference those here instead of doing a further manual check of menu extension points.

klhuff (approved):

components/toolbar/toolbar-builder.js
PaulKrznaric: The function early returns if there's no externalPlugins on line 32, but that's the only case we return a value as far as I can tell. 

So doesn't it as a result have slightly different behaviour in it's return values?

PaulKrznaric (approved):

PR #1882, opened 2023-11-07
Add new toolbar component to render constructed toolbar state

grant-cleary:
PR 7/8 - Now we're finally getting into the parts where this new work actually gets used/rendered. This PR does a couple things:
1. Adds a new generic `d2l-htmleditor-toolbar` component that knows how to render the toolbar structure provided by the builders.
2. Tweaks our existing buttons and menu items to play a bit more nicely with these new creations.

Note that while I would love to be able to test the toolbar component explicitly, we're going to have to live with the existing html-editor visual-diff tests for now. Unfortunately the various toolbar button/menu-item components are still strongly coupled to TinyMCE and we can't really scaffold that in a sensible way during tests. I hope to decouple that at some point, but for now we're kind of stuck.


components/toolbar/toolbar.js
```diff
@@ -0,0 +1,658 @@
+import './button.js';
```
grant-cleary, line 1: This is almost exclusively a copy of the [existing toolbar mixin](https://github.com/BrightspaceUI/htmleditor/blob/main/components/toolbar/toolbar-mixin.js). I'll comment where there are important differences.

components/toolbar/toolbar.js
```diff
@@ -0,0 +1,658 @@
...
+	render() {
+		const classes = {
+			'd2l-htmleditor-toolbar-container': true,
+			'd2l-htmleditor-toolbar-left': this._anchorLeft,
+			'd2l-htmleditor-toolbar-overflowing': this._overflowing,
+			'd2l-htmleditor-toolbar-chomping': this._chomping,
+			'd2l-htmleditor-toolbar-measured': this._measured
+		};
+
+		const { items, pinnedItems } = this._renderToolbarItems();
```
grant-cleary, line 267: The way we fetch the toolbar items is new as of this component. The mixin used to rely on its children generating the HTML markup and passing `items` and `pinned` items into its `_render` method.

components/toolbar/toolbar.js
```diff
@@ -0,0 +1,658 @@
...
+
+		return html`
+			<div
+				class="${classMap(classes)}"
+				data-state="${this._state}"
+				@keydown="${this._handleKeyDown}"
+				role="toolbar">
+				${keyed(this.floating, html`
```
grant-cleary, line 275: The use of `keyed` here is new. What this does is instructs Lit that it shouldn't try to re-use DOM elements unless the key (in this case, `this.floating`) stays the same between renders. If we don't have this, then Lit tries to reuse the existing DOM elements where possible when switching between the inline and fullscreen toolbars. This messes up their internal state, so we really want a fresh DOM every time we swap between toolbars.

grant-cleary:
Eh, I forgot that `d2l-icon-custom` overwrites the fill colour of its contained SVGs. That's maybe not desirable, but I'll have to double-check with Design on that in the long term. I'll probably just switch back to rendering the SVGs for now because this is technically an impactful change.

components/toolbar/toolbar.js
```diff
@@ -0,0 +1,658 @@
...
+			</div>
+			${!this.floating ? html`<div class="d2l-htmleditor-toolbar-border"></div>` : ''}
+		`;
+	}
+
+	updated(changedProperties) {
+		super.updated(changedProperties);
+		changedProperties.forEach(async(_, propName) => {
+			if (propName === 'toolbarStructure' && this.toolbarStructure) {
+				await this._handleResize(this._measures);
```
grant-cleary, line 292: This is new. The old toolbars and mixin were designed to be completely destroyed and recreated every time we needed to change out the toolbar. To make this new toolbar resilient to its structure changing, we need to force a resize when that happens.

components/toolbar/toolbar.js
```diff
@@ -0,0 +1,658 @@
...
+		if (this.floating) return;
+
+		this._showFull = (measures.available > breakPoints.FULL);
+		this._showLarger = (measures.available > breakPoints.LARGER);
+		this._showLarge = (measures.available > breakPoints.LARGE);
+		this._showMedium = (measures.available > breakPoints.MEDIUM);
+		this._showSmall = (measures.available > breakPoints.SMALL);
+		this._showSmaller = (measures.available > breakPoints.SMALLER);
+		this._showTiny = (measures.available > breakPoints.TINY);
+		this._showMinimum = (measures.available > breakPoints.MINIMUM);
```
grant-cleary, line 396: These are titled more generically than the breakpoint properties used in the old mixin, but they serve the same purpose.

components/toolbar/toolbar.js
```diff
@@ -0,0 +1,658 @@
...
+		const chomper = this.shadowRoot.querySelector('.d2l-htmleditor-toolbar-chomper');
+
+		this._measures = {
+			available: availableWidth,
+			flex: {
+				items: [],
+				total: 0
+			},
+			pinned: {
+				chomper: chomper ? chomper.offsetWidth : 0,
```
grant-cleary, line 645: The chomper is no longer baked into the toolbar, so we need to add a bit of extra safety here (it may not exist on first render).

components/toolbar/toolbar.js
```diff
@@ -0,0 +1,658 @@
...
+			_overflowing: { type: Boolean, state: true },
+			_state: { type: String, state: true },
+			_showFull: { type: Boolean, state: true },
+			_showLarger: { type: Boolean, state: true },
+			_showLarge: { type: Boolean, state: true },
+			_showMedium: { type: Boolean, state: true },
+			_showSmall: { type: Boolean, state: true },
+			_showSmaller: { type: Boolean, state: true },
+			_showTiny: { type: Boolean, state: true },
+			_showMinimum: { type: Boolean, state: true },
```
grant-cleary, line 66: These `state` decorators are new; they're effectively a mechanism to tell Lit that it shouldn't expect these to be changed by anything external to the component.

components/toolbar/button-menu.js
```diff
@@ -20,10 +21,9 @@ class ButtonMenu extends DropdownOpenerMixin(ToolbarItemMixin(LitElement)) {
 	static get properties() {
 		return {
 			cmd: { type: String },
+			explicitIcon: { type: Boolean, attribute: 'explicit-icon' },
 			icon: { type: String },
-			noOpenerContent: { type: Boolean, attribute: 'no-opener-content' },
```
grant-cleary, line 24: I'm getting rid of some of these properties and slots because they're just not used. I think this was originally done because we usually use slots for our icons, but it's just not necessary for the editor's use cases.

components/focus-helpers.js
```diff
@@ -11,7 +11,10 @@ const keyCodes = Object.freeze({
 export function resetActiveFocusable(focusables) {
 	const item = focusables.find(item => item.activeFocusable);
 	if (item) item.activeFocusable = false;
-	focusables[0].activeFocusable = true;
+
+	const eligibleFocusable = focusables.find(item => !item.hasAttribute('responsive-hidden'));
+	if (eligibleFocusable) eligibleFocusable.activeFocusable = true;
+	else focusables[0].activeFocusable = true;
```
grant-cleary, line 17: This focusing logic previously assumed that the first item in the list of focusables could never be `responsive-hidden`. This was formerly true in the old toolbars, but now isn't. To harden this, we look for the first _visible_ item when determining when to apply focus (`find` iterates through the array from start to end and returns the first match).

components/toolbar/button.js
```diff
@@ -1,17 +1,29 @@
...
 
+const undoCommands = [
+	'undo',
+	'd2l-undo'
+];
+
+const redoCommands = [
+	'redo',
+	'd2l-redo'
+];
```
grant-cleary, line 18: Speaking of editor logic wrapped into these components... Unfortunately we currently have to have hard-coded logic for managing undo/redo button states tied in here, so now that these commands are wrapped, we need to be ready to look for both their wrapped and unwrapped versions.

components/toolbar/button-menu.js
grant-cleary: As a side note, most of these toolbar button changes can't really be feature flagged effectively, so unfortunately we're going to have to just test extensively in the hopes that none of them bleed through. Visual diff tests passing give me fairly high confidence that this isn't an issue, but I'll make sure to call it out as an area requiring more detailed testing.

PaulKrznaric (approved):

components/toolbar/button-menu.js
```diff
@@ -53,7 +53,8 @@ class ButtonMenu extends DropdownOpenerMixin(ToolbarItemMixin(LitElement)) {
 			:host([dir="rtl"]) button {
 				padding: 0 10px 0 6px;
 			}
-			button > div {
+			button > div,
```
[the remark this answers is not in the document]
grant-cleary: That's the CSS child combinator, so the selector will apply to any `div`s that are immediate children of any `button`s.

components/toolbar/button-menu.js
```diff
@@ -149,9 +146,13 @@ class ButtonMenu extends DropdownOpenerMixin(ToolbarItemMixin(LitElement)) {
 					tabindex="${this.activeFocusable ? 0 : -1}"
 					title="${this.text}">
 					<div class="d2l-htmleditor-button-background"></div>
-					${!this.noOpenerContent ? html`<div class="${hasIcon ? 'd2l-htmleditor-button-menu-icon-container' : ''}">
-						${this.icon ? unsafeHTML(icons[this.icon] || '') : html`<slot @slotchange="${this._handleIconSlotChange}" name="icon">${this._valueText ? this._valueText : this.text}</slot>`}
-					</div>` : null}
+					${icon ? html`
```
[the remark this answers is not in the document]
grant-cleary: Hmm, I don't think so. The check against `this._valueText` is independent of `icon` (if `icon` doesn't exist) so there's nothing nice to create a `switch` from. I think it would complicate the code to do this.

klhuff (approved):

PR #1884, opened 2023-11-07
Formatting context menu should render regardless of toolbar refactor flag state

grant-cleary:
Just an oversight in the toolbar refactor... The formatting context menu should be registered _before_ the flag check to ensure it renders with or without it enabled.


grant-cleary:
FYI @jesseag, this appears to be the culprit behind the behaviour you were seeing.

PaulKrznaric (approved):

njostonehouse (approved):

PR #1886, opened 2023-11-08
Expose layouts property for plugin consumption

grant-cleary:
This is needed in the `insert-element` plugin to determine whether to show the layouts option in its internal editor.


grant-cleary:
> Is this to replace our checking of insert-element-dialog-open?

Ah, no, it's to replace this: https://github.com/BrightspaceUI/creator-plus-authoring-tools/blob/main/plugins/insert-element.js#L2356

Once all of the various refactoring work is done, I'm going to lock down the editor's exports to avoid exposing its internals, and that `editor-adapter` is one of the things I'll likely be cutting out.

PR #1887, opened 2023-11-08
Hook up new toolbar component

grant-cleary:
PR 8/8 - And finally, the finish line! Now that we have our fancy new toolbar component, we need to hook it up and disconnect the bits and bobs that used the old toolbar. Once the flag is removed, there's more restructuring that can be done to make this all even simpler, but while we're supporting both flag states, there's some complexity we have to keep.


grant-cleary:
Interesting. The linter locally disagrees with the sorting CI is expecting. I must have an older version installed locally.

htmleditor.js
```diff
@@ -19,16 +20,6 @@ import { SkeletonMixin } from '@brightspace-ui/core/components/skeleton/skeleton
...
-// To update from new tinyMCE install
-// 1. copy skins from installed node_modules/tinymce into tinymce/skins
-// 2. copy new language packs from https://www.tiny.cloud/get-tiny/language-packages/ into tinymce/langs
-// 3. copy new enterprise plugins into tinymce/plugins
-
-// TODO: review whether pasted content needs prepcessing to avoid pasted image links getting converted to images
-// TODO: configure paste_as_text if using tinyMCEs paste as text feature (fra editor sets to false if power paste enabled) - probably not needed
-// TODO: review whether we need to stop pasting of image addresses (see fra editor)
-// TODO: review allow_script_urls (ideally we can turn this off)
-
```
grant-cleary, line 31: I keep forgetting to remove this. I will not forget again.

htmleditor-mixin.js
```diff
@@ -492,6 +495,16 @@ export const HtmlEditorMixin = superclass => class extends Localizer(LabelledMix
 		});
 	}
 
+	willUpdate(changedProperties) {
```
grant-cleary, line 498: For those unfamiliar with Lit's reactive lifecycle, `willUpdate` allows us to pre-compute properties without triggering another update. In some sense, `_toolbarStructure` is a computed property that depends on the value of `this._fullscreen`.

PaulKrznaric (approved):

klhuff (approved):

PR #1893, opened 2023-11-10
Use CSS to prevent toolbar pop-in and avoid unnecessary rendering logic

grant-cleary:
This is just a tiny optimization to allow some of the toolbar structure to pre-render before we compute all of the plugins. The original goal was to prevent some gross-ish pop-in where the toolbar border would render at the top of the editor on first initialization, taking up a few pixels and causing some icky pop-in.

This new change does the same, but it uses CSS to hide the toolbar until it has contents instead of preventing its rendering entirely. Slightly cleaner code, and allows for the toolbar framework to be rendered on initial load instead of forcing it to be deferred.


PaulKrznaric (approved):

klhuff (approved):

grant-cleary:
Blech, hopefully that fullscreen test isn't still flaky. I certainly explicitly tested fullscreen to make sure this small change didn't cause any weird issues there, and it definitely works as expected.

PR #1895, opened 2023-11-10
Use exports to restrict external access to HTML editor internals

grant-cleary:
Right now, despite all of the effort put in to restrict plugin access to TinyMCE and other HTML editor internals, it's still quite easy for a project importing the HTML editor package from NPM to just use any of the functionality it exports internally. This isn't ideal, so I want to try to make use of the `exports` property to limit which files well-behaved apps can access. 

Note that this isn't foolproof. Nothing stops an app from reaching down into `node_modules` and plucking whatever it wants out of the editor, so this isn't intended to stop malicious behaviour. After all, absolutely nothing stops someone from writing a plugin that just does `delete window.tinymce` and thus completely bricks the editor. The intent is to make it more difficult to import things we don't want imported so developers don't accidentally do this.


test-helpers/plugin-helpers.js
```diff
@@ -1,534 +1,10 @@
...
-	if (success) return { success: true };
-	else return {
-		success: false,
-		messages: messages
-	};
-}
+export {
+	verifyContextMenuPluginStructure,
+	verifyExternalPluginStructure as verifyPluginStructure
+};
```
grant-cleary, line 10: Moving the bulk of the actual logic into a different file and selectively exporting only the bits meant to be externally accessible makes things a lot cleaner. Otherwise we would need to bleed some of the internal helpers in a way that could be confusing to other developers.

test/plugins.test.js
```diff
@@ -47,7 +47,7 @@ describe('d2l-htmleditor plugins', () => {
 	].forEach(info => {
 		describe(`${info.name} plugin structure`, () => {
 			it('should pass validation', () => {
-				const result = verifyInternalPluginStructure(info.plugin, text => text);
+				const result = verifyPluginStructure(info.plugin, false, text => text);
```
grant-cleary, line 50: There was really no reason to have an explicit export here. Any tests internal to the editor repo should know how to use this internal function, so the wrapper was just unnecessary.

package.json
```diff
@@ -7,6 +7,18 @@
...
+    "./components/toolbar/button.js": "./components/toolbar/button.js",
+    "./components/toolbar/button-toggle.js": "./components/toolbar/button-toggle.js",
+    "./components/toolbar/button-menu.js": "./components/toolbar/button-menu.js",
+    "./components/toolbar/item-params.js": "./components/toolbar/item-params.js",
+    "./components/toolbar/menu-item.js": "./components/toolbar/menu-item.js",
+    "./components/toolbar/separator.js": "./components/toolbar/separator.js",
+    "./htmleditor": "./htmleditor.js",
+    "./htmleditor.js": "./htmleditor.js",
+    "./test-helpers/plugin-helpers.js": "./test-helpers/plugin-helpers.js"
+  },
```
grant-cleary, line 21: From what I could tell by searching through our three relevant orgs, these are the only imports currently being consumed within our platform. There's [one exception to this](https://github.com/BrightspaceUI/creator-plus-authoring-tools/blob/dee767d2fe5c792885fecedec4653d7375226640/plugins/events/layouts/layouts.js#L3), but I'm going to shift a copy of that code into the C+ repo as opposed to exposing it here. We don't really have a paved path for front-end telemetry, so I don't want to promote it as part of the editor contract.

grant-cleary: Note that the exception has been resolved. That code has been copied into the repo that was using it and the import has been removed.

PaulKrznaric (approved):

Dan-DeAraujo (approved):

PR #1898, opened 2023-11-13
Toolbar menu item separators should not display unless at least one item is displaying

grant-cleary:
Currently, at certain responsive breakpoints, we can get into a situation where multiple separators render despite not having any items to separate, displaying what looks like one much thicker separator instead.

Ideally, we want menu separators to inherit breakpoints such that they're only visible when at least one of the items they're separating are visible. To accomplish that in the general case would be fairly tricky, since we could, in theory, have a mix of `ABOVE`, `BELOW`, and `BETWEEN` breakpoints all in the same set of menu items.

However, in practice, this only occurs in the Insert More menu, and in that case, the only breakpoints at play are `BELOW` breakpoints, so for simplicity, that's the only case I'm going to address here.


components/toolbar/toolbar-builder.js
```diff
@@ -103,8 +103,23 @@ function addExtensionPointSeparators(items, extensionPoint) {
...
 		|| viewMenuExtensionPoints.includes(extensionPoint);
 
+	// Menu separators should only display when there's at least one
+	// item that satisfies its breakpoint. Currently, this will always
+	// be of breakpoint type BELOW. If this changes, we will need to
+	// revisit this logic.
+	const getMenuSeparatorBreakpoint = () => {
+		const maxBreakpointValue = Math.max(items.filter(
+			item => item.breakpoint && item.breakpoint.type === breakPointTypes.BELOW
+		).map(item => item.breakpoint.value));
```
grant-cleary, line 113: Choosing this value will cause the menu separator to satisfy its breakpoint as soon as the first item within the segment also does so.

PaulKrznaric (approved):

davewagler-d2l (approved):

PR #1899, opened 2023-11-13
Reset focusables when toolbar structure changes

grant-cleary:
Now that the toolbar's structure can actually change after first render, we need to make sure we're resetting focusables properly when the toolbar structure does change. Otherwise we could end up with some random toolbar item becoming the focusable toolbar element after a re-render (or worse, no item at all becoming focusable).


PaulKrznaric (approved):

components/toolbar/toolbar.js
```diff
@@ -292,6 +292,7 @@ export class Toolbar extends RtlMixin(LitElement) {
 		changedProperties.forEach(async(_, propName) => {
 			if (propName === 'toolbarStructure' && this.toolbarStructure) {
 				await this._handleResize(this._measures);
+				this._resetActiveFocusable();
```
grant-cleary, line 295: This feels a bit heavy-handed, but if the toolbar structure changes, I don't think we can really make any assumptions about how focus should be set, beyond simply resetting it to the first available toolbar item. There's maybe a world in which we could try to scan the toolbar to see if the previously focusable item still exists and is focusable post-update, but that's a massive performance penalty for what amounts to (in my opinion) very little benefit.

Dan-DeAraujo (approved):

PR #1900, opened 2023-11-13
Force context menu to sit before footer in the tab order

grant-cleary:
A while ago, as part of the WCAG 2.2 work, the editor's resize handle was changed to be keyboard accessible. As a consequence, it became part of the tab order, which led to an awkward situation where the resizer actually ended up _before_ the context menu in the tab order. The goal with this PR is to fix this by having the context menu render in a container that comes before the footer.


htmleditor-mixin.js
```diff
@@ -919,7 +919,7 @@ export const HtmlEditorMixin = superclass => class extends Localizer(LabelledMix
 
 				if (useNewPluginFramework) {
 					if (this.fullPage) this._initializeFullPageState(editor);
-					initializeContextMenus(this, editor, true);
+					initializeContextMenus(this, editor, true, this._context && this._context.useContextMenuContainer);
```
grant-cleary, line 922: I hate multiple boolean flag parameters as much as the next person, but JS basically doesn't do function overrides, so this is what we get instead.

grant-cleary: Or I guess it's more accurate to say that JS _only_ does function overrides 😆. Same outcome though.

ca2hill (approved):

davewagler-d2l (approved):

PR #1907, opened 2023-11-16
Toolbar buttons should use events to interact with TinyMCE

grant-cleary:
The intent here is to decouple the toolbar items from TinyMCE by using events to drive those particular interactions. This allows us to actually test these components separately, as well as further centralizing which files within the editor access TinyMCE functionality directly.


components/toolbar/event-helpers.js
```diff
@@ -0,0 +1,69 @@
...
+		this.#handler = handler;
+	}
+
+	registerEvent(editor) {
+		editor.on(this.#eventName, () => this.#handler(editor));
+	}
+
+}
+
+export function dispatchExecuteCommandEvent(component, cmd, value) {
```
grant-cleary, line 16: Given that these all need to consume the component, I do think this might be better as a mixin.

grant-cleary:
I think there are some timing changes here that have finally tipped the fullscreen visual diff tests back into flaky mode, though at least this time the failure is consistent within the tests. I'll put some work into the tests when I'm closer to getting this across the finish line.

htmleditor-mixin.js
```diff
@@ -1311,4 +1364,12 @@ export const HtmlEditorMixin = superclass => class extends Localizer(LabelledMix
 		this.requestValidate(showNewErrors);
 	}
 
+	_validateToolbarEvent(event) {
+		const rootTarget = event.composedPath()[0];
+
+		return rootTarget
+			&& rootTarget.nodeType === Node.ELEMENT_NODE
+			&& validToolbarEventTargets.includes(rootTarget.tagName);
```
grant-cleary, line 1372: Context menus and overlays can include arbitrary components. We definitely do not want these to be able to craft events to allow directly interacting with TinyMCE.

grant-cleary:
Going to hold off on merging until after 20.24.2 branching.

klhuff (approved):

davewagler-d2l (approved):

PR #1914, opened 2023-11-22
[DE55422] Adding a URL in the mini html editor in a rubric criterion …

jesseag:
…box changes focus.

Rally ticket: [DE55422](https://rally1.rallydev.com/#/57732444928d/dashboard?detail=%2Fdefect%2F728905531923)

Adding in a click handler to stop propagation of the click event in the Title/Link field to the `rubric-feedback` component.
https://www.loginradius.com/blog/engineering/javascript-events-bubbling-capturing-and-propagation/#:~:text=For%20example%2C%20if%20your%20button,click%20event%20along%20the%20way.

Rubric click handler which I am hoping to stop propagation to.
https://github.com/Brightspace/d2l-rubric/blob/main/src/components/rubric-feedback.js#L627
https://github.com/Brightspace/d2l-rubric/blob/main/src/components/rubric-feedback.js#L448-L452

Fix:

mini-html-editor-click-handler-fix


components/context-menu/link/link-edit.js
```diff
@@ -113,7 +113,7 @@ class LinkEditContextMenu extends Localizer(RtlMixin(LitElement)) {
 
 	render() {
 		return html`
-			<div class="d2l-htmleditor-link-context-menu-edit" tabindex="-1" @focus="${this._handleFocus}">
+			<div class="d2l-htmleditor-link-context-menu-edit" tabindex="-1" @focus="${this._handleFocus}" @click="${this._handleClick}">
```
grant-cleary, line 116: Oh, you know, I wonder if this is bubbling up from here: https://github.com/BrightspaceUI/core/blob/main/components/inputs/input-text.js#L557

I'll investigate this a bit more, but for now I think this change is fine as is.

grant-cleary (approved):

components/context-menu/link/link-edit.js
jesseag: I did a little investigation and you are right this is where the event is bubbling up from. Would you prefer that I moved this code there instead? I also tried e.stopPropagation() here and it worked as well.

The question on my mind and for the core/Gaudi team would be is anyone relying on this click event bubbling up? If so, then I think we go with this current solution. Otherwise, it probably makes more sense to go with a solution in the `input-text` component instead.

jesseag: @dbatiste Wondered what your thoughts are here? Or is there someone else on Gaudi who I could tag about the `d2l-input-text` component?

grant-cleary: Yeah, that was also my concern. If it's definitely not intended/preferred that the click event bubbles up (which I think would be a question for Gaudi), then yeah, this code likely better lives there. If that's a desired feature (or if Gaudi is concerned by the risk), then I agree that making the change here is good.

devpow112 (approved):
Looks good to me :tada:

components/context-menu/link/link-edit.js
jesseag: I've opened a core PR with these changes to see what the Gaudi team says.
https://github.com/BrightspaceUI/core/pull/4278

grant-cleary: Cool beans. Dave's response seems reasonable to me, so I see no reason not to merge this. :)

jesseag: Sounds good. Did you want a comment here in the code indicating this is a defect fix?
Dave L's response indicates that click events are generally supposed to propagate upwards so this change is a bit opposite to the normal behaviour.

grant-cleary: I'm fine either way. I think the mostly-empty handler is enough of a signal that we're deliberately stopping propagation, but if you want to add extra context that's also fine.

PR #1920, opened 2023-11-27
Ensure we properly reset plugin registrations when reconnecting to the DOM

grant-cleary:
Under the new plugin refactor feature, if the HTML editor is disconnected from the DOM and reconnected, we run into an issue where the cached plugins don't get reset properly. We need to reset these because TinyMCE needs its plugins re-registered when we re-create the editor, but we want to make sure we don't wipe out stuff that should stay static after the editor is first rendered.

@semibran, @jesseag - FYI, but this should address the re-ordering issue you were seeing with the use of `repeat`.


PaulKrznaric (approved):

components/plugin-registry.js
```diff
@@ -46,13 +46,19 @@ export class PluginRegistry {
...
 
 	constructor(component, context) {
 		this.#component = component;
 		this.#context = context;
+		this.#pluginInitializationComplete = false;
 
 		this.#defaultPlugins = new Map();
 		this.#externalPlugins = new Map();
+		this.#filteredDefaultPlugins = new Map();
+		this.#filteredExternalPlugins = new Map();
```
Dan-DeAraujo, line 61: nit: `filtered` isn't a particularly descriptive word, maybe something to explain what filtering is applied, eg: `resettableDefaultPlugins` ? 
super-nit: now that there's 2 types, maybe a descriptor for the others as well, eg: `static` 🤷 

neither of these is blocking of course... and i don't spend enough time in this repo to say if it's even worth doing, so that's upto you 😛

Dan-DeAraujo (approved):

components/plugin-registry.js
grant-cleary: Fair point. In the interest of keeping the backports moving, I'll likely just merge this for now, but I've made a point to revisit the way things are named here.

The gist of it is that the `defaultPlugins` and `externalPlugins` sets always contain all plugins provided to the editor, whether or not they should or should not be displayed. The filtered variants filter on whether the plugin should be loaded based on the properties passed into the editor (e.g. don't load the ISF plugin if we're not in an LMS).

PR #1936, opened 2023-11-29
Ignore false positives from a11ychecker when an image is decorative (has empty alt text)

grant-cleary:
WCAG guidelines state that an image having empty alt text is sufficient to alert to screenreaders that it's intended to be decorative. TinyMCE's a11ychecker plugin, however, is overly zealous here and is expecting other conditions to be met that aren't required or advised under the guidelines. As such, we want to make sure we filter out these false positives.


htmleditor-mixin.js
```diff
@@ -779,6 +779,12 @@ export const HtmlEditorMixin = superclass => class extends Localizer(LabelledMix
 		tinymce.init({
 			a11y_advanced_options: true,
 			a11ychecker_allow_decorative_images: true,
+			a11ychecker_filter_issue: issue => {
+				if (issue.id !== 'I1') return true;
+				const elm = issue.element;
+				if (elm.nodeType !== Node.ELEMENT_NODE || elm.tagName.toLowerCase() !== 'img') return true;
+				return elm.getAttribute('alt') !== '';
```
grant-cleary, line 786: Worth noting that this code block is only invoked once the a11ychecker has decided an issue is present. So while it seems odd to imply it's not okay for alt text to be some non-empty string, the idea is that if we're in that position, there must be something else wrong that's causing the a11ychecker to flag an issue.

davewagler-d2l (approved):

PaulKrznaric (approved):

PR #1937, opened 2023-11-29
Add new attachmentsOnly property and pass it through to ISF/Insert Image

grant-cleary:
We need to make a couple changes within the editor to support the pending ISF work:
1. Pass through to ISF whether we're allowed (or forced) to upload to course files.
2. Generalize the `attached-images-only` attribute to extend to arbitrary files for use in ISF.


components/plugins/isf.js
```diff
@@ -94,11 +94,29 @@ export class ISFPlugin {
...
+			}
+
 			const result = await helpers.uiRegistry.openLmsLegacyDialog(
 				`/d2l/common/dialogs/isf/selectItem.d2l
 					?ou=${context.orgUnitId}
 					&extensionPoint=${isfContextId || ''}
-					&filterMode=${noFilter ? 'None' : 'Strict'}`,
+					&filterMode=${noFilter ? 'None' : 'Strict'}
+					&allowSaveToCourseFiles=${allowSaveToCourseFiles ? 1 : 0}
+					&forceSaveToCourseFiles=${forceSaveToCourseFiles ? 1 : 0}`,
```
grant-cleary, line 119: These query parameters won't be read within the LMS for now, so this is ultimately a no-op. They will be needed to support the associated LMS changes, however.

htmleditor-mixin.js
```diff
@@ -139,10 +139,15 @@ export const HtmlEditorMixin = superclass => class extends Localizer(LabelledMix
 	static get properties() {
 		return {
 			/**
-			 * ADVANCED: Whether or not to restrict image uploads to attachments and prevent saving to course/shared files
+			 * DEPRECATED: Whether or not to restrict image uploads to attachments and prevent saving to course/shared files
```
grant-cleary, line 142: This feels kind of unnecessary because the LMS is the only consumer currently setting this attribute. However, I still want to approach this as a breaking change. Once the LMS has been updated to pass in the new attribute, I'll come back and remove references to the old one.

PaulKrznaric (approved):

PR #1940, opened 2023-11-30
Properly check for existence of tinymce global before trying to clear editor state

grant-cleary:
Implicitly checking globals works fine, except when that global hasn't yet been defined. Then there's ambiguity and the JS engine expects it to be a local variable. Note that this isn't breaking anything client-side, as if the `tinymce` global isn't defined, then there's no possible way any editors could exist to have their undo states cleared. It's just generating log noise.


grant-cleary:
FYI @TristanDevos, but this should resolve the issue you noted. Thanks for bringing it to my attention!

njostonehouse (approved):

PR #1944, opened 2023-12-04
Account for file data return from ISF dialog result

grant-cleary:
This is another pre-requisite PR for the ISF work. In order to trigger file uploads on save, the editor needs to know there's a file to upload in the first place. This PR aims to add support for this to the editor.


components/plugins/isf.js
```diff
@@ -124,7 +125,23 @@ export class ISFPlugin {
...
+							result.FileInfo.FileId,
+							result.FileInfo.FileName,
+							result.FileInfo.Size,
+							result.Url
+						)
+					);
+
+					helpers.internalAPIWrapper.insertContent(result.EmbedSource);
+				} else {
+					helpers.internalAPIWrapper.insertContent(result);
```
grant-cleary, line 142: This `else` condition won't be relevant anymore once the associated LMS code starts to merge. For now, however, we need this here, lest we break ISF entirely until then

PaulKrznaric (approved):

davewagler-d2l (approved):

PR #1953, opened 2023-12-08
Announce inline toolbar when it renders after the user clicks into it

grant-cleary:
When a user clicks into the inline editor and the toolbar renders, there's no indication to screenreader users that they can navigate to the toolbar (or how to do it). The intent here is to explicitly announce those instructions when the toolbar renders.


grant-cleary:
Will merge this in 20.24.2

davewagler-d2l (approved):

klhuff (approved):

PR #1954, opened 2023-12-08
Pull HTML from the editor earlier in the disconnect process before TinyMCE has discarded it

grant-cleary:
When the inline editor is disconnected from the DOM, we rely on TinyMCE's `detach` event as a marker for us to persist TinyMCE's contents before wiping them out. It turns out, however, that by the time this event fires, TinyMCE has already wiped its internal state, so we end up saving blank HTML and the `html-block` renders nothing once it's moved. This isn't actually destroying any content because the underlying `textarea` TinyMCE uses to save its contents is still intact, but it does make it appear like data has been lost until the user clicks into the editor.

The solution here is to persist earlier in the process, before we try to remove TinyMCE at all.


grant-cleary:
Will merge this in 20.24.2.

klhuff (approved):

davewagler-d2l (approved):

PR #1960, opened 2023-12-13
Don't decode the MathML attribute on rendered math containers, as this is no longer encoded

grant-cleary:
When we still rendered MathML as images within the HTML editor iframe, we stored these encoded because that's how the data we got back from WIRIS was formatted. Now that we're not using WIRIS for rendering, however, we no longer encode the `data-d2l-mathml` attribute, and so we shouldn't try to decode it either.

As for why we shouldn't try to decode it, `decodeUriComponent` throws if it runs into a `%` symbol that isn't followed by two hex characters, i.e. if it runs into something that looks like it should be encoded but clearly isn't.


components/plugins/equation.js
```diff
@@ -152,7 +152,7 @@ export class EquationPlugin {
 
 			let mathML = '';
 			if (contextNode && getEditorTypeForMathContainer(contextNode) === editorType) {
-				mathML = decodeURIComponent(contextNode.attributes.getNamedItem('data-d2l-mathml').value);
+				mathML = contextNode.dataset.d2lMathml;
```
grant-cleary, line 155: Remove the encoding, but also switch to the simpler and easier-to-read dataset notation.

grant-cleary:
I'm only altering the new plugin version here because I won't be merging this until the plugin refactor flag has been removed.

davewagler-d2l (approved):

klhuff (approved):

PR #1965, opened 2023-12-18
Update to use the latest fonts version from the CDN.

dbatiste:
This PR simply updates the fonts version we're pointing to on the CDN.

Eventually (after the holidays) we will add BC Sans as a choice to the editor toolbar, but not in this PR.


components/plugins/fullpage.js
```diff
@@ -4,7 +4,7 @@ export const fullDocTestRegex = /<head[^>]*>.*<\/head>/s;
    authored content may require it as long as we support Lato in the editor, and
    removing Lato from the editor means disruptive change when editing existing content
 */
-const fontsUrl = 'https://s.brightspace.com/lib/fonts/0.5.0/fonts.css';
+const fontsUrl = 'https://s.brightspace.com/lib/fonts/0.6.0/fonts.css';
```
dbatiste, line 7: I figured I would put this PR up so we can discuss.

Existing content that is using Lato would already have `https://s.brightspace.com/lib/fonts/0.5.0/fonts.css` in its HTML. 

In this particular case, `0.6.0` is additive in the sense that it has everything that `0.5.0` has, plus the BC Sans font-face definitions, so it would be "ok" if we replaced the old link with a new link.

However, if the change was "breaking" (like when we removed the OpenDyslexic font), we would not want to replace the old link if the content is relying on something that was removed. In that case, I would normally expect a major version bump.

On the other hand, if the change was to fix something with the fonts, we would probably want to replace the old link so that existing content would get the fixed font version the next time the user edits the topic. In that case, I would normally expect a patch revision.

If we don't replace the old link, we'll have some redundancy if a user edits an old topic containing the `0.5.0` version, which is probably fine (I can test that).

Maybe... we never replace the font if the major version is different, otherwise we do replace the link.

Not sure how smart we want to make this. Interested in your thoughts.

grant-cleary: Yeah, there are some interesting points there. As it stands now, we look for the exact fonts URL when deciding whether to add the new one, so in cases like this where we have a version bump, I believe we'll end up with both the old and the new versions sitting on top of each other. As a side note, we also explicitly only add this if our Lato regex fails, so we won't end up adding this if the content uses BC Sans.

On the broader point though, I think it would indeed be ideal if we could set some bar for what types of changes we were willing to replace. i.e. We should have the latitude to replace the font if something is broken, but we should otherwise be a bit nervous about it. e.g. I don't think we can meaningfully ever remove Lato from that font package without breaking a bunch of users' content.

I think using the usual semver rules for that does make sense, so on that I agree. The nice thing is that because we control the regex we're using to find and replace the fonts, we can also change this whenever we feel like it, so at least we're not really tied down to a particular scheme here.

dbatiste:
Closing this PR since I am combining this change with the rest of what is needed to support BC Sans in the editor.

PR #1968, opened 2023-12-20
Update shared lint config to 1.0

grant-cleary:
Bumping our shared lint configuration to 1.0, and hiding a few errors that cropped up because of the new rules being added.


components/plugins/insert-element.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable lit/no-native-attributes */
```
grant-cleary, line 1: All of these plugins are going away next release, so I'm opting to ignore the linting rules here instead of fixing them.

joshhoey (approved):

PR #1969, opened 2023-12-21
When sending an execute command event, menu items should pass their value, if applicable

grant-cleary:
I missed this when switching the toolbar items over to an event-based communication model, since it's actually a property on one of the mixins.


PaulKrznaric (approved):
LBTM

PR #1986, opened 2024-01-04
Update link format API to account for new link workflow (with text instead of title)

grant-cleary:
PR 2: Now we're getting a bit closer... We need to update the link formatting API to correctly handle adding or updating link text as appropriate when we're looking to add a link. This is more complicated than it seems, since links can contain more than just plain text, so we need a chunk of logic here to account for that. I'll comment within the PR as appropriate, but the general logic is as follows:
- If we're updating an existing anchor, add text if and only if that anchor's descendants contain only inline text nodes (i.e. plain text, bold, underline, etc.) and the text to be added doesn't already match the anchor's `innerText`.
- If we're adding a new anchor without existing HTML selected in the editor, always add text.
- If we're adding a new anchor to a selection of existing HTML, add text if and only if the selection contains only inline text nodes and the text to be added doesn't already match the selection.


components/api-wrapper.js
```diff
@@ -57,7 +57,7 @@ export class InternalAPIWrapper {
 		this.#component._addToSharedDataCache(key, value);
 	}
 
-	applyLinkFormatting(href, title) {
+	applyLinkFormatting(href, title, text, target) {
```
grant-cleary, line 60: This `title` argument will be removed when the flag governing this new link experience is removed, but for now we still need to support it.

components/api-wrapper.js
```diff
@@ -66,7 +66,38 @@ export class InternalAPIWrapper {
...
+		if (!text) return;
+
+		const nodeHasNonTextElements = node => {
+			const inlineTextElements = this.getEditorInlineTextElements();
+			return node.nodeType === Node.ELEMENT_NODE && !Object.keys(inlineTextElements).includes(node.tagName);
+		};
+
+		if (anchor) {
+			const wrappedAnchor = new EditorNode(anchor, this.#editor.selection);
+			if (!wrappedAnchor.searchDescendants(nodeHasNonTextElements)) anchor.innerText = text;
```
grant-cleary, line 84: We're using the wrapping here solely to make searching easier, but we're actually just editing the raw anchor tag. This is the condition where we have an existing anchor and need to decide whether to update its text.

components/api-wrapper.js
```diff
@@ -66,7 +66,38 @@ export class InternalAPIWrapper {
...
+			return;
+		}
+
+		const range = this.#editor.selection.getRng();
+
+		if (range.collapsed) {
```
grant-cleary, line 90: If the range is collapsed, that means the user has no text selected and we should always add the text.

components/api-wrapper.js
```diff
@@ -66,7 +66,38 @@ export class InternalAPIWrapper {
...
+			range.insertNode(document.createTextNode(text));
+			return;
+		}
+
+		const wrappedRange = new EditorRange(range, this.#editor.selection);
+
+		if (!wrappedRange.searchContainedNodes(nodeHasNonTextElements) && range.toString() !== text) {
```
grant-cleary, line 97: This is the case where we have a selection and are adding a link to it.

components/api-wrapper.js
```diff
@@ -66,7 +66,38 @@ export class InternalAPIWrapper {
...
+		if (!text) return;
+
+		const nodeHasNonTextElements = node => {
+			const inlineTextElements = this.getEditorInlineTextElements();
+			return node.nodeType === Node.ELEMENT_NODE && !Object.keys(inlineTextElements).includes(node.tagName);
+		};
+
+		if (anchor) {
+			const wrappedAnchor = new EditorNode(anchor, this.#editor.selection);
+			if (!wrappedAnchor.searchDescendants(nodeHasNonTextElements) && anchor.innerText !== text) anchor.innerText = text;
```
grant-cleary, line 84: We're using the wrapping here solely to make searching easier, but we're actually just editing the raw anchor tag. This is the condition where we have an existing anchor and need to decide whether to update its text.

grant-cleary: There's also a bit of a trick with `anchor.innerText` here, along with `range.toString()` below. These functions will wipe out any inline formatting, so, for example, `This text is <strong>emphasized</strong>` will match `This text is emphasized`. This matches TinyMCE's default behaviour, and is done so that users can add a link to formatted text without wiping out the formatting.

components/api-wrapper.js
```diff
@@ -143,6 +174,10 @@ export class InternalAPIWrapper {
 		}
 	}
 
+	getEditorInlineTextElements() {
```
grant-cleary, line 177: I forgot to mention that I'm exposing this specifically as an API because we also need this sort of detection elsewhere. It will be consumed in a later PR.

components/api-wrapper.js
```diff
@@ -66,7 +66,38 @@ export class InternalAPIWrapper {
 		const anchor = this.#editor.dom.getParent(node, 'a');
 		if (!anchor) this.#editor.formatter.remove('link');
 
-		this.#editor.formatter.apply('link', { href: href, title: title }, anchor);
+		const linkAttributes = { href: href };
+		if (title) linkAttributes.title = title;
```
Dan-DeAraujo, line 70: TIL [??=](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_assignment) is a thing in JS, if you want to use it 🤷‍♂️

grant-cleary: We still support Safari 13, so unfortunately I can't. 😞

grant-cleary: It's possible that Rollup would transpile it into something Safari would actually like, but the editor is used outside of BSI as well, so I'm generally nervous about introducing features that require transpiling to work.

Dan-DeAraujo: 13!?!? geezuz, that's like 5 years old... 😬 
but yeah in that case, i wouldn't risk it either 😐

Dan-DeAraujo: oh, sorry i was thinking iOS13, which is older than macOS13... but whatever 🙄

Dan-DeAraujo (approved):

PaulKrznaric (approved):

PR #1987, opened 2024-01-04
Add new link dialog in place of the context menu plugin

grant-cleary:
PR 3: This should be the final HTML editor PR, pending any changes requested by Design. It creates the new add/edit link dialog and hooks it up in the following places:
1. The "Add link" button within the formatting context menu
2. The "Edit link" button within the link display context menu
3. Use of the shortcut Ctrl-K, which formerly launched TinyMCE's link dialog


components/toolbar/link-button.js
```diff
@@ -89,6 +89,13 @@ class LinkButton extends ToolbarItemMixin(LitElement) {
 		`;
 	}
 
+	constructor() {
+		super();
+		this.href = '';
+		this.icon = '';
+		this.text = '';
+	}
```
grant-cleary, line 97: This component wasn't used in a way where the lack of a constructor would be an issue, but it still should really have one.

components/context-menu/link/link-display.js
```diff
@@ -39,8 +39,10 @@ export class LinkDisplayContextMenuPlugin {
 	}
 
 	updateContentElement(contentElm, range, node) {
-		if (node.hasAttribute('href')) contentElm.linkLocation = node.getAttribute('href');
-		if (node.textContent) contentElm.linkText = node.textContent;
+		const anchorNode = this._getAncestorAnchorNode(node);
```
grant-cleary, line 42: This was a bug that couldn't actually come up when using the context menu before, but does become an issue now... If the selection is entirely contained within a mix of text and inline text nodes (e.g. `Test <strong>text</strong>`, then the node passed in here can be one of the contained elements (e.g. the `<strong>` in this case), which leads to using an incorrect `innerText`.

components/context-menu/link/link-display.js
```diff
@@ -23,7 +23,7 @@ export class LinkDisplayContextMenuPlugin {
 		const anchorNode = this._getAncestorAnchorNode(node);
 
 		contentElm.linkLocation = anchorNode.getAttribute('href');
-		contentElm.linkText = anchorNode.textContent;
+		contentElm.linkText = anchorNode.innerText;
```
grant-cleary, line 26: This could capture the contents of script tags and such. This isn't actually relevant because that's not something you can do with TinyMCE's iframe anyways (can't put your cursor inside a `<script>`), but this is still more correct.

components/plugins/link.js
```diff
@@ -17,50 +32,245 @@ export class LinkPlugin {
...
+			const handleShortcutKey = e => {
+				if (e.keyCode !== 75) return;
+				if (!e.metaKey && !e.ctrlKey) return;
+				helpers.commandRegistry.executePluginCommand(cmds.linkDialog);
+				e.preventDefault();
+			};
+
+			helpers.eventRegistry.registerShortcutKeyEvent(handleShortcutKey);
+		} else {
+			helpers.commandRegistry.registerCommand(cmds.linkOpenMenu, undefined, () => {
```
grant-cleary, line 120: Actually, this is a bit weird, and I think I want to take another look at this. This command exists in order to decide whether the `Link` button in the formatting menu should be highlighted (if your selection is in a link). However, it's only ever used in the same context as launching the dialog, and exists separately only because swapping context menus isn't command-based. With the new dialog launch, however, this can be embedded as part of the open dialog command and its existing usages can be refactored to account for that. I'll just need to do some testing to see if I can do that safely now, or if that's something I have to do after the flag has been removed.

components/context-menu/formatting/formatting.js
```diff
@@ -56,14 +57,20 @@ export class FormattingContextMenuPlugin {
...
+		this._useNewLinkExperience = context && context.useNewLinkExperience;
+	}
 
 	render() {
 		return this._render(html`
 			<d2l-htmleditor-button-toggle cmd="${cmds.bold}" icon="bold" text="${this.localize('format:bold')}" no-resize></d2l-htmleditor-button-toggle>
 			<d2l-htmleditor-button-toggle cmd="${cmds.italic}" icon="italic" text="${this.localize('format:italic')}"></d2l-htmleditor-button-toggle>
 			<d2l-htmleditor-separator no-resize></d2l-htmleditor-separator>
-			<d2l-htmleditor-button-toggle cmd="${cmds.linkOpenMenu}" @click="${this._handleLinkAddEditClick}" icon="link" text="${this.localize('link:addEdit')}" no-execute no-resize></d2l-htmleditor-button-toggle>
+			<d2l-htmleditor-button-toggle cmd="${this._useNewLinkExperience ? cmds.linkDialog : cmds.linkOpenMenu}" @click="${this._handleLinkAddEditClick}" icon="link" text="${this.localize('link:addEdit')}" ?no-execute="${!this._useNewLinkExperience}" no-resize></d2l-htmleditor-button-toggle>
```
grant-cleary, line 73: I tested this, and this does indeed work fine with and without the feature enabled.

components/context-menu/link/link-actions.js
```diff
@@ -38,31 +40,21 @@ class LinkActionsContextMenuToolbar extends RequesterMixin(ContextMenuToolbarMix
 
 		return this._render(html`
 			<d2l-htmleditor-link-button href="${location}" icon="open-link" text="${this.localize('link:open')}" no-resize></d2l-htmleditor-link-button>
-			<d2l-htmleditor-button @click="${this._handleEditClick}" icon="edit" text="${this.localize('link:edit')}" no-resize></d2l-htmleditor-button>
-			<d2l-htmleditor-button @click="${this._handleRemoveClick}" icon="remove-link" text="${this.localize('link:remove')}" no-resize></d2l-htmleditor-button>
+			<d2l-htmleditor-button cmd="${ifDefined(this._useNewLinkExperience ? cmds.linkDialog : undefined)}" @click="${this._handleEditClick}" icon="edit" text="${this.localize('link:edit')}" no-resize></d2l-htmleditor-button>
+			<d2l-htmleditor-button cmd="${ifDefined(this._useNewLinkExperience ? cmds.linkRemove : undefined)}" icon="remove-link" text="${this.localize('link:remove')}" no-resize></d2l-htmleditor-button>
```
grant-cleary, line 44: Ah shoot, this was an incomplete application of the feature flag. Going to need to correct this.

PaulKrznaric (approved):

Dan-DeAraujo (approved):

PR #1994, opened 2024-01-08
Always display text input for links, but disable when it's not safe

grant-cleary:
Design has requested that we always display the link text input, but that we disable it when it's not safe to add/edit text.


PaulKrznaric (approved):

components/plugins/link.js
```diff
@@ -217,17 +217,22 @@ class LinkDialog extends Localizer(LitElement) {
...
-							<d2l-input-text
-								label="${this.localize('link:textLabel')}"
-								name="text"
-								size="50"
-								type="text"
-								.value="${live(this.text || '')}">
-							</d2l-input-text>
-						</div>
-					` : ''}
+					<div class="d2l-link-text-input-container">
```
klhuff, line 220: Is there any potential for problems like data within the field being sent with a form submission where it may not have before?

klhuff (approved):

components/plugins/link.js
grant-cleary: Nah, that's handled here: https://github.com/BrightspaceUI/htmleditor/pull/1994/files#diff-865659674a8c83adbf109ccee0eaff5f4bdec9746eb04f1b4f27f56c8b38aac0R281

It shouldn't be sent through the form submission anyways, but as a precaution, if we're not supposed to be consuming text here, we just don't propagate it back.

PR #2012, opened 2024-01-24
Remove toolbar refactor and BSI plugin loading features

grant-cleary:
Time to finally remove the toolbar refactor feature. With it, we also need to remove the BSI loading feature (which has been soaking for the same a mount of time), or else C+ plugins will no longer load properly in the editor.

There's not too much to say here, beyond the removal being relatively complex (lots of moving pieces, unfortunately). Tests and linting passed locally, which provides at least some confidence I haven't removed anything egregious, but it will still require extensive testing.


grant-cleary:
The removed goldens look fine, but I'm going to need to check the dimensions of the editor locally to make sure those breakpoints are playing nicely. It's plausible that the issue might be a loading issue; i.e. perhaps the toolbar hasn't fully rendered by the time the screenshot has taken, so its measurements are off. But I'll need to check that.

test/visual-diff/htmleditor.vdiff.js
```diff
@@ -112,8 +112,8 @@ describe('htmleditor', () => {
 				{ name: 'width', opts: { width: '600px' } },
 				{ name: 'word-count-in-footer', opts: { wordCount: true } },
 				{ name: 'invalid', opts: { omitHtml: true, required: true }, action: elem => triggerValidation(elem) },
-				{ name: 'breakpoint-table', opts: { wrapper: true, wrapperWidth: '650px' }, action: elem => toggleChomping(elem) },
-				{ name: 'breakpoint-view-options', opts: { wrapper: true, wrapperWidth: '640px' }, action: elem => toggleChomping(elem) },
+				{ name: 'breakpoint-table', opts: { wrapper: true, wrapperWidth: '700px' }, action: elem => toggleChomping(elem) },
+				{ name: 'breakpoint-view-options', opts: { wrapper: true, wrapperWidth: '650px' }, action: elem => toggleChomping(elem) },
```
grant-cleary, line 116: There wasn't enough give room in these values, so between padding, margins, etc., the toolbars were actually rendering below the intended width breakpoints.

components/plugins/preview.js
```diff
@@ -34,7 +34,7 @@ export class PreviewPlugin {
 			{
 				breakpoint: {
 					type: breakPointTypes.BELOW,
-					value: breakPoints.SMALLER
+					value: breakPoints.SMALL
```
grant-cleary, line 37: Stuff in the view section/menu actually had a breakpoint one level too high, so I changed those across the board. This is strictly a visual change so I'm not really concerned about it having been live for a bit.

klhuff (approved):

davewagler-d2l (approved):

PR #2022, opened 2024-02-02
Make sure code snippets aren't focusable when we render them in the HTML editor

grant-cleary:
When rendering a code snippet, Prism checks to see if it has a `pre` element parent, and if so, sets `tabindex="0"` on it to make it focusable, presumably because (by default) Prism's code snippets are interactive/editable. However, we explicitly don't want these to be editable, especially directly in the HTML editor, so we also don't really want this focusable behaviour added.

Unfortunately this can't be configured within Prism without making changes that could impact rendered code snippets outside of the editor, so my solution here is to just make sure the code snippet isn't focusable in the editor itself.


components/plugins/insert-code.js
```diff
@@ -114,7 +114,8 @@ export class InsertCodePlugin {
 				elem.dataset.originalCode = helpers.internalAPIWrapper.domEncode(elem.textContent);
 			}
 
-			elem = await formatCodeElement(elem);
+			await formatCodeElement(elem);
+			if (elem.hasAttribute('tabindex')) elem.setAttribute('tabindex', '-1');
```
grant-cleary, line 118: This is technically stronger than just removing the tabindex, because if the element is (somehow) natively focusable, we'll also break that behaviour here. In practice that will never happen, because only `code` and `pre` elements get passed in here, but this is still deliberate. By design, we _never_ want these code snippets to participate in the tab order.

components/plugins/insert-code.js
```diff
@@ -114,7 +114,8 @@ export class InsertCodePlugin {
 				elem.dataset.originalCode = helpers.internalAPIWrapper.domEncode(elem.textContent);
 			}
 
-			elem = await formatCodeElement(elem);
```
grant-cleary, line 117: `formatCodeElement` used to return the element after Prism was done working on it, but this was removed because it was redundant. Up until now, it didn't matter that we were setting this to `undefined` because nothing referenced `elem` after this line. This is no longer the case, so we need to remove this assignment.

grant-cleary:
I poked around a bit with VoiceOver and I don't think the screenreader experience here is fantastic when arrowing into one of these code snippets. That's not new, so it's not specific to this defect, but I'm still thinking that after this is done, I want to follow up on that and see if there are improvements we can make.

PaulKrznaric (approved):

dbatiste (approved):

PR #2024, opened 2024-02-05
Don't fetch context object through requestor/provider pattern

grant-cleary:
Exposing the editor's context object via the requestor/provider method allows plugin authors to do an end-run around the encapsulation we're trying for here, since it includes a full copy of the component and thus all of its internals. It also complicates efforts to modify properties on the component and the context object, because it becomes very difficult to tell where they may be used (any descendant of any element rendered within the editor would be suspect).

The goal of this PR is to remove this usage of the context object, noting that we have several general use cases we need to account for:
1. Passing the context object to _potentially external_ consumers (context menu contents)
2. Passing the context object to entirely internal consumers (resizer, word count button)
3. Localizing the internal `table-create` component
4. Redundant usages that can already be handled by the new plugin framework

Case (3) is redundant with recent changes to the `HierarchicalViewMixin` removing the former incompatibility with our resizing, and case (4) is a trivial fix. Thus, the bulk of the changes in this PR will be addressing cases (1) and (2).


components/context-menu/context-menu.js
```diff
@@ -57,9 +57,9 @@ export function initializeContextMenus(component, editor, useContainer) {
 
 	const addOrReplacePlugin = async(contextMenu, plugin, sel) => {
 		if (contextMenu.firstChild && contextMenu.firstChild.tagName === plugin.tagName) {
-			plugin.updateContentElement(contextMenu.firstChild, sel.range, sel.node);
+			plugin.updateContentElement(contextMenu.firstChild, sel.range, sel.node, context);
 		} else {
-			const content = plugin.createContentElement(sel.range, sel.node);
+			const content = plugin.createContentElement(sel.range, sel.node, context);
```
grant-cleary, line 62: The idea here is two-fold:
1. Leave it to the plugin author to decide how to make use of the context object (if at all). We don't really put any restrictions on what this `content` element can be, so we can't/shouldn't assume we can just attach the context object to it.
2. Provide the context object both on creation and on update in case it's necessary to accommodate any structural changes in either case.

components/footer/resizer.js
```diff
@@ -65,14 +64,12 @@ class Resizer extends ThemeMixin(Localizer(RtlMixin(LitElement))) {
 		let content;
 
 		if (this._keyboardActive) {
-			const minHeight = this._context.component._getMinHeight();
-			const maxHeight = this._context.component._getMaxHeight();
-			const height = this._context.component._getHeight();
+			const measures = this._getHeightMeasures();
```
grant-cleary, line 67: @dbatiste, heads up because I know you originally put this resizer together. As far as I can tell, this all functions the same way as it used to from some manual testing.

components/footer/resizer.js
```diff
@@ -99,14 +96,22 @@ class Resizer extends ThemeMixin(Localizer(RtlMixin(LitElement))) {
...
+		const heightCalculationEvent = new CustomEvent(
+			'd2l-htmleditor-resize-height-calculation', {
+				bubbles: true,
+				composed: true,
+				detail: {}
 			}
-		}
+		);
+
+		this.dispatchEvent(heightCalculationEvent);
```
grant-cleary, line 108: The idea here is for the resizer to take on the same behaviour as the toolbar buttons here, communicating with the component via events instead of coupling to the component directly.

htmleditor-mixin.js
```diff
@@ -916,7 +915,7 @@ export const HtmlEditorMixin = superclass => class extends Localizer(LabelledMix
 				});
 
 				if (this.fullPage) this._initializeFullPageState(editor);
-				initializeContextMenus(this, editor, true, this._context && this._context.useContextMenuContainer);
+				initializeContextMenus(this, editor, true, this._context, this._context && this._context.useContextMenuContainer);
```
grant-cleary, line 918: This maybe poses a problem in the future because it assumes `this._context` is mutable (if it's not, context menus will no longer get updated if the embedded context object changes). In practice though, this context object already has to be more or less immutable after it's first instantiation, because in most cases, we would need to reinitialize TinyMCE entirely if the context object changed.

htmleditor.js
```diff
@@ -766,6 +781,14 @@ class HtmlEditor extends HtmlEditorMixin(SkeletonMixin(LitElement)) {
 		});
 	}
 
+	_validateResizerEvent(event) {
+		const rootTarget = event.composedPath()[0];
+
+		return rootTarget
+			&& rootTarget.nodeType === Node.ELEMENT_NODE
+			&& rootTarget.tagName === 'D2L-HTMLEDITOR-RESIZER';
```
grant-cleary, line 789: Similar to toolbar items, we don't want arbitrary descendants of the editor calling these functions.

davewagler-d2l (approved):

PaulKrznaric (approved):

PR #2029, opened 2024-02-08
Use correct context object with link context menu

grant-cleary:
There were two main issues at play here, one of which was hiding the other:
1. A previous change in November introduced an unexpected extra parameter when initializing the context menu. This wouldn't have had a noticeable impact.
2. Because of the above, passing in the context object to the context menu framework didn't work correctly, as the context menu was taking on the value of that extra parameter. Fixing this exposed a bug preventing the link context menu from working correctly due to the context object not being passed down properly.


htmleditor-mixin.js
```diff
@@ -915,7 +915,7 @@ export const HtmlEditorMixin = superclass => class extends Localizer(LabelledMix
 				});
 
 				if (this.fullPage) this._initializeFullPageState(editor);
-				initializeContextMenus(this, editor, true, this._context, this._context && this._context.useContextMenuContainer);
+				initializeContextMenus(this, editor, this._context, this._context && this._context.useContextMenuContainer);
```
grant-cleary, line 918: This is the removal of the extra unintended parameter.

components/context-menu/link/link-display.js
```diff
@@ -103,7 +106,7 @@ class LinkDisplayContextMenu extends RtlMixin(LitElement) {
 					<h4 class="d2l-heading-4">${this.linkText}</h4>
 					<span class="d2l-body-small">${this.linkLocation}</span>
 				</div>
-				<d2l-htmleditor-context-menu-link-actions location="${this.linkLocation}"></d2l-htmleditor-context-menu-link-actions>
+				<d2l-htmleditor-context-menu-link-actions .context="${this.context}" location="${this.linkLocation}"></d2l-htmleditor-context-menu-link-actions>
```
grant-cleary, line 109: This is the first place where the context object wasn't correctly passed in.

components/context-menu/formatting/formatting.js
```diff
@@ -76,7 +76,7 @@ class FormattingContextMenu extends Localizer(ContextMenuToolbarMixin(LitElement
 			<d2l-htmleditor-button-toggle cmd="${cmds.bold}" icon="bold" text="${this.localize('format:bold')}" no-resize></d2l-htmleditor-button-toggle>
 			<d2l-htmleditor-button-toggle cmd="${cmds.italic}" icon="italic" text="${this.localize('format:italic')}"></d2l-htmleditor-button-toggle>
 			<d2l-htmleditor-separator no-resize></d2l-htmleditor-separator>
-			<d2l-htmleditor-button-toggle cmd="${this._useNewLinkExperience ? cmds.linkDialog : cmds.linkOpenMenu}" @click="${this._handleLinkAddEditClick}" icon="link" text="${this.localize('link:addEdit')}" ?no-execute="${!this._useNewLinkExperience}" no-resize></d2l-htmleditor-button-toggle>
+			<d2l-htmleditor-button-toggle cmd="${this.context.useNewLinkExperience ? cmds.linkDialog : cmds.linkOpenMenu}" @click="${this._handleLinkAddEditClick}" icon="link" text="${this.localize('link:addEdit')}" ?no-execute="${!this.context.useNewLinkExperience}" no-resize></d2l-htmleditor-button-toggle>
```
grant-cleary, line 79: And this was the second place, since the context object was passed in, but ultimately unused.

PaulKrznaric (approved):

klhuff (approved):

PR #2050, opened 2024-02-28
Cannot read property total of undefined

MitchelPaulin:
A lot of the errors we are getting in rubrics are related to `total` being undefined 

Untitled

This is being thrown from the HTML editor. It seems this is due to some race conditions between the call to `_updateMeasures` and `_position`, though I've been unable to reproduce this locally. To avoid this I've used 0 as the default if the total is not available. 

Namely, the entire error is 

```
TypeError: Cannot read properties of undefined (reading 'total')
    at https://s.brightspace.com/lib/bsi/20.24.2-169/unbundled/htmleditor.js:507:2600
```

Inspecting https://s.brightspace.com/lib/bsi/20.24.2-169/unbundled/htmleditor.js: and jumping to line 507 you can see it corresponds to line 420 in toolbar.js


grant-cleary (approved):

grant-cleary:
Thanks for the fix! I'm reasonably sure that if visual diff tests pass, this should be safe. The toolbar anchoring isn't likely to change after the fact. 🤔

PR #2067, opened 2024-03-15
Switch over HTML editor to publish to CodeArtifact

grant-cleary:
We want to get the HTML editor publishing to CodeArtifact instead of npm. This needs to wait on:
1. This repo being made `internal`.
2. This PR merging: https://github.com/Brightspace/repo-settings/pull/1771

Once those two steps have been completed and this is merged, we can start switching over consumers.


.github/workflows/release.yml
```diff
@@ -34,17 +44,13 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.D2L_RELEASE_TOKEN }}
           MINOR_RELEASE_WITH_LMS: true
           NPM: true
-          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
           RALLY_API_KEY: ${{ secrets.RALLY_API_KEY }}
-      - name: Get new git HEAD
-        id: git
-        run: echo "head=$(git rev-parse HEAD)" >> $GITHUB_OUTPUT
```
svanherk, line 41: Looks like this is used by the publish job below to checkout the right version of the code

grant-cleary: Oh, yeah, I meant to remove this from another repo where I copied this in by accident. 🤦

package.json
```diff
@@ -1,12 +1,9 @@
 {
-  "name": "@brightspace-ui/htmleditor",
+  "name": "@d2l/htmleditor",
   "description": "An HTML editor that integrates with Brightspace",
   "version": "2.111.2",
   "type": "module",
   "repository": "https://github.com/BrightspaceUI/htmleditor.git",
-  "publishConfig": {
-    "access": "public"
-  },
```
svanherk, line 9: We tend to set the `publishConfig` to [this](https://github.com/BrightspaceHypermediaComponents/users/blob/6aa6e649d5507c37e828b0f09a0fae75f4e8b344/package.json#L30-L32), although for some reason we don't in the other `internal` repos in `BrightspaceUI` 🤔

grant-cleary: Yeah, I based this config mostly on what `lms-core` is doing. I'm not really sure what the difference is; certainly the new repo I just spun up doesn't have this config set and it did publish fine. So I'm kind of curious what it's for.

grant-cleary: Hmm, looking at the documentation, it feels like it's good to include, since it would prevent publishing to npm by accident. I don't think it's really a thing in the release action because we're not providing the appropriate credentials to publish to our npm org. But I assume someone could `npm publish` from the command line and if they provided valid credentials, they could end up publishing this to npm by accident? I'm not sure. Maybe it's worth including just in case.

svanherk: Ah that makes sense, up to you!

svanherk (approved):
You'll need to wait until it's actually `internal` for CI to run, but this looks good to me!

